PR #1786, opened 2014-08-07
[BACKLOG-296] - copying doIsParameterizable from FileResource into a new...

kurtwalker:
... class ContentResource.  The original endpoint remains and calls the new method.


pentaho-nbaker:
I like this change. Going to re-open and push it back in.

pamval:
This needs to be rebased and reissued when appropriate (and using a more recent jira key)

PR #2102, opened 2014-10-17
[TEST] Uncomment tests in MultipleComponentTest

annapopova1:
Recover tests xaction with jfree reports.

For correct process jfree reports are needed:
  1. ClassicEngineBoot.getInstance().start();
  2. classic-engine.properties in test-src;
  3. meta-datafactory.xml in org.pentaho.platform.plugin.action.jfreereport.helper

For resolve problem with lock sampledata file are needed use readonly=true:
  In extensions/test-src/solution/system/simple-jndi/jdbc.properties use property SampleData/url=jdbc:hsqldb:file:test-src/solution/system/data/sampledata;readonly=true

@rmansoor verify please


pamval:
test org.pentaho.test.platform.plugin.MultipleComponentTest.testCrosstab is failing.
I'm closing this one. Please reissue when that is fixed.

PR #2125, opened 2014-11-12
Add checkbox to schedule wizard for overwriting existing content

bluezio:
This merge request is based on this forum thread:

http://forums.pentaho.com/showthread.php?176243-Overwrite-scheduled-report-output-from-5-0-7-CE

I've added a checkbox to the "job parameters" step of the schedule creation wizard which controls the "autoCreateUniqueFilename" job parameter ("overwrite existing content" seems easier to understand than "create unique filename"). Normally I would have put it in the first step, together with the output location and schedule name, but that'd require propagating the state of the checkbox through the rest of the steps of the wizard.


dkincade:
This can no longer be automatically merged. Please re-issue the pull request. 
Is this in reference to an existing JIRA improvement?

PR #2127, opened 2014-11-14
[BISERVER-12271] Select button in Schedule... stops working if there is a file in the root solution folder

bluezio:
More information is available here:

http://jira.pentaho.com/browse/BISERVER-12271

It's a simple nullness check, nevertheless :-).


dkincade:
Code reviewed and approved

PR #2152, opened 2014-12-09
[BISERVER-10354] - Adding the requestParameterProcessingFilter filter to...

apratkin:
... the /api/** and /plugin/** pattern causes Analyzer to no longer work.

In case we use requestParameterProcessingFilter parameters of post request comes with httpServletRequest not with formParams (RepositoryResource. doFormPost)

Issue:
In case we use requestParameterProcessingFilter parameters of post request comes with httpServletRequest not with formParams (RepositoryResource. doFormPost).
Solution:
Add parameters from httpServletRequest to map of params.
Tested:
1.	Build pentaho-platformand put pentaho-platform-extensions-TRUNK-SNAPSHOT.jar at 
\biserver-ee\tomcat\webapps\pentaho\WEB-INF\lib\ 
2.	Run server. Open report Country Performance (heat grid). Verify it looks correct.
3.	Open applicationContext-spring-security.xml file and add requestParameterProcessingFilter to /api/** and /plugin/**
4.	 Run server. Open report Country Performance (heat grid). Verify it looks correct.


apratkin:
Pull request was updated with new changes

dkincade:
@pentaho-nbaker pull request updated...

PR #2188, opened 2015-01-09
[BACKLOG-1561] - Add the ability to get security/ACL information about a...

aliakseihaidukou:
... datasource

add get/set methods to AclAware interfaces and realizations

@pentaho-nbaker this PR is for https://github.com/pentaho/data-access/pull/592

Diogo offered to add one more interface with common methods for IAclAwareMondrianCatalogService and IAclAwarePentahoMetadataDomainRepositoryImporter. Do you think we need to do this?

### DIFF
--- a/extensions/src/org/pentaho/platform/plugin/action/mondrian/catalog/IAclAwareMondrianCatalogService.java
+++ b/extensions/src/org/pentaho/platform/plugin/action/mondrian/catalog/IAclAwareMondrianCatalogService.java
@@ -26,9 +26,7 @@ public interface IAclAwareMondrianCatalogService extends IMondrianCatalogService
   void addCatalog( InputStream inputStream, MondrianCatalog catalog, boolean overwriteInRepository,
                    RepositoryFileAcl acl, IPentahoSession session );
 
-  /**
-   * Never returns <tt>null</tt>.
-   * @return    an ACL helper instance
-   */
-  IAclNodeHelper getAclHelper();
+  void setAclFor( String catalogName, RepositoryFileAcl acl );
+
+  RepositoryFileAcl getAclFor( String catalogName );
 }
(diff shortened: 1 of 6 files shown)


pentaho-nbaker:
I would not object to a common Interface between the two, if it's a good fit. I don't want to overly complicate things. It's not needed for now though.

extensions/src/org/pentaho/platform/plugin/action/mondrian/catalog/IAclAwareMondrianCatalogService.java
```diff
@@ -31,4 +31,8 @@ void addCatalog( InputStream inputStream, MondrianCatalog catalog, boolean overw
    * @return    an ACL helper instance
    */
   IAclNodeHelper getAclHelper();
```
pentaho-nbaker: I would like to not have this method if possible.

extensions/src/org/pentaho/platform/plugin/services/metadata/IAclAwarePentahoMetadataDomainRepositoryImporter.java
```diff
@@ -27,4 +27,8 @@ void storeDomain( InputStream inputStream, String domainId, boolean overwrite, R
   IAclNodeHelper getAclHelper();
```
pentaho-nbaker: Same, would like to not have getAclHelper()

extensions/src/org/pentaho/platform/plugin/services/metadata/IAclAwarePentahoMetadataDomainRepositoryImporter.java
```diff
@@ -27,4 +27,8 @@ void storeDomain( InputStream inputStream, String domainId, boolean overwrite, R
...
 
   RepositoryFile getDomainRepositoryFile( String domainId );
```
pentaho-nbaker: Who needs this? If possible, I would avoid adding this method

aliakseihaidukou:
@pentaho-nbaker I removed those methods from interfaces. No changes to data-access needed, they were used only inside implementations.

PR #2205, opened 2015-01-14
[BISERVER-11865] CDE folder creation mechanism (on plugin boot-up) does ...

pedrofvteixeira:
...not work in 5.1.0


pedrofvteixeira:
@CodeOnCoffee, @pamval please review

pedrofvteixeira:
@pamval : should be merged BEFORE https://github.com/webdetails/cde/pull/249

pentaho-nbaker:
I've reverted this change due to issues found in testing. We'll be using a different approach which won't require this change

PR #2207, opened 2015-01-15
BISERVER-12183 Starting BA server from tomcat (as service) fail

AliaksandrDrebenchuk:
service scripts was added


dkincade:
We need to discuss adding these new scripts as there is significant impact to doing this and it may not be the correct way to resolve this problem. We will re-address this item post 5.3

PR #2208, opened 2015-01-15
BISERVER-12194 -- Unable to remove sample JDBC connection through PUC

kolinus:
What was done:
1) on fail with connection permanent removing -- move it to trash folder

So this fix is to return 5.0.x behavior which was asked by customers

### DIFF
--- a/repository/src/org/pentaho/platform/repository/JcrBackedDatasourceMgmtService.java
+++ b/repository/src/org/pentaho/platform/repository/JcrBackedDatasourceMgmtService.java
@@ -138,11 +138,15 @@ public class JcrBackedDatasourceMgmtService implements IDatasourceMgmtService {
             "DatasourceMgmtService.ERROR_0002_UNABLE_TO_DELETE_DATASOURCE", "", "" ) ); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$       
       }
     } catch ( UnifiedRepositoryException ure ) {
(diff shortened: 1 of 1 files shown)


kolinus:
@rmansoor  Ramaiz, please review it.

rmansoor:
@kolinus  The change does not make sense. The datasource was not deleted because of the referential integrity. This means that this datasource is being reference by a KTR or KJB. We need to either delete those KTR or KJBs first then try to delete this datasource.  The reason you were able to delete this datasource because all you are doing is moving this datasource into the trash folder. This is not a bug

kolinus:
@rmansor please look at the bug. I do agree that this does not really deletes but just hides the datasource, and it did not work  in the past either, but customers want this.

rmansoor:
@kolinus By moving it into the trash, we will cause other issues. We are only moving file and folder to trash. Datasource and all other shared object are being deleted permanently because there is no way to restore them. Repository can grow without the knowledge of customer.

kolinus:
@rmansoor  perhaps this should be communicated to PMs ?

kolinus:
@rmansoor  yes, that's why I'm moving it to trash only if we failed to really delete that. I do not think we should automatically remove reports if a person just want to remove dataspurce. So I have no idea what to do here. Perhaps this is not a bug at all?

dkincade:
This is being pulled from 5.3 ... this pull request can be reopened in the 5.4 codeline after more decisions have been made.

PR #2209, opened 2015-01-15
[TEST] - fix unit-tests

aliakseihaidukou:
now every project has it's own repo path

@dkincade this PR and https://github.com/pentaho/data-access/pull/599 are fixing failing repository and data-access tests, but I'm afraid I'm late..


dkincade:
@mdamour1976 @e-cuellar can you please take a look ... (it is ok for this to go into master and be cherry-picked into 5.3 if only unit tests)

PR #2211, opened 2015-01-19
[PDI-13012] - 5.2.0 Missing "MSSQL" directory with DDL's from the "data"...

aliakseihaidukou:
... folder

added sql server dialect, hibernate config


assembly/package-res/biserver/pentaho-solutions/system/dialects/sqlserver/applicationContext-spring-security-hibernate.properties
```diff
@@ -0,0 +1,5 @@
+jdbc.driver=com.microsoft.sqlserver.jdbc.SQLServerDriver
+jdbc.url=jdbc:sqlserver://localhost:1433;DatabaseName=hibernate
```
rmansoor: @aliakseihaidukou  we are making this change for PDI  only. This change will effect ba server. So we do not want to this right now. Lets keep this change to pdi only

PR #2220, opened 2015-01-20
PDI-13385 - [Regression] Pan and Kitchen could not run transformations a...

kolinus:
...nd jobs from DI server repository

What was done:
1) moved the call of getting owner to include ACL check. Currently that
call causes Kettle to fail when dealing with DI repository as it tries to
initialize web service endpoint on Spoon side which should not be done.


kolinus:
@pentaho-nbaker , Nick could you please review it?

kolinus:
@pentaho-nbaker , Nick, any update?

PR #2222, opened 2015-01-23
[BISERVER-12393] - Fixed Non-JNDI Datasource cache initial setup

diogofscmariano:
@dkincade please review

### DIFF
--- a/core/src/org/pentaho/platform/engine/services/connection/datasource/dbcp/NonPooledDatasourceSystemListener.java
+++ b/core/src/org/pentaho/platform/engine/services/connection/datasource/dbcp/NonPooledDatasourceSystemListener.java
@@ -66,6 +66,10 @@ public class NonPooledDatasourceSystemListener implements IPentahoSystemListener
           if ( !databaseConnection.getAccessType().equals( DatabaseAccessType.JNDI ) ) {
             // if connection's port used by server there is no sense to get DataSource for this
             ds = isPortUsedByServer( databaseConnection ) ? null : setupDataSourceForConnection( databaseConnection );
+          } else {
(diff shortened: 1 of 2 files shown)


dkincade:
@pentaho-nbaker please review

core/src/org/pentaho/platform/engine/services/connection/datasource/dbcp/NonPooledDatasourceSystemListener.java
```diff
@@ -66,6 +66,8 @@ public boolean startup( final IPentahoSession session ) {
           if ( !databaseConnection.getAccessType().equals( DatabaseAccessType.JNDI ) ) {
             // if connection's port used by server there is no sense to get DataSource for this
             ds = isPortUsedByServer( databaseConnection ) ? null : setupDataSourceForConnection( databaseConnection );
+          } else {
+            ds = setupDataSourceForConnection( databaseConnection );
```
pentaho-nbaker: setupDataSourceForConnection doesn't seem to handle JNDI connections. Would it be better to simply not populate the cache with this datasource period?

pentaho-nbaker: Yea looks like if it's JNDI, setupDataSourceForConnection will catch an exception from PooledDatasourceHelper.convert and simply return null. Since we know it's JNDI ahead of time, I would rather simply "continue;" the loop and skip it entirely.

diogofscmariano: Ok. Changing

diogofscmariano:
@pentaho-nbaker pr updated

PR #2225, opened 2015-01-29
[PRD-5353] - Regression: Hide Parameter UI System Parameter on Drilldown...

aliakseihaidukou:
... function downloads a PRPT report instead of hiding the parameter UI of the target report

switched from "content" to "viewer?showParameters=false" URL

@tmorgner please review(it was hard to realize that the drilldown options are overwritten here, is the usage of this file documented somewhere?)


aliakseihaidukou:
@tmorgner looks like you missed this PR

aliakseihaidukou:
@tmorgner can you look at this, please?

PR #2228, opened 2015-02-09
[BISERVER-12360] Importing or uploading an Analyzer report triggers

mkambol:
Mondrian schema cache to be cleared for all cubes on Server

Formerly cache clearing used an unnecessarily conservative strategy,
flushing all Mondrian catalog during any import action that caused a
reInit of the Mondrian catalog.  This may have been done in part
because the PentahoSystem cache (MONDRIAN_CATALOG_CACHE_REGION) needs
to be flushed in its entirety when changes are made, and the Mondrian
cache was being flushed at the same time.

This change separates Mondrian cache flushing from the
MondrianCatalogHelper.reInit() process, and confines cache flushing to
just the relevant catalog being added.  This eliminates cache flush
during analyzer report imports, for example, when the import does not
actually bring in a schema.  Cache clearing during .removeCatalog was
deliberately left off, since Mondrian will reclaim memory independently, 
and avoiding the flush speeds the UI experience.

@lucboudreau


pentaho-nbaker:
This looks good to me. Thanks Matt

mkambol:
Luc  just pointed out I missed a connection close.  A second PR with a fix for hthat will be coming shortly.

PR #2230, opened 2015-02-09
[BISERVER-12360]  2 more tweaks:  added a check for schema existence bef...

mkambol:
...ore flushing to avoid  attempts to connect to schemas that do not yet exist.  As a result of this change, also brought flush into the .removeCatalog(), since otherwise it would be possible for a schema which was removed and then re-added from platform to pick up an out-of-sync cache from Mondrian.


mkambol:
@lucboudreau

PR #2232, opened 2015-02-17
[BISERVER-12358] - Currently out-of-box Report Author Role does not work...

aliakseihaidukou:
... with existing permissions

added Create Content permission to the Report Author role


aliakseihaidukou:
@pentaho-nbaker please review

PR #2242, opened 2015-03-02
[BACKLOG-2305] Validate the function call and dev doc the parameters and...

annapopova1:
... return values.

Added jmeter test and changed Javadoc for documentation REST service. 

@brendan-lynch @krivera-pentaho 
Could you check how this Javadoc will be represented into MindTouch?


brendan-lynch:
@annapopova1 Yes I am going to generate the enunciate and check it out, if there are any corrections I can make them myself. Everything looks good though.

PR #2251, opened 2015-03-06
[TEST] DO NOT MERGE

lucboudreau:


extensions/test.deleteme
```diff
@@ -0,0 +1 @@
+this is a test. ignore.
```
buildguy: Nice change

buildguy: Terrible change

PR #2259, opened 2015-03-13
[BISERVER-12474] - Disable "Recents" widgets from PUC Home Screen

matthewtckr:
Change the if statement to properly disable the recents widget

Original PR #2258


dkincade:
@mdamour1976 @rfellows 
Is this fix correct and complete (do we have tests on this area os JS?)

rfellows:
Looks fine. 
No, we don't have unit tests for this stuff.

PR #2266, opened 2015-03-25
BISERVER-12450 - Regression : pentaho_user causes issues when browsing r...

Aliaksandr-Kastenka:
...epository


Aliaksandr-Kastenka:
@e-cuellar 
Please review.

PR #2270, opened 2015-03-31
PRD-5384 - Get publish success message when user has no write permission...

Aliaksandr-Kastenka:
... to target location

Goes with https://github.com/pentaho/pentaho-reporting/pull/606


Aliaksandr-Kastenka:
@e-cuellar
Please review.

PR #2273, opened 2015-04-10
[BACKLOG-2452] - Added support for WADL extensibility via external file ...

diogofscmariano:
...generated in build time

 - Added new project that generates a jar used in build time to generate WADL files
 - Unit tests created for every new class added

### DIFF
--- a/build-utils/src/org/pentaho/wadl/PentahoResourceDoclet.java
+++ b/build-utils/src/org/pentaho/wadl/PentahoResourceDoclet.java
@@ -1,0 +1,157 @@
+package org.pentaho.wadl;
+
+import com.sun.javadoc.AnnotationDesc;
+import com.sun.javadoc.ClassDoc;
+import com.sun.javadoc.MethodDoc;
+import com.sun.javadoc.RootDoc;
+import com.sun.jersey.server.wadl.generators.resourcedoc.model.ClassDocType;
+import com.sun.jersey.server.wadl.generators.resourcedoc.model.MethodDocType;
+import com.sun.jersey.server.wadl.generators.resourcedoc.model.ResourceDocType;
+import com.sun.jersey.wadl.resourcedoc.DocProcessorWrapper;
+import com.sun.jersey.wadl.resourcedoc.ResourceDoclet;
+import org.apache.xml.serialize.OutputFormat;
+import org.apache.xml.serialize.XMLSerializer;
+
+import javax.xml.bind.JAXBContext;
+import javax.xml.bind.Marshaller;
+import java.io.BufferedOutputStream;
+import java.io.FileOutputStream;
+import java.io.OutputStream;
+import java.text.MessageFormat;
+
+public class PentahoResourceDoclet extends ResourceDoclet {
+  private static final String OUTPUT_FILE_NAME = "wadlExtension.xml";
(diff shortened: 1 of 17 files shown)


diogofscmariano:
@CodeOnCoffee @lgrill-pentaho @pamval please review
As soon as this merge is accepted a build in ci needs to be created to generate the new artefact (from build-utils) that is needed by platform-extensions. 
Once this gets merged I will do the following commit that changes the pentaho-platform/extensions and data-access that need this new project

diogofscmariano:
Just for reference, the other commit is under https://github.com/diogofscmariano/pentaho-platform/tree/BACKLOG-2452-part2
The same will happen for data-access under https://github.com/diogofscmariano/data-access/tree/BACKLOG-2452

diogofscmariano:
@pentaho-nbaker can you please review?

build-utils/src/org/pentaho/wadl/PentahoResourceDoclet.java
```diff
@@ -0,0 +1,129 @@
...
+    for ( AnnotationDesc annotationDesc : annotationDescs ) {
+      if ( "deprecated".equalsIgnoreCase( annotationDesc.annotationType().name() ) ) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private static boolean isPublic( String documentation ) {
+    return documentation != null && !"".equals( documentation.trim() );
```
pentaho-nbaker: Shouldn't we be looking for @org.codehaus.enunciate.Facet ( name = "Unsupported" ) here? Note I want to create our own annotation for this, but this is the one the team is currently using:

diogofscmariano: Agree. We can search for that too. But we added it probably because
https://github.com/pentaho/pentaho-platform/blob/master/extensions/src/org/pentaho/platform/web/http/api/resources/SchedulerResource.java#L341
Do they represent the same?

build-utils/src/org/pentaho/wadl/PentahoResourceDoclet.java
```diff
@@ -0,0 +1,129 @@
+package org.pentaho.wadl;
```
pentaho-nbaker: I would love to see this class be extensible/pluggable with a collection of IDocContributors or something which can be plugged in. This is fine for now though

build-utils/src/org/pentaho/wadl/PentahoResourceDoclet.java
pentaho-nbaker: This is the visibility (supported method) check right? Deprecated would be a different flag.

diogofscmariano: Supported methods are identified by having documentation defined... Deprecated need to be explicitly marked. Not sure i got your question

pentaho-nbaker: Just because there's Doc doesn't mean it's supported. We doc for things internal for us only. If @Facet ( name = "Unsupported" ) is not present, it's supported.

diogofscmariano:
@CodeOnCoffee updated pull request with your comments
can you please review?

PR #2278, opened 2015-04-16
[BACKLOG-2452] - Added support for WADL extensibility via external file ...

diogofscmariano:
...generated in build time

 - Updated builds to build this new project and generate a new WADL file
 - WADL is generated via PentahoResourceDoclet
 - In run time, the WADL is patched with the content of the file using WadlGeneratorConfig referenced in web.xml and file is retrieved via SprintResource, where each plugin that wants to expose a WADL file just needs to add a bean with that reference
 - Unit tests created for every new class added

@pentaho-nbaker please review


diogofscmariano:
@lgrill-pentaho please review

diogofscmariano:
https://github.com/pentaho/data-access/pull/613
this is the same for the data-access plugin

PR #2282, opened 2015-04-24
[BACKLOG-3032] - Enable editing of schedules created from either pdi-client or user console

rmansoor:
...


rmansoor:
@pedrofvteixeira please review

PR #2285, opened 2015-04-24
[SP-1846] [BACKLOG-3078] Schemas with a space in their names were not ge...

lucboudreau:
...tting flushed after modifications.


lucboudreau:
@mkambol Same for 5.3. Thanks!

PR #2290, opened 2015-04-27
[BACKLOG-3132] Fix sub-resource generation in WSDL file generation for p...

pedrofvteixeira:
...lugins

	- previously we were only checking if endpoint ended in "application.wadl"
	- now we check the endpoint using a regex that allows for either "application.wadl" and wall sub-resources ( ex: "application.wadl/xsd0.xsd" )


pedrofvteixeira:
@rmansoor, @pentaho-nbaker, @tkafalas please review

PR #2294, opened 2015-04-28
[BISERVER-12558] Error accessing WADL sub-resources ( for plugins )

pedrofvteixeira:
- previously we were only checking if endpoint ended in "application.wadl"
	- now we check the endpoint using a regex that allows for either "application.wadl" and wall sub-resources ( ex: "application.wadl/xsd0.xsd" )
        - cherry-picked from https://github.com/pentaho/pentaho-platform/pull/2290


pedrofvteixeira:
@mbatchelor , @rmansoor please review

PR #2307, opened 2015-05-14
BACKLOG-3361 -- French - Login Page, login button text gets cut off.

kolinus:
What was done:
1) added class describing current locale to allow locale specific UI
changes


kolinus:
@rmansoor  could you please review

PR #2313, opened 2015-05-19
BACKLOG-3306 - Checkin 5.4 Localization files from Acclaro (2nd drop)

kolinus:
What was done:
1) added updated files


kolinus:
@dkincade @kcruzada could you please review

PR #2319, opened 2015-05-22
[BISERVER-11812] - Remove hard coded pentaho context and used a

rmansoor:
different parameter to retrieve the base url


rmansoor:
@kcruzada , @dkincade  please review

PR #2321, opened 2015-05-22
BACKLOG-3471 -- L10N - PUC login page is pared down on Login as evalu…

kolinus:
…ator

pop-up

What was done:
1) added support for lang families such as fr-fr, fr-ca, fr-be, ...


kolinus:
@rmansoor , @kcruzada  could you please take alook

PR #2323, opened 2015-05-22
[BISERVER-11812] - BA Server is telling browser to request content us…

aliakseihaidukou:
…ing http even when using an Apache+SSL as reverse proxy.

fix brightcove video player (Welcome and Tutorials tabs)
@rmansoor @dkincade @kcruzada please review

### DIFF
--- a/user-console/source/org/pentaho/mantle/public/home/content/getting_started_launch.html
+++ b/user-console/source/org/pentaho/mantle/public/home/content/getting_started_launch.html
@@ -10,7 +10,7 @@
 
   <!-- We need web context for requirejs and css -->
   <script type="text/javascript" src="webcontext.js?context=mantle&cssOnly=true"></script>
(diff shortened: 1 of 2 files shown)


rmansoor:
@aliakseihaidukou I tried to test this on my end but it still does not bring the brightcoveExperience.js is not being retrieved from the update url. Can you tell me how did you test this one

aliakseihaidukou:
@rmansoor I used apache ssl engine for testing. After I removed "http" the brightcoveExperience.js url was using parent's page protocol. And secureConnections parameter told the plugin to use https for content stream.

rmansoor:
The issue here is that we need to use a different url
in case of http the url is 
http://admin.brightcove.com/js/BrightcoveExperiences.js

for https
https://sadmin.brightcove.com/js/BrightcoveExperiences.js

aliakseihaidukou:
oh, you are right! It works with "https://admin...", but with incorrect ssl certificate. I'll fix now.

kcruzada:
@aliakseihaidukou I'm closing this pull request... please re-open when you have your fix.

PR #2324, opened 2015-05-22
[BISERVER-11812] - BA Server is telling browser to request content us…

aliakseihaidukou:
…ing http even when using an Apache+SSL as reverse proxy.

fix brightcove video player (Welcome and Tutorials tabs)
@rmansoor @dkincade @kcruzada please review


rmansoor:
verified it is working good. processing the pull request now

PR #2327, opened 2015-05-22
[BISERVER-12591]  Flushing the Mondrian Schema Cache invokes

mkambol:
many unnecessary flush/load schema calls

This change splits out flushes of hosted and remote.  For hosted
connections we only need to invoke .flushSchemaCache on a single
catalog, since that will clear the schema pool.  We can't make that
assumption for remote.
(cherry picked from commit 8a6b640)


mkambol:
@lucboudreau

PR #2330, opened 2015-06-01
BISERVER-12584 -- Encrypted repository passwords with Import-Export

kolinus:
utility does not work

What was done:
1) added usage of KettleEncoder


kolinus:
@deinspanjer  could you please review?
last time there was unnecessary import, this time I removed it
thank you

PR #2335, opened 2015-06-04
[BACKLOG-3264] Update the platform code to use jackrabbit 2.10.0 and …

pedrofvteixeira:
…resolve resulting errors

- created new class PentahoEntry that mimics org.apache.jackrabbit.core.security.authorization.acl.Entry but exposes private constructors
- created new inner class PentahoEntries that extends jackrabbit's inner class Entries, but serves a list of PentahoEntry objects
- created new class PentahoEntryFilterImpl that extends jackrabbit's EntryFilterImpl, but works for PentahoEntry objects
- PentahoCompiledPermissionsImpl now checks the object ( instanceof PentahoEntry or instanceof Entry )
- PentahoEntryCollector kept the same ace gathering logic, mapping the ( now deprecated ) List<AccessControlEntry> to List<PentahoEntry>

### DIFF
--- a/extensions/ivy.xml
+++ b/extensions/ivy.xml
@@ -78,7 +78,7 @@
 
     <dependency org="dom4j" name="dom4j" rev="1.6.1" transitive="false"/>
     <dependency org="wsdl4j" name="wsdl4j" rev="1.6.2" transitive="false"/>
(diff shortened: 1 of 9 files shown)


pedrofvteixeira:
@rmansoor , @tkafalas , @CodeOnCoffee , @brendan-lynch please review

PR #2342, opened 2015-06-10
[BACKLOG-3649] - update group-id from com.pentaho to pentaho. changed…

rfellows:
… artifact ids for the metaverse parent and core bundles.


rfellows:
depends on https://github.com/pentaho/pentaho-metaverse-ee/pull/345

PR #2352, opened 2015-06-18
[BACKLOG-3581] Upgrade Hypersonic Library to 2.3.2

tkafalas:


lucboudreau:
Is this really the best solution?

rmansoor:
@lucboudreau , Tim has done most of the research and found out that there is no flag in the latest hsql to ignore case sensitivity. So apparently this is the best solution for us right now

PR #2354, opened 2015-06-21
[BACKLOG-3902] - Update the assembly to use tomcat 8.0.23

rmansoor:
Update the jndi configuration to tomat 8


rmansoor:
@CodeOnCoffee I am going to process this now to check if everything is goes will with the build. But, I would like you to do a post merge code review.

PR #2363, opened 2015-06-26
BISERVER-12646  Cannot run program javadoc.exe: The filename or extension is too long

AliaksandrDrebenchuk:
«Many problems with running javadoc stem from command lines that have become too long - even though the error message doesn't give the slightest hint this may be the problem. If you encounter problems with the task, try to set the useexternalfile attribute to true first.»
Full text at the page https://ant.apache.org/manual/Tasks/javadoc.html

### DIFF
--- a/extensions/build.xml
+++ b/extensions/build.xml
@@ -86,7 +86,7 @@
     <if>
       <istrue value="${ant.version.atleast.1.9.2}"/>
       <then>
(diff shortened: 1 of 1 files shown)


AliaksandrDrebenchuk:
@tkafalas please merge

AliaksandrDrebenchuk:
@tkafalas please merge

PR #2365, opened 2015-06-26
[SP-1910][BISERVER-12592] "Use Description For Tooltips" not reliable

SergeyTravin:


SergeyTravin:
@rmansoor, @e-cuellar, @mdamour1976, @pentaho-nbaker 
I would be grateful if you could review the PR before end of the July SP (6.26.2015)

PR #2368, opened 2015-06-30
Revert "SP-1948 - Backport of BISERVER-12579 - Non-admin User Unable …

Aliaksandr-Kastenka:
@rmansoor 
Please review.

Reverts pentaho/pentaho-platform#2356

Marc Batchelor's comment on SP-1948:
Changing repository.xml cannot be done on an SP for several reasons:
This is a file we expect many of our customers to change. It is backed up and restored as part of the Upgrade Utility as well, so it cannot be fixed / changed in a point release either.
Changing repository.xml will have no effect unless you blow away the repository folder under jackrabbit. This is because the file gets exploded into several other files in that folder.
AFAIK, this change wouldn't take effect unless they started with a completely clean repository.
So - this SP CANNOT be done. Please back out the change, and close this ticket as "Won't Fix".

Goes with https://github.com/pentaho/data-access/pull/643 and https://github.com/pentaho/pentaho-platform-ee/pull/735


mbradbury2014:
@mbatchelor Review and merge please?

PR #2369, opened 2015-06-30
[BACKLOG-3758] - Updating pentaho platform to using the modified spring jar for JRE 8

rmansoor:


rmansoor:
@pentaho-nbaker please review. This change will be removed once we move to latest spring

PR #2370, opened 2015-07-02
[SP-1910][BISERVER-12592] "Use Description For Tooltips" not reliable

SergeyTravin:


SergeyTravin:
@rmansoor, @e-cuellar, 
I would be grateful if you could review the PR as soon as possible. It is included in JulySP.
And it is root case https://github.com/pentaho/pentaho-platform/pull/2371

PR #2387, opened 2015-07-20
[BACKLOG-4266] Save Model Annotations as a DSW Repo File property

bytekast:


bytekast:
@kurtwalker @pminutillo @DFieldFL @bennychow - please review

bennychow:
Hi Rowell,

I thought our design was to save the annotations as a separate file that was separate from the XMI file.  This separation will be cleaner when we eventually support private data sources and there is a 1-M relationship between an XMI file and multiple annotation XML files.

From the code, it looks like you are storing the annotations as a serialized string into the XMI's file property.

Benny

bytekast:
Hi Benny,

This implementation is in line with the recommendations from @kurtwalker's spike.

- Rowell

kurtwalker:
@bytekast Benny is correct.  recommendation was a file outside the xmi and a reference to it in the metadata

bytekast:
so we're storing the xml in another file and storing the reference a property?

kurtwalker:
yes

bennychow:
I don't think we should even have a reference to the annotations.xml file from the XMI file.

bytekast:
@bennychow - the reference would be stored as a property of the repo file, not in the actual xmi file. can you review http://jira.pentaho.com/browse/BACKLOG-3772 to clarify some things? let me look into saving the annotations as a separate file in the meantime...

bytekast:
closing this. will issue a new pull request with the change.

PR #2400, opened 2015-07-28
[BACKLOG-3790] Add support for ranged assignment of conflicting karaf…

tkafalas:
… ports.  Add pentaho/api/karaf/properties web service.


tkafalas:
Nick, can you take a look at this?  Thanks.

assembly/package-res/biserver/pentaho-solutions/system/pentahoServices.spring.xml
```diff
@@ -148,4 +148,5 @@
   </bean>
 
   <bean class="org.pentaho.platform.web.http.api.resources.PasswordResource" scope="request"/>
+  <bean class="org.pentaho.platform.web.http.api.resources.KarafResource" scope="request"/>
```
pentaho-nbaker: Shouldn't be Karaf Specific I wouldn't think. I envision the endpoint returning service-> port assignments for the entire server. The fact that we're using Karaf shouldn't be apparent from the service layer

extensions/src/org/pentaho/platform/osgi/KarafInstancePropertyDto.java
```diff
@@ -0,0 +1,93 @@
...
+import javax.xml.bind.annotation.XmlRootElement;
+
+/**
+ * 
+ * @author tkafalas
+ *
+ */
+@XmlRootElement(name="KarafInstanceProperty")
+@XmlAccessorType(XmlAccessType.FIELD)
+public class KarafInstancePropertyDto implements Serializable {
```
pentaho-nbaker: Why a separate DTO class, couldn't you just annotate the KarafInstanceProperty class?

extensions/src/org/pentaho/platform/osgi/KarafPropertyImporter.java
```diff
@@ -0,0 +1,120 @@
...
+package org.pentaho.platform.osgi;
+
+import java.io.BufferedReader;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+
+import org.apache.commons.lang.StringUtils;
+
+public class KarafPropertyImporter {
```
pentaho-nbaker: Importer already has meaning in the platform. I would rename is KarafPropertyParser or KarafPropertyCsvFactory or something else

extensions/src/org/pentaho/platform/osgi/KarafPropertyImporter.java
```diff
@@ -0,0 +1,120 @@
...
+
+            }
+          }
+        } catch ( Exception e ) {
+          throw new IllegalStateException( getErrorPrefix() );
+        }
+      }
+
+    } catch ( FileNotFoundException e ) {
+      e.printStackTrace();
```
pentaho-nbaker: printStackTrace needs to be replaced with logging. Shouldn't we throw Exceptions here?

extensions/src/org/pentaho/platform/osgi/KarafPropertyImporter.java
```diff
@@ -0,0 +1,120 @@
...
+    } catch ( IOException e ) {
+      e.printStackTrace();
+    } finally {
+      if ( br != null ) {
+        try {
+          br.close();
```
pentaho-nbaker: You can use IOUtils.closeQuietly()

extensions/src/org/pentaho/platform/osgi/KarafPropertyImporter.java
```diff
@@ -0,0 +1,120 @@
...
+
+import org.apache.commons.lang.StringUtils;
+
+public class KarafPropertyImporter {
+  final static String delimiter = ",";
+  protected static KarafInstance karafInstance;
+  private String importFilePath;
+  int lineNumber;
+
+  public KarafPropertyImporter( String importFilePath ) {
```
pentaho-nbaker: I would rather see this rewritten to use RegEx and capture groups within that to extract the values. At least with RegEx I can see the patterns and easily understand how the file should look. All of the split/indexOf/loop code is hard to follow

extensions/src/org/pentaho/platform/web/http/api/resources/KarafResource.java
```diff
@@ -0,0 +1,66 @@
...
+import org.pentaho.platform.osgi.KarafInstance;
+import org.pentaho.platform.osgi.KarafInstancePropertyDto;
+
+/**
+ * This service provides methods related to karaf instances
+ *
+ * @author tkafalas
+ */
+@Path( "/karaf" )
+public class KarafResource extends AbstractJaxRSResource {
```
pentaho-nbaker: Yea want to see this reimagined as a generic service for the platform. Anyone should be able to add a service/port registration to be returned by this service. Karaf would be just one of the systems feeding ports into it. So probably need a ServiceRegistry class and a ServiceResource. Add the ability to add a Service, with name, ID, Description and used ports

pentaho-nbaker:
In general there needs to be more Javadoc

extensions/src/org/pentaho/platform/osgi/KarafInstance.java
```diff
@@ -25,29 +48,60 @@
...
   private static final int MAX_NUMBER_OF_KARAF_INSTANCES = 50;
   private static final int BANNER_WIDTH = 79;
+  private final PortAssigner portAssigner = new PortAssigner();
+  private boolean started;
+  private static KarafInstance instance; // Not final because unit test creates multiple instances
+
   private StringBuilder banner = new StringBuilder();
 
-  public KarafInstance( String root ) {
+  protected KarafInstance( String root ) {
```
pentaho-nbaker: Just leave it package protected, not "protected" so that Unit Tests can call the constructor. Everyone else should be calling getInstance() (KarafBoot seems to be calling the constructor)

extensions/src/org/pentaho/platform/osgi/KarafInstance.java
```diff
@@ -25,29 +48,60 @@
...
-    System.setProperty( "karaf.port", "" + assignedPorts[0] );// was 8081
-    System.setProperty( "org.pentaho.osgi.service.http.port", "" + assignedPorts[1] );// was 8181
-    System.setProperty( "org.pentaho.jmx.rmi.server.port", "" + assignedPorts[2] );// was 44444
-    System.setProperty( "org.pentaho.jmx.rmi.registry.port", "" + assignedPorts[3] );// was 1099
-    System.setProperty( "karaf.data", cachePath );
+    registerProperty( new KarafInstanceProperty( "CACHE_DIR", "karaf.data", "Karaf Cache Directory", cachePath ) );
+  }
+
+  public static KarafInstance getInstance() {
+    if ( instance == null ) {
```
pentaho-nbaker: Create instance if null here

extensions/src/org/pentaho/platform/settings/ServerPortService.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * This serves to group {@link ServerPort} objects under an arbitrary name. It also allows a user friendly description
+ * of the service category to be supplied. Some web services will provide information on ServerPorts grouped by the
+ * {@link ServerPortService} they are under. To fully activate the ServerPortService it must be registered with
+ * {@link ServerPortRegistry#addService(ServerPortService)}
+ * 
+ * @author tkafalas
+ *
+ */
+public class ServerPortService {
```
pentaho-nbaker: This name is really confusing. I would think it's a service returning a ServicePort. Can we just make it Service?

extensions/src/org/pentaho/platform/web/http/api/resources/PortResource.java
```diff
@@ -0,0 +1,73 @@
...
+import org.pentaho.platform.settings.ServerPortService;
+import org.pentaho.platform.settings.ServerPortServiceDto;
+
+/**
+ * This service provides methods related to karaf instances
+ *
+ * @author tkafalas
+ */
+@Path( "/port" )
+public class PortResource extends AbstractJaxRSResource {
```
pentaho-nbaker: Should be a ServiceResource. We're returning a list of services, embedded within that are the ports being used.

PR #2404, opened 2015-08-03
[SP-2041][BISERVER-12524] Regression: Getting create deep link dialog…

SergeyTravin:
… does not work with proxy or when running in port 80


dkincade:
Can this small bit of code be extracted into a protected method and added to a unit test ... should be simple but prevents this issue from recurring...

SergeyTravin:
@dkincade, This small bit of code is included  in GWT class and we do not have any unit tests for gwt classes in pentaho user console project.  It may take a pretty much time than we have for patch release. 

@mbatchelor, could you look on it?

PR #2424, opened 2015-08-12
[BACKLOG-3985] - When downloading an Analysis schema, include the ann…

kurtwalker:
…otated version of the schema if there are saved inline annotations.


kurtwalker:
@bytekast @pminutillo @DFieldFL please review and merge

PR #2443, opened 2015-08-20
[BACKLOG-4724] - Unable to initialize karaf due to xalan transformer …

aliakseihaidukou:
…issue

@rmansoor @CodeOnCoffee please review.

Maybe we need to add some comment to "pentaho.system.properties" file?
You can tell me how to adapt "PentahoSystem.WARN_XML_FACTORIES_LOCATION_CHANGE" message

### DIFF
--- a/core/src/org/pentaho/platform/engine/core/system/PentahoSystem.java
+++ b/core/src/org/pentaho/platform/engine/core/system/PentahoSystem.java
@@ -28,6 +28,7 @@ import org.osgi.framework.BundleContext;
 import org.pentaho.platform.api.engine.IActionParameter;
 import org.pentaho.platform.api.engine.IApplicationContext;
 import org.pentaho.platform.api.engine.ICacheManager;
+import org.pentaho.platform.api.engine.IConfiguration;
 import org.pentaho.platform.api.engine.IContentOutputHandler;
 import org.pentaho.platform.api.engine.ILogger;
 import org.pentaho.platform.api.engine.ILogoutListener;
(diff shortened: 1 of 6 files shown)


rmansoor:
@CodeOnCoffee , @pentaho-nbaker  The code looks ok to me. We do have an interface change. So please give us your feedback.

aliakseihaidukou:
@CodeOnCoffee @pentaho-nbaker Can you look at this PR, please? I'll be on vacation for two weeks and will not be able to make changes.

pentaho-nbaker:
I really see no need to modify existing interfaces.

A new file for System Properties sounds great. Obviously we already have a system.properties for a different purpose which we cannot touch. So java-system-properties.properties or something hinting that these will become JVM System Properties would be nice.

Failing to find the entries in pentaho.xml, I would simply load the new configuration: systemConfig.getConfiguration("java-system-properties").getProperties() which is your Property object you'll need to loop over and call System.setProperties()

core/src/org/pentaho/platform/engine/core/system/PentahoSystem.java
```diff
@@ -506,6 +513,20 @@ private static void initXMLFactories() {
                 xpathToXMLFactoryNodes ) );
           }
         }
+      } else {
```
pentaho-nbaker: This is only getting called if PentahoSystem.systemSettingsService is null. I want the java-system-properties loaded all the time. I would move the following code to another method "setSystemProperties(), called right before initXmlFactories() in PentahoSystem.init().

aliakseihaidukou:
@pentaho-nbaker changed the code according to your suggestions. On vacation already - cannot change anything for two weeks.

pentaho-nbaker:
Got it just in time then!

dkincade:
@pentaho-nbaker @rmansoor @aliakseihaidukou This commit has caused:
39 Unit Test Errors in pentaho-platform-core
255 Unit Test Erros in pentaho-platform-repository
39 Unit Test Errors in pentaho-platform-scheduler

There have also been a number of other downstream project that have started having problems since this pull request was added. Not saying this is the cause, but based on the instability identified above, there is something to investigate.

Wingman server says - let's revert this and then resubmit when we know the impact of the change.

aliakseihaidukou:
@pentaho-nbaker sorry - didn't check all tests locally - there is no ISystemConfig instance in the system for other tests. Looks like we need to check if they exist in PentahoSystem. You can ask someone from EPAM team to fix my issue. I think Aliaksandr Drebenchuk or Kanstantsin can help with it.

PR #2445, opened 2015-08-20
[PDI-13736] - Regression: PDI Server Startup Time with invalid databa…

akhayrutdinov:
…se connections

- do not pre-populate the pool with idle connections unless the property 'dbcp-defaults/pre-populate-pool' is set to 'true'

@mbatchelor, @pentaho-nbaker, review it please. This change turns off pool's pre-populating unless deliberately allowed
/cc @pamval

### DIFF
--- a/core/src/org/pentaho/platform/engine/services/connection/datasource/dbcp/PooledDatasourceHelper.java
+++ b/core/src/org/pentaho/platform/engine/services/connection/datasource/dbcp/PooledDatasourceHelper.java
@@ -45,6 +45,7 @@ import org.pentaho.database.service.IDatabaseDialectService;
 import org.pentaho.platform.api.data.DBDatasourceServiceException;
 import org.pentaho.platform.api.data.IDBDatasourceService;
 import org.pentaho.platform.api.engine.ICacheManager;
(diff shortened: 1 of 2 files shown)


core/src/org/pentaho/platform/engine/services/connection/datasource/dbcp/PooledDatasourceHelper.java
```diff
@@ -277,8 +278,15 @@ public static PoolingDataSource setupPooledDataSource( IDatabaseConnection datab
...
           + " idle connections." ); //$NON-NLS-1$
 
-      for ( int i = 0; i < maxIdleConnection; ++i ) {
-        pool.addObject();
+      String prePopulatePoolStr = PentahoSystem.getSystemSetting( "dbcp-defaults/pre-populate-pool", null );
+      if ( Boolean.parseBoolean( prePopulatePoolStr ) ) {
+        for ( int i = 0; i < maxIdleConnection; ++i ) {
+          pool.addObject();
+        }
+        if ( Logger.getLogLevel() <= ILogger.DEBUG ) {
```
mbatchelor: I believe this line should be:

if ( Logger.isDebugEnabled() ) {

PR #2455, opened 2015-08-21
Revert "[BACKLOG-4724] - Unable to initialize karaf due to xalan tran…

dkincade:
Reverts pentaho/pentaho-platform#2443


dkincade:
Reverting this commit due to large number of unit tests which started failing due to the inclusion of this commit.

PR #2466, opened 2015-08-25
[BACKLOG-4526] - Dev subtask for Create public REST interfaces for Us…

apratkin:
…er Create

### DIFF
--- a/extensions/src/org/pentaho/platform/web/http/api/resources/UserRoleDaoResource.java
+++ b/extensions/src/org/pentaho/platform/web/http/api/resources/UserRoleDaoResource.java
@@ -35,13 +35,13 @@ import org.pentaho.platform.core.mt.Tenant;
 import org.pentaho.platform.engine.core.system.PentahoSessionHolder;
 import org.pentaho.platform.engine.core.system.PentahoSystem;
 import org.pentaho.platform.security.policy.rolebased.IRoleAuthorizationPolicyRoleBindingDao;
-import org.pentaho.platform.security.policy.rolebased.RoleBindingStruct;
 import org.pentaho.platform.security.policy.rolebased.actions.AdministerSecurityAction;
 import org.pentaho.platform.security.policy.rolebased.actions.RepositoryCreateAction;
 import org.pentaho.platform.security.policy.rolebased.actions.RepositoryReadAction;
 import org.pentaho.platform.web.http.api.resources.services.UserRoleDaoService;
 
 import javax.ws.rs.GET;
+import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.WebApplicationException;
@@ -500,6 +500,40 @@ public class UserRoleDaoResource extends AbstractJaxRSResource {
       logger.warn( e.getMessage(), e );
     }
     roleDao.createUser( getTenant( tenantPath ), userName, password, "", new String[0] );
+    return Response.ok().build();
+  }
+  
+  /**
+   * Create new user with specified name and password
+   *
+   * <p>
+   * <b>Example Request:</b><br />
+   * POST pentaho/api/userroledao/user with JSON {"userName": "name", "password": "password"}
+   * </p>
+   *
+   * @param user object with name and password
+   * 
+   * @return Response object containing the status code of the operation
+   */
+  @POST
+  @Path( "/user" )
+  @Consumes( { WILDCARD } )
+  @StatusCodes( { 
+    @ResponseCode( code = 200, condition = "Successfully created new user." ), 
+    @ResponseCode( code = 400, condition = "Provided data has invalid format." ),
+    @ResponseCode( code = 403, condition = "A non administrative user is trying to access this endpoint." ), 
+    @ResponseCode( code = 500, condition = "Unable to create User objects." ) 
+  } )
+  public Response createUserPublicApi( User user ) {
+    try {
+      userRoleDaoService.createUser( user );
+    } catch ( SecurityException e ) {
+      throw new WebApplicationException( Response.Status.FORBIDDEN );
+    } catch ( UserRoleDaoService.ValidationFailedException e ) {
+      throw new WebApplicationException( Response.Status.BAD_REQUEST );
+    } catch ( Exception e ) {
+      throw new WebApplicationException( Response.Status.INTERNAL_SERVER_ERROR );
+    }
     return Response.ok().build();
   }
 
--- a/extensions/src/org/pentaho/platform/web/http/api/resources/services/UserRoleDaoService.java
+++ b/extensions/src/org/pentaho/platform/web/http/api/resources/services/UserRoleDaoService.java
@@ -17,6 +17,7 @@
 
 package org.pentaho.platform.web.http.api.resources.services;
 
+import org.apache.commons.lang3.StringUtils;
 import org.pentaho.platform.api.engine.IAuthorizationPolicy;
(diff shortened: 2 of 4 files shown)


apratkin:
@brendan-lynch could you review?

extensions/src/org/pentaho/platform/web/http/api/resources/UserRoleDaoResource.java
```diff
@@ -485,6 +486,34 @@ public Response createUser( @QueryParam ( "tenant" ) String tenantPath, User use
   }
 
   /**
+   * Create new user with specified name and password
```
brendan-lynch: Comments need to include the response and request examples for using this call

extensions/src/org/pentaho/platform/web/http/api/resources/UserRoleDaoResource.java
```diff
@@ -485,6 +486,34 @@ public Response createUser( @QueryParam ( "tenant" ) String tenantPath, User use
...
+  @POST
+  @Path( "/user" )
+  @Consumes( { WILDCARD } )
+  @StatusCodes( { 
+    @ResponseCode( code = 200, condition = "Successfully created new user." ), 
+    @ResponseCode( code = 403, condition = "A non administrative user is trying to access this endpoint." ), 
+    @ResponseCode( code = 500, condition = "Unable to create User objects." ) 
+  } )
+  public Response createUserPublicApi( User user ) {
+    if ( canAdminister() ) {
```
brendan-lynch: canAdminister logic should exist in the service layer. The service layer can throw a SecurityException in the canAdminister else block, which the catch clause in the resource maps to the web application exception 403 error code.

extensions/src/org/pentaho/platform/web/http/api/resources/services/UserRoleDaoService.java
```diff
@@ -17,15 +17,46 @@
...
+    return nameValid && passValid;
+  }
+
+  public void createUser( User user ) throws UnauthorizedException, Exception {
+    if ( userValid( user ) ) {
+      IUserRoleDao roleDao =
+          PentahoSystem.get( IUserRoleDao.class, "userRoleDaoProxy", PentahoSessionHolder.getSession() );
+      roleDao.createUser( null, user.getUserName(), user.getPassword(), "", new String[0] );
+    } else {
+      throw new ValidationFailedException();
```
brendan-lynch: If we are going to throw a validation exception failure here we should also catch this in the resource layer and map it to a response code, to let the user know that the username contains invalid characters. I like this exception for validation though.

apratkin:
Tested:
1.	Try to create user - ok
2.	Try to create user with name which already exist - 500 response code
3.	Try to create user not by administrator - 403
4.	Try to create user with name on Russian - ok
5.	Try to create user without password (empty) – 400
6.	Try to create user without password (null) – 400
7.	Try to create user with empty name (empty) – 400
8.	Try to create user with empty name (null) – 400
9.	Try to create user with name more than 300 symbols – ok
We do not have limitation on length of name
10.	Try to create user with reserved symbols (/,\,\t,\r,\n) at name– 400

extensions/test-src/org/pentaho/platform/web/http/api/resources/UserRoleDaoResourceTest.java
```diff
@@ -360,4 +362,56 @@ public void testSetLogicalRolesSecurityException() {
...
+  public void testCreateUserValidationFailed() throws Exception {
+    UserRoleDaoService mockService = mock( UserRoleDaoService.class );
+    doThrow( new UserRoleDaoService.ValidationFailedException() ).when( mockService ).createUser( any( User.class ) );
+    UserRoleDaoResource resource =
+        new UserRoleDaoResource( roleBindingDao, tenantManager, systemRoles, adminRole, mockService );
+    try {
+      resource.createUserPublicApi( new User( "\\/validation", "failed" ) );
+    } catch ( WebApplicationException e ) {
+      assertEquals( Response.Status.BAD_REQUEST.getStatusCode(), e.getResponse().getStatus() );
+      return;
```
brendan-lynch: remove the return statement and the assert failed

extensions/test-src/org/pentaho/platform/web/http/api/resources/UserRoleDaoResourceTest.java
```diff
@@ -360,4 +362,56 @@ public void testSetLogicalRolesSecurityException() {
...
+    doThrow( new SecurityException() ).when( mockService ).createUser( any( User.class ) );
+    UserRoleDaoResource resource =
+        new UserRoleDaoResource( roleBindingDao, tenantManager, systemRoles, adminRole, mockService );
+    try {
+      resource.createUserPublicApi( new User( "not", "admin" ) );
+    } catch ( WebApplicationException e ) {
+      assertEquals( Response.Status.FORBIDDEN.getStatusCode(), e.getResponse().getStatus() );
+      return;
+    }
+    Assert.fail();
```
brendan-lynch: remove the return statement and the assert failed

extensions/test-src/org/pentaho/platform/web/http/api/resources/UserRoleDaoResourceTest.java
```diff
@@ -360,4 +362,56 @@ public void testSetLogicalRolesSecurityException() {
...
+    UserRoleDaoService mockService = mock( UserRoleDaoService.class );
+    doThrow( new AlreadyExistsException( "message" ) ).when( mockService ).createUser( any( User.class ) );
+    UserRoleDaoResource resource =
+        new UserRoleDaoResource( roleBindingDao, tenantManager, systemRoles, adminRole, mockService );
+    try {
+      resource.createUserPublicApi( new User( "user", "duplicate" ) );
+    } catch ( WebApplicationException e ) {
+      assertEquals( Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(), e.getResponse().getStatus() );
+      return;
+    }
```
brendan-lynch: Same comment as above

extensions/test-src/org/pentaho/platform/web/http/api/resources/services/UserRoleDaoServiceTest.java
```diff
@@ -523,4 +526,39 @@ public void testGetRoleBindingStructSecurityException() {
...
+    return false;
+  }
+
+  @Test
+  public void testCreateUserValidation() {
+    List<String> reservedChars = Arrays.asList( "/,\\,\t,\r,\n".split( "," ) );
+    IUnifiedRepository repo = mock( IUnifiedRepository.class );
+    doReturn( reservedChars ).when( repo ).getReservedChars();
+    PentahoSystem.registerObject( repo );
+    UserRoleDaoService service = new UserRoleDaoService(){
```
brendan-lynch: See my unit test examples, we should not be mocking this, the code in private methods should be exercised through the call to the method that accesses them. Mock the objects that PentahoSystem.get() returns.

extensions/src/org/pentaho/platform/web/http/api/resources/services/UserRoleDaoService.java
```diff
@@ -170,7 +202,7 @@ public void setLogicalRoles( LogicalRoleAssignments roleAssignments ) throws Sec
     }
   }
 
-  private boolean canAdminister() {
+  protected boolean canAdminister() {
```
brendan-lynch: This should continue to be a private method, there is no business exposing this to the package just for unit testing when we have viable alternatives using pentaho system.

PR #2468, opened 2015-08-26
[BACKLOG-4382] - Create public REST interfaces for Operation Permissions

brendan-lynch:
Exposed the REST api for operation permissions, to add permissions and remove them from existing role. Added unit tests to the REST and service layers.

### DIFF
--- a/extensions/src/org/pentaho/platform/web/http/api/resources/services/UserRoleDaoService.java
+++ b/extensions/src/org/pentaho/platform/web/http/api/resources/services/UserRoleDaoService.java
@@ -18,30 +18,36 @@
 package org.pentaho.platform.web.http.api.resources.services;
 
 import org.pentaho.platform.api.engine.IAuthorizationPolicy;
-import org.pentaho.platform.api.engine.IPentahoSession;
 import org.pentaho.platform.api.engine.security.userroledao.IPentahoRole;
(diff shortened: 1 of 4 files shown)


extensions/src/org/pentaho/platform/web/http/api/resources/services/UserRoleDaoService.java
```diff
@@ -134,11 +140,49 @@ public void deleteRoles( String roleNames ) throws SecurityException, Uncategori
     }
   }
 
+  public SystemRolesMap getRoleBindingStruct( String locale ) throws SecurityException {
+    if ( canAdminister() ) {
+      RoleBindingStruct roleBindingStruct = getRoleBindingDao().getRoleBindingStruct( locale );
+      SystemRolesMap systemRolesMap = new SystemRolesMap();
+      for ( Map.Entry<String, String> localalizeNameEntry : roleBindingStruct.logicalRoleNameMap.entrySet() ) {
```
rfellows: can ```getRoleBindingDao().getRoleBindingStruct( locale );``` ever return null? if so, you might want to guard against that given an invalid/unsupported locale causing an NPE

brendan-lynch: This will return the system default locale, if it cannot find the locale, it also handles the null case.

PR #2469, opened 2015-08-27
[SP-2051] - Backport of BISERVER-12637 - Jackrabbit's ds_repos_datastore table can grow out of control causing performance issues (5.4 Suite)

akhayrutdinov:
- prohibit returning false except the case when Scheduler has not been found
- add test case

@mbatchelor, @pentaho-nbaker, @rmansoor, review it please.

While preparing an overview for Sandra, I realize I don't like the current behaviour. Now, if something goes wrong, the listener ruins the system as well. However, this is not an essential listener, and the platform can work w/o it.

That's why I made this PR. The only case, when the listener still returns false, is when Scheduler cannot be obtained.


akhayrutdinov:
@mbatchelor, @pentaho-nbaker, @rmansoor, review it please.

PR #2481, opened 2015-08-31
[BACKLOG-4456] - Implement the Users & Roles portion of the Export pr…

brendan-lynch:
…ocess

Added User and Role information to the export manifest.

### DIFF
--- a/extensions/test-src/org/pentaho/platform/plugin/services/exporter/PentahoPlatformExporterTest.java
+++ b/extensions/test-src/org/pentaho/platform/plugin/services/exporter/PentahoPlatformExporterTest.java
@@ -1,12 +1,19 @@
 package org.pentaho.platform.plugin.services.exporter;
 
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
+
 import org.pentaho.database.model.DatabaseConnection;
 import org.pentaho.database.model.IDatabaseConnection;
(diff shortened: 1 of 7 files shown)


extensions/test-src/org/pentaho/platform/plugin/services/exporter/PentahoPlatformExporterTest.java
```diff
@@ -42,9 +57,16 @@ public void setUp() throws Exception {
     scheduler = mock( IScheduler.class );
     session = mock( IPentahoSession.class );
     PentahoSessionHolder.setSession( session );
-    exporter = spy( new PentahoPlatformExporter( repo ) );
-    exporter.setScheduler( scheduler );
+    exporterSpy = spy( new PentahoPlatformExporter( repo ) );
+    exporterSpy.setScheduler( scheduler );
     doReturn( "session name" ).when( session ).getName();
+    IUnifiedRepository repository = mock( IUnifiedRepository.class );
```
rfellows: why have duplicate objects like exporter & exporterSpy and repo and repository? looks like they are configured the same way.

brendan-lynch:
Good catch, I removed the repository, I didn't the duplicate. The other exporter is a spy object the one I am using is a plain java object. I didn't want to remove the spy from the class, but I didn't want to use it either.

PR #2488, opened 2015-09-01
[MONDRIAN-2400] Removes the one-flush-per-server policy. The XmlaExtr…

lucboudreau:
…a will now flush only the schema associated to a given connection, so we must call it for each hosted schema known to the olap service.

This is a sibling of PR https://github.com/pentaho/mondrian/pull/538


lucboudreau:
@mkambol Please review.

PR #2506, opened 2015-09-08
BACKLOG-3121 -- Save As dialog should not show when using Save [only]…

kolinus:
… on a

report that was already saved overwriting an existing report

What was done:
1) added storing file info to tab context if it was overwritten


kolinus:
@rmansoor  could you please review
no unit test as this is gwt UI change

kolinus:
@mdamour1976 @pentaho-nbaker  could you please review

PR #2507, opened 2015-09-08
[BACKLOG-5073] - Create Command Line interface for full backup

brendan-lynch:
Create command line script option for running the system backup


extensions/src/org/pentaho/platform/plugin/services/importexport/CommandLineProcessor.java
```diff
@@ -244,6 +254,12 @@ public static void main( String[] args ) throws Exception {
         case REST:
           commandLineProcessor.performREST();
           break;
+        case BACKUP:
+          commandLineProcessor.performBackup();
+          break;
+        case RESTORE:
+          //commandLineProcessor.performREST();
```
rfellows: no import yet?  assume this is just a placeholder, right?

PR #2508, opened 2015-09-08
[BACKLOG-4454] - Encrypting user's password only if it is a plain text password

rmansoor:


rmansoor:
Unit test is using a MockUserRoleDao, so can't test this logic test.

rmansoor:
@dkincade , @rfellows , @pentaho-nbaker @CodeOnCoffee please review

rfellows:
looks good to me. should we wait for @pentaho-nbaker put his 2 cents in?

rfellows:
the only thing that comes to mind is more of a future thing that we really can't do much about...

if we were in a completely OSGi environment, we couldn't provide classes from the same package from 2 different jars without issues. neither jar involved now are osgi bundles in our karaf setup... but in the future we might need this "fixed" in jackrabbit rather than in our repository jar.

PR #2514, opened 2015-09-09
[BACKLOG-5087] Updated the start scripts for the server to be the new…

dkincade:
… defaults for prodution deployments of min:2G max:6G


dkincade:
@rmansoor @mbatchelor @pentaho-nbaker  please review...

PR #2515, opened 2015-09-10
[BACKLOG-4446] - Handling inline-modeling annotations on backup/restore of mondrian schemas.

rfellows:
on export: generate the ExportManifestMondrian with a new, optional element for a reference to the annotations file in the zip for this mondrian schema.

on import: drop the annotations file into the same folder in /etc/mondrian/<catalog-name>/ as the schema.xml file. One caveat, if the annotations file is imported first, the catalog folder is hidden by default. Had to make a change on the addHostedCatalog method of MondrianCatalogRepositoryHelper to set that to not be hidden when the schema file is added. Otherwise, the model will not appear in PUC.


rfellows:
@rmansoor, @pentaho-nbaker please review
@brendan-lynch have a look when you get back

PR #2517, opened 2015-09-10
[BACKLOG-3124] Blockstatus method can't compare dates in proper way

mdamour1976:


mdamour1976:
@e-cuellar review please

PR #2518, opened 2015-09-11
[BACKLOG-5086] - data source wizard is showing incorrect display on C…

aliakseihaidukou:
…hrome and IE

@rmansoor please review


dkincade:
@rmansoor @pentaho-nbaker please review ...

pentaho-nbaker:
This isn't a good fix. The order of stylesheets is important. Latter stylesheets can override styles provided by previous stylesheets with the same selectors. This code will effectively reverse the order of the stylesheets, potentially causing different cascade behavior.

PR #2520, opened 2015-09-11
[BACKLOG-4445] - Implementation of exporting the metastore as part of backup/restore

rfellows:
[BACKLOG-4445] - Implementation of exporting the metastore as part of backup/restore

Overview:
- get the metastore from the pur respoitory
- convert it to an xml metastore
- zip it up
- add that zip to the export zip as metastore.zip
- add a new type in the export manifest to support this.


rfellows:
@brendan-lynch please review
@mattyb149 have a look, as it's related to https://github.com/pentaho/metastore/pull/35 (one you just merged)

extensions/src/org/pentaho/platform/plugin/services/importexport/exportManifest/bindings/ExportManifestMetaStore.java
```diff
@@ -0,0 +1,81 @@
+package org.pentaho.platform.plugin.services.importexport.exportManifest.bindings;
```
mattyb149: Needs a header

PR #2523, opened 2015-09-12
[BACKLOG-4445] - Removing the extensions dependency on pur-repository…

rfellows:
…-base-plugin. This was resulting in a circular dependency. So, the code we tried to re-use has been copied into a utility class in extensions for usage.

  Also, added back the dependency of pur-repository-base-plugin in the assembly, where it was originally.


rfellows:
@lgrill-pentaho want to merge this? this puts the dependencies back to how they were and doesn't require a complete revert. i copied the code over that we were trying to re-use.

PR #2524, opened 2015-09-14
[BACKLOG-4156] - hierarchy table was defined twice, which will cause …

kurtwalker:
…schema workbench to consider it an invalid schema.


kurtwalker:
@mkambol @lucboudreau please review

PR #2526, opened 2015-09-14
[BACKLOG-4449] -Implement the MetaStore portion of the Import process

rfellows:
Added a new import handler for metastore import (supports mime type "application/vnd.pentaho.metastore" and extension is mzip (just a regular zip file however, but different to allow for another handler other than the SolutionImportHandler).

Expects the mzip file to contain the zipped up contents of an XmlMetaStore. This is extracted to a temp location on the fs to allow it to be loaded in as an XmlMetaStore before walking the structure and copying it to the repository meta store.


rfellows:
depends on https://github.com/pentaho/pentaho-kettle/pull/1758

PR #2532, opened 2015-09-15
BACKLOG-4822 -- /pentaho/api/userrolelist/allroles API result doesn't…

kolinus:
… show

Administrator role

What was done:
1) fixed backlog-4922 and 4821 cause they should be fixed in the same
method and it should not affect app. The check of already included roles
was added

### DIFF
--- a/extensions/src/org/pentaho/platform/web/http/api/resources/services/UserRoleListService.java
+++ b/extensions/src/org/pentaho/platform/web/http/api/resources/services/UserRoleListService.java
@@ -74,7 +74,22 @@ public class UserRoleListService {
 
   public RoleListWrapper getAllRoles() {
     List<String> roles = getUserRoleListService().getAllRoles();
-    roles.addAll( getExtraRoles() );
+    List<String> extraRoles = getExtraRoles();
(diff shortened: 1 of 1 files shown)


kolinus:
@brendan-lynch  @rmansoor  could you please review

extensions/src/org/pentaho/platform/web/http/api/resources/services/UserRoleListService.java
```diff
@@ -74,7 +74,22 @@ public RoleListWrapper getRoles() {
...
+    if ( extraRoles != null ) {
+      if ( systemRoles != null ) {
+        extraRoles.addAll( systemRoles );
+      }
+    }
+    else {
+      extraRoles = systemRoles;
+    }
+    if ( extraRoles != null ) {
+      for ( String ex : extraRoles ) {
```
mbatchelor: This would be way faster using a HashSet. Especially if the number of roles is large (like for instance, a large LDAP installation). I strongly suggest re-working this to not use contains and a list, and instead, use a HashSet.

brendan-lynch:
+1 Marc's suggestion here

PR #2534, opened 2015-09-15
Fixes

rfellows:
Unit test failure fix

Also a fix for a scenario when doing multiple restores (different import files) without restarting the server. basically got a cached set of the metastore xml folder (from the first load).


rfellows:
@brendan-lynch please review

PR #2538, opened 2015-09-15
[BACKLOG-4328] - Adding the PentahoBeanScopeProcessor to the plugin manager

rmansoor:
This is a same changes we made to the PentahoSystemPluginManager https://github.com/pentaho/pentaho-platform/blob/master/extensions/src/org/pentaho/platform/plugin/services/pluginmgr/PentahoSystemPluginManager.java#L715


rmansoor:
This change is necessary to use the DefaultPluginManager after the 6.0 upgrade
@pentaho-nbaker @dkincade please review

PR #2557, opened 2015-09-17
[BACKLOG-4444] - Adding an interface to extend the IUserSettingService

rfellows:
This is to support working with user settings for other users than the one on the active session. Basically for backup/restore initially, but could be useful in some admin tool down the road.

The plan will be for this interface to get folded into the IUserSettingService in the future, probably at the next major release. But we can't change that interface now.


rfellows:
@pentaho-nbaker and @brendan-lynch please review

brendan-lynch:
Looks good to me

PR #2564, opened 2015-09-18
[BACKLOG-4443] - Removing duplicate definition in the subclass of a m…

rfellows:
…ember variable that was resulting in NPE when trying to export/download content from PUC


rfellows:
@dkincade please review. this is a cherry pick of the same fix in master (https://github.com/pentaho/pentaho-platform/pull/2563)

PR #2568, opened 2015-09-21
[BACKLOG-4452] - removing the hard-coded boolean (true) form the Solu…

rfellows:
…tionImportHandler for mondrian and metadata data sources. they now obey the parent bundle's setting for overwrite.

Also fixed 2 failing unit tests


rfellows:
@brendan-lynch please review. 6.0 PR to follow (post devci build success)

PR #2570, opened 2015-09-21
[BISERVER-12787] Restoring fall-back behavior in <pen:bean> resolutio…

pentaho-nbaker:
…n. Now checks PentahoSystem.get() to find beans matching the simplename of the class.

This is a large regression in <pen:bean> behavior. Tests have been added to verify the behavior which regressed.


pentaho-nbaker:
@dkincade please review

PR #2580, opened 2015-09-23
[BACKLOG-5217] - Adding a fallback to the datasource lookup logic to using the datsource as it is

rmansoor:


rmansoor:
@CodeOnCoffee , @pentaho-nbaker , @dkincade please review

PR #2582, opened 2015-09-23
[BACKLOG-4813] - Documentation edits and test case update.

brendan-lynch:
The documentation has been cleaned up around the errors found by QA. Since a mapping of return code to exception has been changed to match the documentation the unit test had to be updated as well.


brendan-lynch:
@dkincade @rfellows -- Rob I decided to bring this change over for 6.0 so that the documentation updates persist, in case they are regenerated off the source branch at some later date.

brendan-lynch:
#2571

PR #2583, opened 2015-09-23
[BACKLOG-5291] - Restore of schedules assigns all schedules to admin,…

rfellows:
… not the original owner.


rfellows:
@brendan-lynch please review

dkincade:
Would it be possible to add some unit testing to this work?

rfellows:
sure. there aren't any tests at all for SchedulerOutputPathResolver. I can add tests for the method i changed

and the other has tests (integration tests) that still pass.

PR #2585, opened 2015-09-23
[BACKLOG-5291] [TEST] - Adding unit tests for the changes made in Sch…

rfellows:
…edulerOutputPathResolver. Also added some null safe-guarding.


rfellows:
@dkincade here are the unit tests you asked for.

I found a scenario where the output path could be null (thanks unit tests). I added a guard against it... so, i assume you want a cherry-pick of this into 6.0.0.0 so it doesn't have that potential issue, right?

rfellows:
@brendan-lynch please review

dkincade:
That is correct. Let's reivew this commit and get it into master tonight/tomorrow and then get it over to 6.0.0 before tomorrow's builds (like the 11am build)

PR #2590, opened 2015-09-24
[BISERVER-12676] Changes to getTree call is expensive. Re-implementin…

pentaho-nbaker:
…g to improve performance.

Also improving performance of the FileProxy.


api/src/org/pentaho/platform/api/repository2/unified/RepositoryFileTree.java
```diff
@@ -89,7 +89,11 @@ public RepositoryFile getFile() {
    * @return
    */
   public List<RepositoryFileTree> getChildren() {
```
pentaho-nbaker: Children changes made to accommodate different getTree algorithm which builds from the top-down where the children aren't known when the parent is created, instead of the bottom-up as the old method did.

PR #2591, opened 2015-09-24
[BISERVER-12676] Changes to getTree call is expensive. Re-implementin…

pentaho-nbaker:
…g to improve performance.

Also improving performance of the FileProxy.


dkincade:
http://devci.pentaho.com/job/6.0-snapshot/job/platform-base/155/

PR #2592, opened 2015-09-24
[BACKLOG-5313] - After restoring DI Server, I get an error when openi…

rfellows:
…ng my transformation and I can't create any database connections.


rfellows:
@brendan-lynch please review

PR #2597, opened 2015-09-25
[BISERVER-12264] (fix. unify filenames encoding) Issues when exportin…

ivanpogodin:
…g or importing a directory with a blank / space character
@dkincade , @pamval , review it please
no additional tests yet
checked manually only


dkincade:
@rfellows @brendan-lynch please review the master pull request for this (we have a history of regressions working on this item)

brendan-lynch:
Lets test backup / restore on this work before merging it into 6.0.0, will update this PR with the results.

dkincade:
@brendan-lynch @rfellows any feedback yet?

brendan-lynch:
So far we have found a few issues around this, if we do an import from browse files, even around a successful import we get an error message logged. Did you see this during the import testing? We are trying to determine if this is the commit that caused this behavior or of the behavior existed prior to this fix, but if the error existed previous to this, you would have seen it when testing the import, so why wasn't this raised up?

PR #2599, opened 2015-09-25
[BISERVER-12676] Fixing regression. Sorting children.

pentaho-nbaker:

PR #2600, opened 2015-09-25
[BISERVER-12676] Fixing regression. Sorting children.

pentaho-nbaker:


dkincade:
master: https://github.com/pentaho/pentaho-platform/pull/2599
devci: http://devci.pentaho.com/job/6.0-snapshot/job/platform-base/160/

PR #2603, opened 2015-09-25
[BACKLOG-5337] -  Fixed the logic to enable/disable the user role ui based on whether user a configured for pentaho security or not

rmansoor:


rmansoor:
@dkincade I found this issue during some of my testing with LDAP and other security providers. If we try to use something other than ldap, jdbc the user role panel remains active. I will leave this up to you to make a decision about including this in 6.0 or not.

PR #2611, opened 2015-09-28
[BISERVER-12637] - Ensuring all versions are purged when a file is de…

pentaho-nbaker:
…leted. This will ensure that the datastore references are cleaned on the next GC

PR #2612, opened 2015-09-28
[BISERVER-12637] - Ensuring all versions are purged when a file is de…

pentaho-nbaker:
…leted. This will ensure that the datastore references are cleaned on the next GC


dkincade:
master: https://github.com/pentaho/pentaho-platform/pull/2611
devci: http://devci.pentaho.com/job/6.0-snapshot/job/platform-base/165/

PR #2614, opened 2015-09-28
[BISERVER-12637] Ensuring all files within a folder have their histor…

pentaho-nbaker:
…y deleted when folder removed.

Previous fix only addressed files, this one handles when a folder is deleted from the trash, ensuring all files within have their VersionHistory removed


dkincade:
master: https://github.com/pentaho/pentaho-platform/pull/2613

PR #2620, opened 2015-09-29
[TEST] Fix OSGI Object Factory unit test

tkafalas:


rmansoor:
@dkincade I have validated the units test on master

PR #2624, opened 2015-09-30
[TEST] - restoring sampledata.properties to the proper content. it wa…

rfellows:
…s accidentally checked in post-unit-test where it gets modified.


rfellows:
@tkafalas please review

PR #2626, opened 2015-10-01
[BISERVER-12657] - SpringSecurityPrincipalProvider uses static maps instead of CacheManager

akhayrutdinov:
- move internal maps to a global cache
- add tests

@pentaho-nbaker, @rmansoor, review it please.
Please do not ignore it to let it be proceeded within October SP.

/cc @pamval

### DIFF
--- a/repository/src/org/pentaho/platform/repository2/unified/jcr/jackrabbit/security/SpringSecurityPrincipalProvider.java
+++ b/repository/src/org/pentaho/platform/repository2/unified/jcr/jackrabbit/security/SpringSecurityPrincipalProvider.java
@@ -40,10 +40,8 @@ import org.apache.jackrabbit.core.security.principal.AdminPrincipal;
 import org.apache.jackrabbit.core.security.principal.EveryonePrincipal;
 import org.apache.jackrabbit.core.security.principal.PrincipalIteratorAdapter;
 import org.apache.jackrabbit.core.security.principal.PrincipalProvider;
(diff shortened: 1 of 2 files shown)


akhayrutdinov:
@pentaho-nbaker, @rmansoor, review it please.

repository/src/org/pentaho/platform/repository2/unified/jcr/jackrabbit/security/SpringSecurityPrincipalProvider.java
```diff
@@ -115,9 +118,35 @@
...
+  private LRUMap getUserCache() {
+    return getCache( USER_CACHE, 4096 );
+  }
+
+  void setCacheManager( ICacheManager cacheManager ) {
+    this.cacheManager = cacheManager;
+  }
 
-  private final LRUMap roleCache = new LRUMap( 512 );
+  private LRUMap getCache( String region, int capacity ) {
```
pentaho-nbaker: This is not how you use/create regions. In your code the region is simply a "key" into the global cache.

Also instead of simply storing the LRUMap in the cache. create a region and then use that region as the map: putInRegionCache(REGION, Username, User), etc.

pentaho-nbaker:
Additionally when you abandon the LRU cache you should supply region definitions in the ehcache.xml specifying the capacity. This gets it out of code and into someplace where the customer can modify.

PR #2629, opened 2015-10-02
[BACKLOG-5394] - Split UT/IT for project: pentaho-platform-extensions

rfellows:
Moving tests that extend BaseTest or use MicroPlatform to integration tests since that is what they really are.


rfellows:
@dkincade or @brendan-lynch please review

PR #2630, opened 2015-10-02
[BISERVER-12657] - SpringSecurityPrincipalProvider uses static maps instead of CacheManager

akhayrutdinov:
- re-implement the previous commit with proper usage of cache regions
- add default settings to ehcache.xml

@pentaho-nbaker, review it please.
If this is needed to be backported, then the regions will be created programmatically
/cc @mbatchelor, @pamval


pentaho-nbaker:
Thanks, much better.

PR #2633, opened 2015-10-05
Biserver 12639

SergeyTravin:
XActionUtil  delete file which generated during executing


SergeyTravin:
@pminutillo, @rmansoor, @mbatchelor could you review it please.

PR #2644, opened 2015-10-08
[BACKLOG-5394] - Found another unit test that should be an integration test, moved it

rfellows:
[BACKLOG-5394] - Found another unit test that should be an integration test, moved it


rfellows:
@dkincade please review

PR #2645, opened 2015-10-08
[TEST] - Fix issue causing integration test failure on build machines…

rfellows:
… due to folder naming difference for the root project (platform-base vs. pentaho-platform)


rfellows:
@dkincade please review. Turns out this integration test was failing out on cerberus 6.0 (the one that publishes to sonar). Well technically it was failing as a unit test (before i sent you the PR to move it to an IT). We aren't getting reports of integration tests on master from anywhere, but this passed locally before  since I didn't change my github project folder that i cloned into like has happened on the build machine. This should fix the failure.

PR #2646, opened 2015-10-08
[TEST] - Fix issue causing integration test failure on build machines…

rfellows:
… due to folder naming difference for the root project (platform-base vs. pentaho-platform)

refer to https://github.com/pentaho/pentaho-platform/pull/2645


rfellows:
@dkincade please review

PR #2654, opened 2015-10-12
[BISERVER-12758] - DefaultUnifiedRepository.getFileById() throws exception if none was found

akhayrutdinov:
- catch JCR's exception and return null instead
- add test
- update java docs
- fix checkstyle violations for affected classes

@pentaho-nbaker, @rmansoor, review it please.


mbatchelor:
The catch swallows the exception (line 391) - it should minimally log the exception or this will end up being another bug we cannot track down.

PR #2655, opened 2015-10-14
[BISERVER-11565] - File upload names fail if the name has double quotes

akhayrutdinov:
- introduce new endpoint '/repo/publish/file' which expects path to be encoded
- deprecate existing and fix their javadocs
- add tests
- update jersey-related artifacts in test condiguration to 1.16 that is used in compile configuration

@pentaho-nbaker, @mbatchelor, @pamval, review it please.
In addition to the tests (where the actual storing in repository is not checked), I verified the case locally, after building my own platform and my own report-designer. It became possible to publish files with quotes

### DIFF
--- a/extensions/ivy.xml
+++ b/extensions/ivy.xml
@@ -379,11 +379,11 @@
     </dependency>
     <!-- end security test dependencies -->
 
-    <dependency org="com.sun.jersey.jersey-test-framework" name="jersey-test-framework-core" rev="1.5"
+    <dependency org="com.sun.jersey.jersey-test-framework" name="jersey-test-framework-core" rev="1.16"
                 conf="test->default"/>
-    <dependency org="com.sun.jersey.jersey-test-framework" name="jersey-test-framework-grizzly" rev="1.5"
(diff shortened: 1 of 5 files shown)


akhayrutdinov:
@pentaho-nbaker, @mbatchelor, @pamval, review it please.

akhayrutdinov:
@pentaho-nbaker, @mbatchelor, @pamval, review it please.

akhayrutdinov:
@pentaho-nbaker, @mbatchelor, @pamval, @rmansoor, review it please.

rmansoor:
@akhayrutdinov we call publish from the design tools like report designer , medata-editor . Based on your change we have to change those as well correct ?

akhayrutdinov:
@rmansoor, yes, initially the problem was reported for inability to do publish from PRD.
And we *should* change them to use new endpoint, however, we *are not forced*, because old method remains and works as usual.

rmansoor:
@akhayrutdinov the code looks ok and unit tests are also passing, processing the pull request now

PR #2663, opened 2015-10-22
[SP-2195] - Backport of BISERVER-12758 - DefaultUnifiedRepository.getFileById() throws exception if none was found (5.4 Suite)

akhayrutdinov:
- change getFileById() handling due to new behaviour

@pamval, review it please

### DIFF
--- a/repository/src/org/pentaho/platform/repository2/unified/jcr/JcrAclNodeHelper.java
+++ b/repository/src/org/pentaho/platform/repository2/unified/jcr/JcrAclNodeHelper.java
@@ -89,11 +89,8 @@ public class JcrAclNodeHelper implements IAclNodeHelper {
     // Obtain a reference to ACL node as "system", guaranteed access
     final RepositoryFile aclNode = getAclNode( repositoryFile );
 
-    // Check to see if user has READ access to file, this will throw and exception if not.
-    try {
-      unifiedRepository.getFileById( aclNode.getId() );
-    } catch ( Exception e ) {
-      logger.warn( "Error checking access for file", e );
+    // Check to see if user has READ access to file, this will return null if not.
+    if ( unifiedRepository.getFileById( aclNode.getId() ) == null ) {
       return false;
     }
 


bkemper-pentaho:
Unit tests or some other form of validation that this change didn't cause any other issues?

Also - have you verified that the callers can handle an exception being thrown here - since you're now throwing an exception rather than simply returning false



Bob Kemper
Executive Vice President of Engineering
Pentaho, a Hitachi Data Systems company.
5950 Hazeltine National Dr.
Suite 340
Orlando, FL 32822
(407)-965-4019

[cid:92493001-7910-48B4-8380-C9903CB3A049]

akhayrutdinov:
@bkemper-pentaho, 

Any additional tests are not needed - we have a failed test. That means the test executes the affected line.

As for other usages: I've reviewed all usages of the method within ```platform``` and found no suspicious places. If the invocation is not surrounded with try-catch, nothing new could happen with it. If it is, then it is followed by ```null``` checks (see JcrBackedDatasourceMgmtService as example).

bkemper-pentaho:
Hmm -   not sure why you're saying "no other tests are needed".    This is a behavior change with a back port - I don't know what you mean by "we have a failed test" - was it failing previously?   Is this a correction of a prior regression?

Also your statement that "if the invocation is not surrounded by try-catch nothing new could happen" doesn't make sense to me.    We used to log a warning and return false (or so it appears from the code) - now we throw an exception.  So if the caller did :

If (!canAccess(file, perms)) {
// do something about the access...
}

We're now going to throw an uncaught exception since the caller wasn't expecting this behavior.
VERY specifically - if you look at MondrianCatalogHelper we used to check the hasAccess (for false) and throw an "Access denied error".   Now we'll throw a different exception, not log the warning that we used to,  and generally just totally change the behavior.

The defect in question was that we throw an exception rather than return null if a file doesn't exist.   You changed the whole behavior of Access permissions to throw exceptions which used to just return false.  That's clearly a behavior change involved here which can impact callers.

So rather than fix the reported bug (throwing an exception rather than returning null) - you changed low level system behavior.   I'm trying to understand thy that would be the case....





Bob Kemper
Executive Vice President of Engineering
Pentaho, a Hitachi Data Systems company.
5950 Hazeltine National Dr.
Suite 340
Orlando, FL 32822
(407)-965-4019

[cid:FE0FB05E-7AB3-4181-8BEA-1480E3344A8A]

PR #2667, opened 2015-10-22
[BISERVER-12758] - DefaultUnifiedRepository.getFileById() throws exception if none was found

akhayrutdinov:
- change getFileById() handling so, that it is able to cope with both exception and null

@pamval, @bkemper-pentaho, review it please.


bkemper-pentaho:
Again - I'd like to see unit tests to cover both the new and old scenarios - especially on master....

PR #2694, opened 2015-11-17
[PDI-14485] Model published from Spoon - the Datasource appears for P…

ivanpogodin:
…IR but generates an error when trying to be used
@tmorgner , review it please
Bug is fixed as required table SALES_DATA is ported from the Kettle sample data.
It is not full synchronization.
See also http://jira.pentaho.com/browse/PDI-14684


ivanpogodin:
@tmorgner , could you review it, please

tmorgner:
I am not sure about that change. The BI-Server relies on the sample data as well, so in the worst case it requires to update our models there and in the PRD samples. Calling @pentaho-nbaker to chime in.

ivanpogodin:
@pentaho-nbaker , @mdamour1976 , @rmansoor , @e-cuellar , could you review it?

pentaho-nbaker:
I don't see a need to add this table to the models if all we want is to allow Kettle samples to run on the server.

PR #2697, opened 2015-11-19
PentahoSystem JUnit test

AliaksandrDrebenchuk:
PentahoSystem refactored
PentahoSystemTest extended

### DIFF
--- a/core/src/org/pentaho/platform/engine/core/system/PentahoSystem.java
+++ b/core/src/org/pentaho/platform/engine/core/system/PentahoSystem.java
@@ -18,20 +18,7 @@
 
 package org.pentaho.platform.engine.core.system;
 
-import java.io.IOException;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Properties;
-import java.util.StringTokenizer;
-import java.util.concurrent.Callable;
-
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.collections.list.UnmodifiableList;
 import org.apache.commons.lang.StringUtils;
 import org.dom4j.Document;
@@ -55,10 +42,6 @@ import org.pentaho.platform.api.engine.IPentahoRegistrableObjectFactory;
 import org.pentaho.platform.api.engine.IPentahoSession;
 import org.pentaho.platform.api.engine.IPentahoSystemListener;
 import org.pentaho.platform.api.engine.IPentahoUrlFactory;
-import org.pentaho.platform.api.engine.IPlatformPlugin;
-import org.pentaho.platform.api.engine.IPlatformReadyListener;
-import org.pentaho.platform.api.engine.IPluginManager;
-import org.pentaho.platform.api.engine.IPluginProvider;
 import org.pentaho.platform.api.engine.IRuntimeContext;
 import org.pentaho.platform.api.engine.IServerStatusProvider;
 import org.pentaho.platform.api.engine.ISessionStartupAction;
@@ -67,7 +50,6 @@ import org.pentaho.platform.api.engine.ISystemConfig;
 import org.pentaho.platform.api.engine.ISystemSettings;
 import org.pentaho.platform.api.engine.ObjectFactoryException;
 import org.pentaho.platform.api.engine.PentahoSystemException;
-import org.pentaho.platform.api.engine.PlatformPluginRegistrationException;
 import org.pentaho.platform.engine.core.messages.Messages;
 import org.pentaho.platform.engine.core.output.SimpleOutputHandler;
 import org.pentaho.platform.engine.core.solution.PentahoSessionParameterProvider;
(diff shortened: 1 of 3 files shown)


core/src/org/pentaho/platform/engine/core/system/PentahoSystem.java
```diff
@@ -252,18 +265,11 @@ public static boolean init( final IApplicationContext pApplicationContext, final
 
     PentahoSystem.applicationContext = pApplicationContext;
 
-    if ( debug ) {
-      Logger.debug( PentahoSystem.class, "Setting property path" ); //$NON-NLS-1$
-    }
+    logDebug("Setting property path" ); //$NON-NLS-1$
     System.setProperty( "pentaho.solutionpath", "solution:" ); //$NON-NLS-1$
-    if ( LocaleHelper.getLocale() == null ) {
-      LocaleHelper.setLocale( Locale.getDefault() );
```
lucboudreau: Why remove this?

lucboudreau:
There are a lot of changes in this PR. This is not a trivial change. Please make sure to ping a couple people for a throughout review.

rmansoor:
@CodeOnCoffee , @pentaho-nbaker  can you take a look at this pull request

core/src/org/pentaho/platform/engine/core/system/PentahoSystem.java
AliaksandrDrebenchuk: Just look at  LocaleHelper.getLocale() method - it can't return NULL.

core/src/org/pentaho/platform/engine/core/system/PentahoSystem.java
```diff
@@ -549,72 +532,47 @@ public static int getInitializedStatus() {
...
+
+    getFailureMessages( rtn, PentahoSystem.SYSTEM_OTHER_FAILED, "USER_INITIALIZATION_SYSTEM_OTHER_FAILED" );//$NON-NLS-1$
+
     return rtn;
   }
 
-  public static void addInitializationFailureMessage( final int failureBit, final String message ) {
-    Integer i = new Integer( failureBit );
-    List l = (List) PentahoSystem.initializationFailureDetailsMap.get( i );
+  public static void addInitializationFailureMessage( final Integer failureBit, final String message ) {
```
mbatchelor: This changes the signature of a public static method. Don't think this is at all necessary (seems a refactor for the sake of refactoring here).

core/src/org/pentaho/platform/engine/core/system/PentahoSystem.java
```diff
@@ -867,98 +825,79 @@ public static Object putInGlobalAttributesMap( final Object key, final Object va
...
+    return parameterProviderMap;
+  }
+  
   public static void sessionStartup( final IPentahoSession session, IParameterProvider sessionParameters ) {
-
     List<ISessionStartupAction> sessionStartupActions =
         PentahoSystem.getSessionStartupActionsForType( session.getClass().getName() );
-    if ( sessionStartupActions == null ) {
-      // nothing to do...
+    if ( CollectionUtils.isEmpty( sessionStartupActions ) || !session.isAuthenticated() ) {
```
mbatchelor: Why are these changes being made for something that's supposed to be increasing test coverage ? This level of refactor needs to be done on a JIRA case.

core/src/org/pentaho/platform/engine/core/system/PentahoSystem.java
```diff
@@ -1161,27 +1089,23 @@ public static String publish( final IPentahoSession session, final String classN
   }
 
   // FIXME: should be named getAdministrationPlugins
-  public static List getPublisherList() {
-    return new ArrayList( PentahoSystem.administrationPlugins );
+  public static List<IPentahoPublisher> getPublisherList() {
```
mbatchelor: Again, changing the return type from a public static method. This change isn't necessary for JUnit testing, and could break community publisher plugins.

AliaksandrDrebenchuk: List list = PentahoSystem.getPublisherList(); // backward compatibility

core/src/org/pentaho/platform/engine/core/system/PentahoSystem.java
AliaksandrDrebenchuk: this is apparent change

core/src/org/pentaho/platform/engine/core/system/PentahoSystem.java
AliaksandrDrebenchuk: I agree with you and I will revert that, because my code can write null into the map instead nullpointer exception.

AliaksandrDrebenchuk: Reverted

AliaksandrDrebenchuk:
@rmansoor please review

bkemper-pentaho:
There are all kinds of changes in PentahoSystem that I don't understand - especially with something that simply says Junit test.    There's a removal of  and isAuthenticated method in startup, there's an addtion of a logging timer, there are some comments by Marc specifically related to why some of these changes are being made without a Jira ticket - since this pull request specifically says it's solely to fix unit tests.   I'm closing it unless there's an extremely valid reason why we're making sweeping changes to startup sequences for a Unit Test case.      BTW -  unit test should always be done BEFORE making refactoring changes.   Without the unit tests in place FIRST we don't know if these changes introduced a regression or not.

PR #2703, opened 2015-11-24
[BACKLOG-5541] -- added rest call to retrieve list of registered plugins

krivera-pentaho:


krivera-pentaho:
@diogofscmariano please review

diogofscmariano:
@krivera-pentaho i think this might be related with the samples framework and not the BACKLOG-5541. Can you please update the commit message? Also, don't we already have a call with this? To retrieve the plugin list?

PR #2705, opened 2015-11-25
[BACKLOG-5658] - Move the pentaho system initialization to a differen…

aliakseihaidukou:
…t thread

- moved PentahoSystem::init code to new thread if needed (false by default for tests purpose)
- added remove listener feature to IServerStatusProvider
- added plugins systemReadyListeners invocation to the end of PentahoSystem::init, moved related test code
- removed caching of PentahoSystem.getInitializedOK() from SystemStatusFilter
- checkstyle fixes

@rmansoor @tkafalas please review. Don't forget about the Wingman report please.


core/src/org/pentaho/platform/engine/core/system/PentahoSystem.java
```diff
@@ -225,133 +230,44 @@ public static boolean init( final IApplicationContext pApplicationContext ) {
...
-    // store a list of the system listeners
-    try {
-      if ( debug ) {
-        Logger.debug( PentahoSystem.class, "Start System Listeners" ); //$NON-NLS-1$
+    } else {
+      PentahoSystem.applicationContext = pApplicationContext;
+      PentahoSystem.initializedStatus = PentahoSystem.SYSTEM_INITIALIZED_OK;
+
+      if ( useThread ) {
+        final Thread startingThread = new Thread( new Runnable() {
```
pentaho-nbaker: Name the thread and set it to Daemon

lucboudreau: +1. This will turn the platform into a zombie.

aliakseihaidukou: @pentaho-nbaker @lucboudreau can you tell why should we make the thread a daemon? Can't find any reasonable reason.

lucboudreau: If the thread is not a daemon, it will not be stopped when we try to shutdown the platform.

PR #2711, opened 2015-11-27
[BACKLOG-5688] - Implement JAXRS and JAXWS endpoints for ServerStatus…

aliakseihaidukou:
…Provider

- as we cannot change config files, I added serverStatus check to GetResource servlet. We can check the status by checking `/pentaho/GetResource?serverStatus` url.

- set MicroPlatform not use thread for PentahoSystem::init

@rmansoor please review


aliakseihaidukou:
@pedrofvteixeira FYI

pedrofvteixeira:
CC'ing @tkafalas

PR #2714, opened 2015-11-30
[BACKLOG-5712] -- added rest call to retrieve list of registered plugins

krivera-pentaho:


krivera-pentaho:
@diogofscmariano @pamval please review

PR #2717, opened 2015-12-01
[BISERVER-12317] - Impossible to publish content without read and write permissions

akhayrutdinov:
- introduce IAggregatingAuthorizationAction interface to hold actions which implicitly require other permissions
  - make Publish action aggregating
- change RoleAuthorizationPolicy to support aggregating actions
- introduce a brand new bean 'nonTransactedAuthorizationPolicy' that uses non-transacted Dao to avoid infinitive loop via indirect recursion
- remove code duplication in PentahoEntryCollector and force it to use non-transacted Policy instead of Dao
- add tests
- fix Checkstyle violations for all affected classes and make some cleanup

@pentaho-nbaker, @rmansoor, review it please

### DIFF
--- a/api/src/org/pentaho/platform/api/engine/IAggregatingAuthorizationAction.java
+++ b/api/src/org/pentaho/platform/api/engine/IAggregatingAuthorizationAction.java
@@ -1,0 +1,36 @@
+/*!
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License, version 2.1 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
+ * or from the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * Copyright (c) 2002-2015 Pentaho Corporation..  All rights reserved.
+ */
+
+package org.pentaho.platform.api.engine;
+
+import java.util.List;
+
+/**
+ * This interface defines aggregating actions which should be implicitly endowed with additional privileges.
+ * For example, {@code Publisher} should be possible to create content as publishing includes it.
+ *
+ * @author Andrey Khayrutdinov
+ */
+public interface IAggregatingAuthorizationAction extends IAuthorizationAction {
+
+  /**
+   * Returns names of aggregated actions
+   *
+   * @return names of aggregated actions
+   */
+  List<String> getAggregatedActions();
+}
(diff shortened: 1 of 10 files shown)


bkemper-pentaho:
Before merging this, I want to validate how this is being addressed.      I'm not sure it makes any sense to say I can publish something if I'm not allowed to write it.   And similarly, does it make any sense to be able to write content that you can't read?    Pretty much as soon as they save it they'd lose their ability to see it?

Let me get some advice on how we expect these various permissions to interact before merging this - since this may be an invalid security setup more than it is a bad enforcement policy...




Bob Kemper
Executive Vice President of Engineering
Pentaho, a Hitachi Data Systems company.
5950 Hazeltine National Dr.
Suite 340
Orlando, FL 32822
(407)-965-4019

[cid:EE55C514-54A2-4BAF-969F-85B0C5235890]

pentaho-nbaker:
So these Write and Read ABS permissions are checked before any ACLs on the Files/Folders. Basically they’re checks to see “can this user read/write to the repository at all?”. If this returns true then the ACLs are consulted to see if the publish target is accessible to the publisher (write on existing file or folder in the case of new). This corrects a situation where someone has been granted “publish”, but they forgot to give them “write” as well.

It does make sense that “Write” is an aggregate granting “Read”, but in reality any Authenticated user is automatically granted “Read”.

-Nick

akhayrutdinov:
@pentaho-nbaker, @bkemper-pentaho, what is the resolution on this PR?

bkemper-pentaho:
No decision yet -   I need to talk to a couple of folks.

The real question on this one is where the actual error occurs - is it on the granting OF the permissions  - should we allow them to set publish if the user doesn't also have read/write,  or is it on the enforcement (which is where this fix is).   My other concern is that this is a fairly significant change - requiring modifications to config files - which means that we couldn't do anything on a patch.     If we make this at the permission grant level - than we could also do this in a patch if required since we could put logic when retrieving the role (as well as preventing them from creating the situation in the first place).

I know Nick likes the permissions aggregator - I just want to see what the proposed solution is.   And the person I want to discuss this with isn't available this week.

Bob Kemper
Executive Vice President of Engineering
Pentaho, a Hitachi Data Systems company.
5950 Hazeltine National Dr.
Suite 340
Orlando, FL 32822
(407)-965-4019

[cid:E0DC70EE-6443-4C79-9F85-A7FB632F6E0C]

pentaho-nbaker:
To be fair he does handle someone upgrading with the old config files. 

The majority of these changes correct bad code in the PentahoEntryCollector which currently doesn't use the IAuthorizationPolicy but maintains pretty much it's own implementation calling directly into the JCR binding code. The effect of this bad code is to make it impossible to supply a new implementation of the IAuthorizationPolicy to the system. So I like that.

Now for the heart of these changes, I gave it more thought and while I do like being able to declare that an Action grants other Actions, I have the following reservations:

1. I don't think it's necessary
If someone is granted a role bound to "publish" but not "write", the system admin can either
  * Bind "write" to the role with "publish"
  * Bind "write" to another role ('content creator' or something) and assign that to the user with "publish"

2. It takes away flexibility from the admin
The definition of aggregate Actions would need to be configurable. leading to...

3. We're further complicating an already complicated system
We already have a flexible way of aggregating actions together, you assign them to a role and then that role to a user. Now it's true that for corporate users creating new roles/groups isn't easy. 
However, I would rather address that pain in another way: let us define supplemental role mappings for users/groups.

4. Because this isn't a major release the code is much more complex than it would be on 7.0
The code would be a lot more clean targeting 7.0 where we can add to existing interfaces.

Lets push this code into a branch on the Pentaho repo and have more discussions with the product owners.

pentaho-nbaker:
Also we need to create a case on the PentahoEntryCollector code duplication and have that addressed separately

PR #2725, opened 2015-12-07
[BACKLOG-5526] Make the ServerStatusProvider pluggable.

tkafalas:


api/ivy.xml
```diff
@@ -22,6 +22,10 @@
 
 		<dependency org="org.springframework.security"   name="spring-security-core"      rev="${dependency.spring.security.revision}" transitive="false"/>
 		<dependency org="org.springframework"            name="spring-beans"              rev="${dependency.spring.framework.revision}"         transitive="false"/>
+		
+		<dependency org="org.slf4j" name="slf4j-api" rev="1.7.3" transitive="false"/>
+    	<dependency org="org.slf4j" name="slf4j-log4j12" rev="1.7.3" transitive="false"/>
```
pentaho-nbaker: We should only need the slf4j-api here, not the log4j stuff

api/src/org/pentaho/platform/api/engine/IServerStatusProvider.java
```diff
@@ -37,4 +43,33 @@
...
+      if ( System.getProperty( PROVIDER_CLASS ) != null ) {
+
+        try {
+          instance = (IServerStatusProvider) Class.forName( System.getProperty( PROVIDER_CLASS ) ).newInstance();
+        } catch ( ClassNotFoundException e ) {
+          logger.error( "ServerStatusProvider class not found", e );
+        } catch ( InstantiationException | IllegalAccessException e ) {
+          logger.error( "ServerStatusProvider class could not be instantiated", e );
+        }
+        return instance;
```
pentaho-nbaker: If any error occurs using their custom class we should instantiate and return the default

api/ivy.xml
tkafalas: Without the log4j stuff the ServerStatusProvider was erroring out on the logger.error

pentaho-nbaker: Ah you mean in a test? You shouldn't need a binding SLF4J just goes silent if no binding is present:
http://www.slf4j.org/manual.html

api/src/org/pentaho/platform/api/engine/IServerStatusProvider.java
tkafalas: Done.

pentaho-nbaker:
Cool thanks Tim

PR #2738, opened 2015-12-18
[BISERVER-12992] - Plugin-samples visible in CE by default

pamval:
@pedrofvteixeira please review.


pedrofvteixeira:
Tested locally, all is good

PR #2739, opened 2015-12-21
[BACKLOG-5858] Introduce ServiceBarrier backed BarrierBeans

tkafalas:
Waits for the jcrRepository Bean to be created before starting the the Karaf 2 minute timeout in KarafFeatureWatcherImpl


pentaho-nbaker:
Closing for now, we're trying to get it working without moving the classes out of the service-coordinator bundle.

PR #2740, opened 2015-12-22
[BISERVER-12857] - manage user role issues tomcat vs jboss

aliakseihaidukou:
show error message for removing role exceptions

@rmansoor please review


dkincade:
@pentaho-nbaker please review

PR #2757, opened 2016-01-14
PPP-3464 - Security Vulnerability- Stored Cross-Site (XSS) Scripting

Aliaksandr-Kastenka:
@rfellows @rmansoor 
Please review.
I did not use ESAPI for encoding on the server side because client's dojox/html/entities cannot decode such values as "&#х27;", "&#х2F;", and so on, and as far as i understand esapi4js is not included in this project, so to be sure that every value will be decoded on the client side, I replaced 4 symbols manually.


Aliaksandr-Kastenka:
@rfellows @rmansoor 
Please take a look.

PR #2763, opened 2016-01-22
[BISERVER-12829] Errors when integrating Pentaho User Console into an…

SergeyTravin:
… IFrame


SergeyTravin:
@pamval  Could you review it?

pamval:
@rfellows, can you check this one, please ?

rfellows:
looks good. basically a search & replace of window.top with window.parent. This should be fine. I can't think of any place that we have more than one nested iframe.

PR #2765, opened 2016-01-25
[BISERVER-12989] KarafBoot is hard-coded with paths for karaf and use…

SergeyTravin:
…s System.setProperty to override all properties

### DIFF
--- a/extensions/src/org/pentaho/platform/osgi/KarafInstance.java
+++ b/extensions/src/org/pentaho/platform/osgi/KarafInstance.java
@@ -12,9 +12,8 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright 2015 Pentaho Corporation. All rights reserved.
+ * Copyright 2016 Pentaho Corporation. All rights reserved.
  */
-
 package org.pentaho.platform.osgi;
 
 import java.io.File;
@@ -70,10 +69,8 @@ public class KarafInstance {
   KarafInstance( String root ) {
     KarafInstance.instance = this;
     this.root = root;
-    assignInstanceNumber();
-    cachePath = root + "/data" + instanceNumber;
-
-    System.setProperty( "karaf.data", cachePath );
+    String presetedCache = System.getProperty( "karaf.data" );
+    updateKarafDataPath(  presetedCache == null ? root + "/" + CACHE_DIR_PREFIX : presetedCache );
   }
 
   public static KarafInstance getInstance() {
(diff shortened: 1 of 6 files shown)


pamval:
@rmansoor @pentaho-nbaker @pedrofvteixeira can you review this one, please ?

lucboudreau:
http://10.100.11.222:8080/job/wingman-ant/264/console

extensions/src/org/pentaho/platform/osgi/KarafInstance.java
```diff
@@ -139,7 +136,7 @@ private void assignInstanceNumber() {
     }
 
     instanceNumber = testInstance;
-
+    System.setProperty( "karaf.data", cachePath );
```
pamval: Why are we inconditionally setting the system property. If it was set from outside, we shouldn't override it, right ?

extensions/test-src/org/pentaho/platform/osgi/KarafBootTest.java
```diff
@@ -124,7 +123,13 @@ public void testBuildExtraKettleEtc() throws Exception {
 
   @Test
   public void testConfigureSystemProperties() throws Exception {
-
+    System.setProperty( "karaf.home", "karaf.home" );
+    String solutionRootPath =  "solutionRootPath";
```
pamval: Shouldn't the test look at the other system.properties ?

karaf.data, for instance, would be useful since we're manipulating it when the instance starts.

extensions/src/org/pentaho/platform/osgi/KarafInstance.java
SergeyTravin: @pamval 
Without this code if we try to start the 2 instance of kettle from one directory than one of them will fail because both try to use one cache folder.
This code add number for instance and will run both instance of kettle with different location of cache.

Which behavior should be in this case?

extensions/test-src/org/pentaho/platform/osgi/KarafBootTest.java
SergeyTravin: @pamval, The similar mechanism is used for all system.properties and I supposed that it is enough. Should I checked all properties?

pamval:
@rmansoor @pentaho-nbaker @pedrofvteixeira can you review this one, please ?

extensions/test-src/org/pentaho/platform/osgi/KarafBootTest.java
pamval: I think so. If we're treating karaf.data differently, we should validate its value is the expected

extensions/src/org/pentaho/platform/osgi/KarafInstance.java
pamval: If the karaf.data property is already set, we should not override its value.
Unless we really need to.
Deferring to @pentaho-nbaker or @rmansoor on whether this is ok or if we should call fillMissedSystemProperty instead

pamval:
@rmansoor @pentaho-nbaker @pedrofvteixeira can you review this one, please ?

PR #2769, opened 2016-01-26
[BISERVER-12794] Permissions can delete but not restore file

IvanNikolaychuk:
**FileService**

- Move the logic of file-copy method to CopyFilesOperation class, represent copy file operation as an immutable object, and move specific method (used only from copyFiles method) to this CopyFilesOperation class.
- Small optimisation in doRestoreFilesToHomeFolder method, when no overwrite mode (passing only non-conflict files for copy operation)

**CopyFilesOperation**

- Contains all the logic of FileService  doCopyFiles method (copy with every possible mode is devided into a separete method). 
- Added a possibility to copy folders, performing deep copy folder.

*Note:* I found no reason to limit deepness of folders hierarchy (the worst possible consequence - long scanning)

**RepositoryFileHelper**
- Class, where some methods from FileService were replaced (so as to be able to use them from CopyFilesOperation class.

**FileResource, RestoreFileCommand**

- Changed returned responce status, that is sent when restoring files in user home directory and there are no conlicts in it. Need to do it, because of IE bug, that doesn't see some statuses ([link to this bug](https://connect.microsoft.com/IE/feedback/details/785990/ie-10-on-win8-does-not-assign-the-correct-status-to-xmlhttprequest-when-the-result-is-401))

**FileServiceTest**
Due to refactoring few tests related to copy operation were changed (behavior wasn't changed and new tests, that test everything, that was tested in previouse tests (and much more) were written ). 

**FileResourceIT** 
One test for copy operation was deleted, as it was dummy. In short: it tested nothing, as destination directory was null  (In my approach, I validate it in constructor), so copy operation didn't even start. If more explicit report needed, I will provide it.

**CopyFilesOperation_DeepFolderCopyTest, CopyFilesOperation_CopyTest**
- Tests for CopyFilesOperation class

**JcrRepositoryFileDao**
- Provided a way to copy/move folders, not only files.
- Small refactoring made (to increase code readility)

**JcrStringHelper**
- Method, checking if the path is decoded or not.
- Test written
- Small refactoring

***Final Note:*** The changes are big enough, and after making them I tested them as carefully as possible. So, I made tests on:
- Simple copy of files (so as to be sure, I didn't break something): here I found a bug (not because of my changed), and it is reproducible since 6.0 version (I made a [JIRA ticket](http://jira.pentaho.com/browse/BISERVER-13022) for it)
- Restoring files.
- Restoring folders with 10 folder hierarchy.

@rmansoor, @pentaho-nbaker, review it please

### DIFF
--- a/extensions/src/org/pentaho/platform/web/http/api/resources/FileResource.java
+++ b/extensions/src/org/pentaho/platform/web/http/api/resources/FileResource.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2015 Pentaho Corporation..  All rights reserved.
+ * Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
  */
 
 package org.pentaho.platform.web.http.api.resources;
--- a/extensions/src/org/pentaho/platform/web/http/api/resources/FileResource.java
+++ b/extensions/src/org/pentaho/platform/web/http/api/resources/FileResource.java
@@ -348,7 +348,7 @@ public class FileResource extends AbstractJaxRSResource {
         if ( !fileService.canRestoreToFolderWithNoConflicts( getUserHomeFolder(), params ) ) {
           return buildStatusResponse( Response.Status.CONFLICT );
         } else {
-          return buildStatusResponse( Response.Status.TEMPORARY_REDIRECT );
+          return buildStatusResponse( Response.Status.NOT_ACCEPTABLE );
         }
       } catch ( InternalError e ) {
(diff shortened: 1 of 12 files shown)


IvanNikolaychuk:
@rmansoor, review it please

IvanNikolaychuk:
@rmansoor, @pentaho-nbaker, review it please

pamval:
@pentaho-nbaker @rmansoor @pedrofvteixeira  guys, this has been pending for a week now. Can one of you take a look at merge this in, please ?

rmansoor:
@pamval @IvanNikolaychuk I validated this locally and ran some scenario using the nightly build and all of them succeeded. I will process the pull request now

dkincade:
NOTE: Sonar is complaining about the static imports that were included in this commit. Unless there is a very good reason, they should be removed / reworked.
http://sonar.pentaho.com:9000/issues/search#projectUuids=AVD9SSK1k9roflTwW6mt|createdAt=2016-02-03T07%3A00%3A02-0500

PR #2771, opened 2016-01-28
[SP-2446][BISERVER-12829] Errors when integrating Pentaho User Consol…

SergeyTravin:
…e into an IFrame


SergeyTravin:
@rfellows Could you review it?
It is backport for #2763
Thanks!

PR #2772, opened 2016-01-29
[BACKLOG-6391] - Administration page does not show any content

aliakseihaidukou:
- set default value of dampeningTimeout to *0* if undefined in system properties.
- Unit test
- checkstyle

@rmansoor @CodeOnCoffee please review


rmansoor:
@aliakseihaidukou validated using the latest nightly and the local code. It works fine now with this change. unit test are also working properly. Processing the pull request now

PR #2780, opened 2016-02-05
[PPP-3455] - Dynamic code injection vulnerabilities found

akhayrutdinov:
- in EventBusUtilGenerator, change JSON object generating approach: use `"` instead of `'` to follow ECMA standard

@mdamour1976, review it please. This previous commit broke PUC, as `JSON.parse` rejects `'` as quoting signs.

We need a lot of backslashes here, because  " inside "" should be escaped as `\"`, and that `\` should be escaped as well -- `\\`, and two quotes need their own escaping...

I build mantle locally, and Browse File works now.

/cc @pamval


pamval:
I think we're missing a \ there. :-P

PR #2783, opened 2016-02-06
[BACKLOG-6516] Relocate and rename pentaho.xml's 'verify-user-on-prin…

pedrofvteixeira:
…cipal-creation' property

	- relocated property from pentaho.xml to security.properties
	- renamed it to it's opposite: from 'verify-user-on-principal-creation' to 'SkipUserVerificationOnPrincipalCreation'
	- ensured property is set to a default value of 'true'
	- safeguard/fallback behaviour: if property does not exist or is set to empty, assumes default value of 'true'
	- added a INFO-level log message with the property's value
	- sanitization: this property's value is checked only once on initialization ( rather that constantly reading its value from pentaho-xml/security.properties )
	- updated unit tests

### DIFF
--- a/assembly/package-res/biserver/pentaho-solutions/system/pentaho.xml
+++ b/assembly/package-res/biserver/pentaho-solutions/system/pentaho.xml
@@ -28,15 +28,6 @@
 	<log-level>DEBUG</log-level>
 
 	<!--
-		This flag indicates whether or not UserDetailsService is called during creation of user's principal.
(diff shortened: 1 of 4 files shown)


pedrofvteixeira:
@pentaho-nbaker , @rmansoor @dkincade please review.
Should be merged alongside https://github.com/pentaho/pentaho-platform-ee/pull/895

PR #2792, opened 2016-02-10
[BISERVER-12865] 5.0 to 6.0 Regressions in Admin > User/Roles screen

plagoa:


plagoa:
Validated Anna's changes, but had to add some IE10 specific styles. @pentaho/orlandocalisbon please review.

user-console/source/org/pentaho/mantle/public/themes/crystal/mantleCrystal.css
```diff
@@ -1283,10 +1283,19 @@ a:active, a:hover {
...
 
 .users-roles-list {
-  height: 432px;
+  height: 100%;
   width: 200px;
 }
 
+/* targetting IE 10 only */
+.IE10 .pentaho-tab-panel {
+    height:541px!important;
```
krivera-pentaho: don't you need a space between the px and !important?

pamval: Please review this. Using !important is as bad as using goto in a programming language

plagoa: Even though it's usually used with a space, it works either way. In this case, the change is reflected on the styles being applied so it's working properly.

krivera-pentaho: @pamval I hate using !important as well, but for IE styles in our product, I have had to use !important b/c I was not able to make a more specific style to "win" for IE. But i agree, !important is not good.

plagoa: @pamval I realize it's a bad option, but in this case it's a very specific scenario for a specific style to be overriden only in IE10. If even so it's not acceptable, I'll try to find another solution.

plagoa:
@pamval I removed the use of !important. Removed adding of styles on the server side and used an existing class to deal with CSS specificity. @pentaho/orlandocalisbon please review.

pamval:
Can't the setHeights done in javascript migrate to css rules ?

plagoa:
@pamval made that change. Only had to keep one of those setHeights as it would be really messy to do that in css. All the other cases were migrated to css. @pentaho/orlandocalisbon please review.,

pamval:
fine by me. Over to the very technical people

PR #2793, opened 2016-02-10
[PPP-3455] - Dynamic code injection vulnerabilities found

akhayrutdinov:
- rework code generator to produce json strings fulfilling ECMA standard
  - add tests
- replace `eval()` with `JSON.parse()` and add defensive code before invocation

@mdamour1976, @pentaho-nbaker, review it please.

So, here we have two problems. The first is java-generated json representation with `'` as enclosures. I re-implemented that part and added several tests.

The second is more tricky. It is about getting `undefined` on calling getters. It can be fixed by making the template for generating more complex (adding `!= null` conditions there). However, I choosed another way: to change the parameter before it is parsed via regular expressions.

As for changes in `ivy.xml`, all added artifacts are already used. Moreover, they are added to `test` configuration. Hence, no component processing is required.

### DIFF
--- a/user-console/ivy.xml
+++ b/user-console/ivy.xml
@@ -79,8 +79,12 @@
           <artifact name="validation-api" type="source" ext="jar" m:classifier="sources"/>
         </dependency>
 
-        <!--  Testing dependencies -->
-        <dependency org="junit"          name="junit"          rev="4.12" conf="test->default" />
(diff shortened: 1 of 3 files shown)


akhayrutdinov:
@mdamour1976, @pentaho-nbaker, review it please.

PR #2811, opened 2016-02-22
[BISERVER-13062] 5.0 to 6.0 spacing between arrows in manage users sc…

diogofscmariano:
…reen has increased greatly


diogofscmariano:
@annapopova1 can you please double check this again?

This is not totally safe, since this is removing css for a panel from gwt modules. ShareFileCommand, MantleTab are some examples. It might be safer if you create a new class.

PR #2819, opened 2016-02-24
[BACKLOG-6660] - Adding the missing version configuration beans

rmansoor:

PR #2821, opened 2016-02-24
[BACKLOG-6660] - Adding the missing version configuration beans

rmansoor:


rmansoor:
@dkincade this is a cherry-pick from master https://github.com/pentaho/pentaho-platform/pull/2819
This is required for master version flags to work in Pentaho Server

pentaho-nbaker:
Ramaiz, who is querying PentahoSystem to find these settings? They should just get the repository configuration from ISystemConfig

rmansoor:
It is right here
 https://github.com/pentaho/pentaho-platform/blob/afaecf7ca9493977d32f018f37a23375587aa135/repository/src/org/pentaho/platform/repository2/unified/DefaultRepositoryVersionManager.java#L47-L57

dkincade:
Deferring to @pentaho-nbaker on this item. If this is good with him then it can be included. Otherwise we will need to implement an alternative and update the master branch as well.

rmansoor:
@dkincade this is how EE has it, if we change it for CE we need to change it for EE as well. I can see why @pentaho-nbaker is saying to get the config from ISystemConfig but how many  PentahoSystem calls have we switched to ISystemConfig ?

dkincade:
Adding to the tech debt pile. We need to get this into 6.1 quick so I am taking it. But we do need to pay down some of this debt.

PR #2829, opened 2016-03-01
6.1 feature

tkafalas:


tkafalas:
@pentaho-nbaker The first commit backs out the bad change.  The second commit makes the new change and does some formatting to get by checkstyle

extensions/src/org/pentaho/platform/osgi/KarafBoot.java
```diff
@@ -12,10 +12,19 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright 2016 Pentaho Corporation. All rights reserved.
+ * Copyright 2014 Pentaho Corporation. All rights reserved.
```
pentaho-nbaker: Copyright should stay 2016

extensions/src/org/pentaho/platform/osgi/KarafBoot.java
```diff
@@ -46,12 +47,12 @@
   public static final String CLEAN_KARAF_CACHE = "org.pentaho.clean.karaf.cache";
   private Main main;
   Logger logger = LoggerFactory.getLogger( getClass() );
-
+  private static boolean initialized;
```
pentaho-nbaker: what is this? Looks like they removed an un-used variable

pentaho-nbaker:
Tim I'd rather you just build on what they put in rather than backing it out. Just make it so we control the data directory. I do like the preservation of system properties for the other karaf stuff that they put in

PR #2830, opened 2016-03-01
Feature

tkafalas:


pentaho-nbaker:
Why is this in code instead of a new configuration file?

tkafalas:
This is what the case calls for.  I think the idea here was to not let the customer have control of it.

pentaho-nbaker:
We're going to work on this some more. Closing for now

PR #2835, opened 2016-03-06
Fix in mondrianCatalogHelper cache handling

igieon:


mbatchelor:
@igieon - can you please create a case in JIRA to cover the issue this is fixing, reproduction steps, etc? I cannot merge the PR without a JIRA case.

http://jira.pentaho.com - thanks!

igieon:
Here is already created http://jira.pentaho.com/browse/BISERVER-13140 and simple hwo to simulate just change default cache configuration and login with users with diferent locale.

mbatchelor:
Fixed by BISERVER-13240. Closing the PR

PR #2841, opened 2016-03-11
[SP-2461] - Backport of PPP-3450 - Backbone.js vulnerable to DOM XSS (5.4 Suite)

akhayrutdinov:
- in current backbone (1.0.0), remove Routing part, simply as it was done in https://github.com/webdetails/cdf/commit/d717a63918a1b151db443078c3f3155bb3bf98d1, to prevent from XSS vulnerability


user-console/source/org/pentaho/mantle/public/browser/lib/backbone/backbone.js
```diff
@@ -1587,7 +1276,7 @@
   };
 
   // Set up inheritance for the model, collection, router, view and history.
-  Model.extend = Collection.extend = Router.extend = View.extend = History.extend = extend;
+  Model.extend = Collection.extend = View.extend = History.extend = extend;
```
afrjorge: @akhayrutdinov @pamval Caution, **History** might be undefined here.

PR #2866, opened 2016-03-30
[BACKLOG-7180] The wait dialog “Just a few moments please” appears be…

dzmitry-hryhoryeu:
The wait dialog “Just a few moments please” appears before the progress bar

@tmorgner please review.

### DIFF
--- a/user-console/source/org/pentaho/mantle/client/ui/tabs/MantleTabPanel.java
+++ b/user-console/source/org/pentaho/mantle/client/ui/tabs/MantleTabPanel.java
@@ -96,7 +96,7 @@ public class MantleTabPanel extends org.pentaho.gwt.widgets.client.tabs.PentahoT
       }
     } );
 
-    setIsAsyncModeEnabled();
+    fetchAsyncModeEnabledStatus();
   }
 
   public void addTab( String text, String tooltip, boolean closeable, Widget content ) {
@@ -784,7 +784,13 @@ public class MantleTabPanel extends org.pentaho.gwt.widgets.client.tabs.PentahoT
     }
   }-*/;
 
-  public void setIsAsyncModeEnabled() {
+  /**
+   * Queries the server once to see whether report-viewer async-mode is enabled. If it is enabled,
+   * the report-viewer provides it own loading screen and the default "please wait" screen interferes
+   * with the user experience of that. Therefore we disable the wait-screen that comes up when a new
+   * page opens. @see case BACKLOG-7180 for details.
+   */
+  public void fetchAsyncModeEnabledStatus() {
     final String moduleBaseURL = GWT.getModuleBaseURL();
     final String moduleName = GWT.getModuleName();
     final String contextURL = moduleBaseURL.substring( 0, moduleBaseURL.lastIndexOf( moduleName ) );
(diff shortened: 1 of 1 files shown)


user-console/source/org/pentaho/mantle/client/ui/tabs/MantleTabPanel.java
```diff
@@ -88,6 +96,7 @@ public void onWindowClosing( ClosingEvent event ) {
       }
     } );
 
+    setIsAsyncModeEnabled();
```
tmorgner: Your method is not a setter, so it should be called differently. Also: You are calling the server on every tab that opens, to get a value that is guaranteed to never change (unless you restart the server).

Rename the method to "fetchAsyncModeEnabledStatus()".  This way you can read the constructor and sensibly guess what happens with that method without having to read the method code itself. 

A setXXX method without parameters simply does not match any java-coding style.

user-console/source/org/pentaho/mantle/client/ui/tabs/MantleTabPanel.java
```diff
@@ -773,4 +783,35 @@ public static native void onTabSelect( Element element )/*-{
         // ignore
     }
   }-*/;
+
+  public void setIsAsyncModeEnabled() {
```
tmorgner: Please add a explaining comment here so that other developers know what this is and why it is there. The "mantle" code is shared and developers who have not worked in the report-viewer may not know anything about the work we do.

A short explanation like: "Queries the server once to see whether report-viewer async-mode is enabled. If it is enabled, the report-viewer provides it own loading screen and the default "please wait" screen interferes with the user experience of that. Therefore we disable the wait-screen that comes up when a new page opens. @see case BACKLOG-xxxx for details."

This way anyone reading the code in a year or two will know why this is here.

tmorgner:
Gosh, I strongly hate the ugliness of having to put reporting specific code into a shared code base. But it seems we cannot avoid that to get rid of the loading screen.

EgorZhuk:
and we will need a build...

tmorgner:
Lets see how complex that is. If it is a hell of work, then lets skip the build. After all, this change only affects the extra "wait" screen.  We are going to do the master-merge soon and then we can see the change without having to fiddle with the build all the time.

user-console/source/org/pentaho/mantle/client/ui/tabs/MantleTabPanel.java
dzmitry-hryhoryeu: Thank you Thomas for review. Your comments are very valuable. The server calling once when MantleTabPanel created on home page. Next config request to server is sent from pentaho-reporting-plugin.

tmorgner: I see. But please rename the function so that it does not try to look like a setter.

dmitriygg: I have renamed.

dzmitry-hryhoryeu: Already renamed.

PR #2868, opened 2016-03-31
[BACKLOG-7324, BACKLOG-7407] Fixing case where two instances resolve …

pentaho-nbaker:
…the same port before opening it to mark it used.


eliofreitas:
Rejected from automatic build ,authorized users only

PR #2910, opened 2016-04-27
[BISERVER-13184] IE8 - Empty UI after login to PUC

mchen-len-son:
added doctype adding to html generated by GWT to force IE9 and earlier using Standards Mode

### DIFF
--- a/user-console/build-res/subfloor-gwt.xml
+++ b/user-console/build-res/subfloor-gwt.xml
@@ -188,6 +188,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
       <fileset dir="${gwt.output.dir}" includes="**/*.cache.html" />
     </replaceregexp>
 
+    <!-- Adding doctype to the start of file to force using Standards Mode in IE9 and earlier -->
+    <replaceregexp match="^^" replace="&lt;!DOCTYPE html&gt;${line.separator}">
+      <fileset dir="${gwt.output.dir}" includes="**/*.cache.html" />
+    </replaceregexp>
   </target>
 
 


brosander:
Build Commands:
==============
Build:
------
1. ant -Dtestreports.dir=bin/reports/unit-test -Djunit.jvmargs=-XX:PermSize=256m -f user-console/build.xml clean-all resolve jacoco && ant -Dpentaho.coding.standards.version=1.0.4 -Dtestreports.dir=bin/reports/integration-test -Djunit.jvmargs=-XX:PermSize=256m -f user-console/build.xml jacoco-integration checkstyle publish-local

Cleanup:
-------
rm -r ~/.ivy2/local || echo no publish local to remove

Changed files:
==============
  user-console/build-res/subfloor-gwt.xml

pamval:
@rfellows can you review this change ? 

I'm uncertain on whether we should add this to master, since it's IE8 specific. We definitely need it in 5.4 but I'm not sure about master.

rfellows:
@pamval we shouldn't need this in master since we aren't going to be supporting IE8 or 9 in the next release.

I would be cautious with this change in general, however. Changing GWT-generated code to force standards-mode might cause other issues. There are cases elsewhere in the platform that where we had to force Quirks-mode for old browser support and standards mode would break it. If GWT is generating it this way, there may be a good/required reason for it. I'd make sure to fully regression test PUC after this change just to be sure.

Why did this problem just arise now? Did we upgrade GWT or something?

pamval:
Good question. We know this broke in a 5.4 service pack so it's very unlikely that was due to a GWT upgrade. @mchen-len-son please investigate and try to figure out what changed in that service pack.

thanks!

mchen-len-son:
Agreed.
Trying to figure out what really happened there.

PR #2911, opened 2016-04-28
[TEST] test for PRPTImportHandler

SergeyTravin:


SergeyTravin:
@pedrofvteixeira Could you review it please? I am extract some code in separate method for test goal and  checked the logic inside PRPTImportHandler.

pedrofvteixeira:
Tested locally:
- unit tests OK
- refactoring work done on PRPTImportHandler does not compromise import handler logic ( tested that as well )

PR #2913, opened 2016-04-28
BACKLOG-7870 [DEV] Update import export command line utility to allow non-admins to import content to their home folder

AliaksandrDrebenchuk:
all users can import/export
HTTP 403 error when import/export failed


rmansoor:
@AliaksandrDrebenchuk we can't make changes to interface for any arbitrary reasons. As I told you we do not need file exist method. I do not see any value of having a file exist method right now. I think this is a little to much change to a simple request.

AliaksandrDrebenchuk:
@rmansoor or new DAO method or 403 error instead 404 always

rmansoor:
@AliaksandrDrebenchuk closing this PR.

PR #2914, opened 2016-04-29
[SP-2662] Backport of BISERVER-13184 - IE8 - Empty UI after login to …

mchen-len-son:
…PUC (5.4 Suite)

- added json2.js lib to the app
- added custom GWT linker adding json2.js for IE8,IE9 permutations
- added unit test for the new linker

### DIFF
--- a/user-console/ivy.xml
+++ b/user-console/ivy.xml
@@ -81,6 +81,7 @@
 
         <!--  Testing dependencies -->
         <dependency org="junit"          name="junit"          rev="4.4" conf="test->default" />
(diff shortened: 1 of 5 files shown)


mchen-len-son:
@pamval @rfellows please review

mchen-len-son:
fixed a couple of checkstyle violations

pamval:
@mchen-len-son the json2.js file doesn't look right. Is it supposed to be an html file ?

pamval:
Also, if this file was not yet part of our deliverables, we need to create a new component ticket for it.

PR #2922, opened 2016-05-03
[TEST] - Junit test for SolutionFileImportHelper

apratkin:


dkincade:
@apratkin please fix the headers in the test file (as wingman describes) and let's get this test included.

PR #2926, opened 2016-05-03
[BACKLOG-7741] java.io.FileNotFoundException, Could not create cache …

PavelSakoun:
…directory & Failed to register plugin sparkl errors in Spoon when starting it from user with read permisions
We do not use persistent caches so they are removed upon shutdown and in worst case disk usage bump we're getting is in case of multiple user scenario (e.g spoon), which is (number of users - 1)*size of cache.

@pentaho-nbaker


PavelSakoun:
@pentaho-nbaker please review

PR #2930, opened 2016-05-05
[BISERVER-12997] - Mondrian 3 connections getting improperly handled …

andrei-viaryshka:
…by PentahoSystemDriver causing a class not found exception


andrei-viaryshka:
@kurtwalker Could you please take a look? It is the fix you suggested.

PR #2931, opened 2016-05-05
[BACKLOG-7872] - Create JMeter script to execute import endpoint usin…

apratkin:
…g admin and non admin user


apratkin:
@rmansoor could you please review my commit?

PR #2936, opened 2016-05-09
[PDI-14385] - Adding extra role to the user  IUserRoleListService.getRoles and IUserRoleListService.getRolesForUser call

rmansoor:


rmansoor:
@mbatchelor please review.

extensions/src/org/pentaho/platform/plugin/services/security/userrole/jdbc/JdbcUserRoleListService.java
```diff
@@ -146,6 +152,14 @@ public String getAllUsernamesQuery() {
       }
 
     }
+
+    // Now add extra role if it does not exist in the list
+    for ( String extraRole : extraRoles ) {
+      if ( !roles.contains( extraRole ) ) {
```
mbatchelor: Strongly suggest a LinkedHashSet instead of this. For LDAP, the list of roles will be potentially very long. This is an o^n search across all the roles for the extra roles.

Something like:
     LinkedHashSet<String> roles = new LinkedHashSet<String>( allAuths.size() );
     for ( GrantedAuthority role : allAuths ) {
       if ( roleMapper != null ) {
         roles.add( roleMapper.toPentahoRole( role.getAuthority() ) );
       } else {
         roles.add( role.getAuthority() );
       }
      }

     // Now add extra role if it does not exist in the list
     for ( String extraRole : extraRoles ) {
       roles.add( extraRoles );
     }
     return new ArrayList<String>(roles);

This is just my $0.02 though...

extensions/src/org/pentaho/platform/plugin/services/security/userrole/ldap/DefaultLdapUserRoleListService.java
```diff
@@ -155,6 +165,13 @@ public void afterPropertiesSet() throws Exception {
     for ( GrantedAuthority role : results ) {
       roles.add( role.getAuthority() );
     }
+    // Now add extra role if it does not exist in the list
+    for ( String extraRole : extraRoles ) {
```
mbatchelor: Whoa ... just because there are extra roles doesn't mean that this user has all of those roles, does it ? Is this correct ?

extensions/src/org/pentaho/platform/plugin/services/security/userrole/ldap/DefaultLdapUserRoleListService.java
```diff
@@ -114,6 +117,13 @@ public void afterPropertiesSet() throws Exception {
         roles.add( roleString );
       }
     }
+    // Now add extra role if it does not exist in the list
+    for ( String extraRole : extraRoles ) {
+      if ( !roles.contains( extraRole ) ) {
```
mbatchelor: Once again, in an LDAP case, this is likely to be poorly performing. With 1000 roles from LDAP, and 5 "extra roles", this could take a considerable time in the "lookup" part. If there are only ever 2 extra roles, then it's likely OK - it depends on what the "extra roles" are.

extensions/src/org/pentaho/platform/plugin/services/security/userrole/jdbc/JdbcUserRoleListService.java
```diff
@@ -241,6 +255,13 @@ protected Object mapRow( final ResultSet rs, final int rownum ) throws SQLExcept
       }
     }
 
+    // Now add extra role if it does not exist in the list
+    for ( String extraRole : extraRoles ) {
```
mbatchelor: Is every user expected to always have all the extra roles ?

extensions/src/org/pentaho/platform/web/http/api/resources/services/UserRoleListService.java
```diff
@@ -71,7 +71,17 @@ public UserListWrapper getUsers() {
...
   public RoleListWrapper getRoles() {
-    return new RoleListWrapper( getUserRoleListService().getAllRoles() );
+    return getRoles( true );
+  }
+
+  public RoleListWrapper getRoles( boolean includeExtraRoles ) {
+    List<String> roles = getUserRoleListService().getAllRoles();
+    if ( !includeExtraRoles ) {
+      for ( String role : getExtraRoles() ) {
+        roles.remove( role );
```
mbatchelor: Moderately concerned here - if the "extra" role is also not an extra role (comes in from LDAP), then this will remove it anyway. This may be why we designed this stuff to have them as separate items.

From LDAP:
abc
def
Authenticated
ghi

Extra Roles
Authenticated
Administration

getRoles( false ):
abc
def
ghi

Is that correct ?

extensions/src/org/pentaho/platform/plugin/services/security/userrole/ldap/DefaultLdapUserRoleListService.java
rmansoor: extra roles are "Authenticated" and "Anonymous" in our platform. The user has an option of adding to that list and will be added to the role retrieved from the third party security provider. In this case we will mostly have "Authenticated" as an extra role.

extensions/src/org/pentaho/platform/plugin/services/security/userrole/jdbc/JdbcUserRoleListService.java
rmansoor: In our case, every authenticated user get the "Authenticated" role. This does not exist in the third-party security provider. So, yes every authenticated user is expected to have atleast one extra role.

extensions/src/org/pentaho/platform/plugin/services/security/userrole/ldap/DefaultLdapUserRoleListService.java
rmansoor: The extra role are added to the user's role that exist on the third-party security provider. So those role are being added to the user. Are you still concerned here ?

extensions/src/org/pentaho/platform/web/http/api/resources/services/UserRoleListService.java
rmansoor: You are correct. This is a scenario where are removing extraRole from the list. If the user has a role name matching the extra role that we understand, it will be removed. I am open to suggestions here.

extensions/src/org/pentaho/platform/plugin/services/security/userrole/jdbc/JdbcUserRoleListService.java
```diff
@@ -146,7 +153,15 @@ public String getAllUsernamesQuery() {
       }
 
     }
-    return roles;
+
+    // Now add extra role if it does not exist in the list
+    for ( String extraRole : extraRoles ) {
+      if ( !roles.contains( extraRole ) ) {
```
mbatchelor: No need for this if statement if using a LinkedHashSet - that was the reason for the change. The LinkedHashSet will only allow unique values (like all other Set types).

extensions/src/org/pentaho/platform/plugin/services/security/userrole/jdbc/JdbcUserRoleListService.java
```diff
@@ -241,6 +256,13 @@ protected Object mapRow( final ResultSet rs, final int rownum ) throws SQLExcept
       }
     }
 
+    // Now add extra role if it does not exist in the list
+    for ( String extraRole : extraRoles ) {
+      if ( !roles.contains( extraRole ) ) {
```
mbatchelor: As before, this will be costly with a lot of roles. No need to do a contains or similar if you just use a LinkedHashSet

extensions/src/org/pentaho/platform/plugin/services/security/userrole/ldap/DefaultLdapUserRoleListService.java
mbatchelor: Cool - with LinkedHashSet no need for this lookup

extensions/src/org/pentaho/platform/plugin/services/security/userrole/ldap/DefaultLdapUserRoleListService.java
```diff
@@ -155,6 +167,13 @@ public void afterPropertiesSet() throws Exception {
     for ( GrantedAuthority role : results ) {
       roles.add( role.getAuthority() );
     }
+    // Now add extra role if it does not exist in the list
+    for ( String extraRole : extraRoles ) {
+      if ( !roles.contains( extraRole ) ) {
```
mbatchelor: Again - costly lookup can be avoided with a LinkedHashSet

extensions/src/org/pentaho/platform/web/http/api/resources/services/UserRoleListService.java
mbatchelor: I suppose it should be left as-is. We just need to document that they cannot have a regular role anywhere that is the same as an "extra" role.

PR #2942, opened 2016-05-23
[BACKLOG-8452] - PIR: sometimes async execution is not cancelled if r…

dzmitry-hryhoryeu:
[BACKLOG-8452] PIR: sometimes async execution is not cancelled if report mode is changed

@tmorgner
please review

### DIFF
--- a/user-console/source/org/pentaho/mantle/client/ui/xul/MantleController.java
+++ b/user-console/source/org/pentaho/mantle/client/ui/xul/MantleController.java
@@ -31,6 +31,7 @@ import com.google.gwt.json.client.JSONArray;
 import com.google.gwt.json.client.JSONParser;
 import com.google.gwt.user.client.Command;
 import com.google.gwt.user.client.Element;
+import com.google.gwt.user.client.Timer;
 import com.google.gwt.user.client.Window;
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.gwt.user.client.ui.MenuBar;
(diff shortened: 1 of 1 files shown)


user-console/source/org/pentaho/mantle/client/ui/xul/MantleController.java
```diff
@@ -875,6 +876,15 @@ public void setContentEditSelected( boolean selected ) {
...
   public void editContentClicked() {
+    // set delay for edit/view mode switching
+    setContentEditEnabled( false );
+    Timer loadOverlayTimer = new Timer() {
+      public void run() {
+        setContentEditEnabled( true );
+      }
+    };
+    loadOverlayTimer.schedule( 1000 );
+
```
tmorgner: Isn't one second a bit harsh? A user will be bound to notice the delay, especially as there will be no visual feedback while the timer runs. 

Would your fix also work if we reduce it to 250ms?

dzmitry-hryhoryeu: The lowest value at which stable 800ms. 
When the timer is running, the user sees a disabled icon (Edit/View mode toggle) - at the time the request is fulfilled as before. This protection against double click, and a very frequent switching modes

tmorgner:
Yeah, Wingman dies. Big deal. Damn I'm so surprised I accidentally hit the merge button :)

PR #2945, opened 2016-06-06
[BACKLOG-7075] - do not mask interrupted exception

prakapenka:
When obtaining session if thread was interrupted - JCR throw
RepositoryException wrapped over InterruptedException.

Since InterruptedException was thrown (for example during login phase
trying to acuire read write lock) - interrupted flag is cleared.

In our code we intercept and supress this RepositoryException during
session extraction - so InterruptedException is suppressed AND interrupted
flag is cleared both.

This cause we unable to interrupt callable using future.cancel( true )
call. And report keep running.

Guava code from the other hand inspect exception - and if it is
InterruptedException thrown from loader code - it is restore interrupted
status.
That what exactly we expect - to be able to handle interruption and stop
parent execution.

So in case RepositoryException was caused by InterruptedException - we
have to throw interrupted exception  and let the guava cache loader
implementation to restore interrupted flag.

### DIFF
--- a/repository/src/org/pentaho/platform/repository2/unified/jcr/sejcr/GuavaCachePoolPentahoJcrSessionFactory.java
+++ b/repository/src/org/pentaho/platform/repository2/unified/jcr/sejcr/GuavaCachePoolPentahoJcrSessionFactory.java
@@ -74,7 +74,15 @@ class GuavaCachePoolPentahoJcrSessionFactory extends NoCachePentahoJcrSessionFac
       }
     } ).recordStats().build( new CacheLoader<CacheKey, Session>() {
       @Override public Session load( CacheKey credKey ) throws Exception {
-        return GuavaCachePoolPentahoJcrSessionFactory.super.getSession( credKey.creds );
+        try {
+          return GuavaCachePoolPentahoJcrSessionFactory.super.getSession( credKey.creds );
+        } catch ( RepositoryException e ) {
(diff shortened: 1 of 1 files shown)


repository/src/org/pentaho/platform/repository2/unified/jcr/sejcr/GuavaCachePoolPentahoJcrSessionFactory.java
```diff
@@ -74,7 +74,15 @@ public GuavaCachePoolPentahoJcrSessionFactory( Repository repository, String wor
       }
     } ).recordStats().build( new CacheLoader<CacheKey, Session>() {
       @Override public Session load( CacheKey credKey ) throws Exception {
-        return GuavaCachePoolPentahoJcrSessionFactory.super.getSession( credKey.creds );
+        try {
+          return GuavaCachePoolPentahoJcrSessionFactory.super.getSession( credKey.creds );
+        } catch ( RepositoryException e ) {
+          if ( e.getCause() instanceof InterruptedException ) {
+            // do not mask interruptions from caller.
+            throw new InterruptedException( "session cache interrupted" );
```
pentaho-nbaker: I would "throw e.getCause()" instead so we get a full stacktrace

tmorgner:
Closed. We removed the need for interrupt-signals from the reporting engine. Let nature take its course ...

PR #2946, opened 2016-06-08
 [BACKLOG-8389] - Upgrade antlr to version 3.4 across Pentaho product suite - changes for pentaho platform repository

alexzhk:
Replace antrl-2.7.6, antlr-runtime-3.1.1 with antlr-3.4-complete which contains antlr-2.7.7 and antlr-runtime-3.4.
 antlr-2.7.6 is used by hibernate-core version 3.6.9.Final. Differences between antrl-2.7.6 and antrl-2.7.7 do not impact on hibernate-core;
antlr-runtime-3.1.1 is used by drools-compiler-5.0.1. Differences between antrl-runtime-3.1.1 and antrl-runtime-3.4 don`t impact on drools-compiler-5.0.1.

To build pentaho-platform it previously needs to build:
- pentaho-reporting (pentaho-reporting-engine-classic-core) as it was changed to use antlr-3.4-complete (https://github.com/pentaho/pentaho-reporting/pull/790);
- pentaho-kettle (kettle-engine) as it has dependency on pentaho-reporting.

### DIFF
--- a/core/ivy.xml
+++ b/core/ivy.xml
@@ -57,10 +57,11 @@
       		<exclude org="dom4j" name="dom4j" />
       		<exclude org="net.sf.ehcache" name="ehcache" />
       		<!-- CM-241 -->
-      		<exclude org="cglib" name="cglib" />      
+      		<exclude org="cglib" name="cglib" />
+			    <exclude org="antlr" name="antlr" />
     	</dependency>
     		<!-- CM-241 -->
-		<dependency org="antlr"             name="antlr"     rev="2.7.6"    transitive="false"/>
+		<dependency org="org.antlr"         name="antlr"     rev="3.4-complete"    transitive="false"/>
 		<dependency org="asm"               name="asm"       rev="3.1"    transitive="false"/>
 		<dependency org="asm"               name="asm-attrs" rev="2.2.3"    transitive="false"/>
     	<dependency org="cglib" name="cglib-nodep" rev="2.2" transitive="false" />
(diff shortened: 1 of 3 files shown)


alexzhk:
Could you @dkincade @hudak  please review my changes?

dkincade:
Notes from WingMan:
`Mon Jun 13 05:13:01 EDT 2016 - stdout - head: BUILD FAILED
Mon Jun 13 05:13:01 EDT 2016 - stdout - head: /home/buildguy/project/repository/build-res/subfloor.xml:903: syntax errors in ivy file: java.text.ParseException: [xml parsing: ivy.xml:163:42: cvc-complex-type.3.2.2: Attribute 'name' is not allowed to appear in element 'exclude'. in file:/home/buildguy/project/repository/ivy.xml`

dkincade:
@alexzhk I am not sure why WingMan in complaining about this ... but I think we should be safe with just the ORG portion of the exclude since it changed to "org.antlr" with the updated dependency. Give that a try and see if the results work for you.

core/ivy.xml
```diff
@@ -57,10 +57,11 @@
       		<exclude org="dom4j" name="dom4j" />
       		<exclude org="net.sf.ehcache" name="ehcache" />
       		<!-- CM-241 -->
-      		<exclude org="cglib" name="cglib" />      
+      		<exclude org="cglib" name="cglib" />
+			    <exclude org="antlr" name="antlr" />
```
dkincade: Please update this to be "module" as well...

PR #2948, opened 2016-06-10
[SP-2708] - [PDI-14835] - Adding extra role to the IUserRoleListService.getRoles and IUserRoleListService.getRolesForUser call

rmansoor:


rmansoor:
@mbatchelor please review

PR #2952, opened 2016-06-17
[SP-2680] Backport of PDI-14736 - MongoDB Input/Output steps fail whe…

mchen-len-son:
…n Authentication is used in MongoDB (5.4 Suite)

- added packages required by updated mongo-java-driver 2.13.3

according to this commit (https://github.com/mongodb/mongo-java-driver/commit/19ccc5f80d1f20b503ab92031a01427c22de7185) we should also remove some javax.security packages, but they are used by other bundles.


mchen-len-son:
This should be merged in a pair with https://github.com/pentaho/pentaho-platform-ee/pull/931.

PR #2958, opened 2016-06-24
[BISERVER-13116] API call asking internally for non existing pentaho/…

denis8703:
…js/postAuth.js file


denis8703:
@pamval @pentaho-nbaker @e-cuellar please review

mchen-len-son:
Guys, this is a clean-up after incomplete revert.
Should we work on test coverage anyway?

PR #2959, opened 2016-06-24
[BISERVER-12319] - User without any permissions on the Browse perspec…

apratkin:
…tive see only process indicators


apratkin:
@rmansoor  could you please review?

rmansoor:
@apratkin looks good

PR #2970, opened 2016-07-07
BACKLOG-7817 - Location prompt in report viewer: custom prompting dialog

EgorZhuk:
Just for Wingman. Please don't merge.

### DIFF
--- a/user-console/source/org/pentaho/mantle/client/commands/AdhocRunInBackgroundCommand.java
+++ b/user-console/source/org/pentaho/mantle/client/commands/AdhocRunInBackgroundCommand.java
@@ -1,0 +1,219 @@
+/*!
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License, version 2.1 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
+ * or from the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
+ */
+
+package org.pentaho.mantle.client.commands;
+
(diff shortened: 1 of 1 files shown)


user-console/source/org/pentaho/mantle/client/commands/AdhocRunInBackgroundCommand.java
```diff
@@ -0,0 +1,233 @@
...
+  @Override
+  protected void performOperation( boolean feedback ) {
+
+    RequestBuilder treeRequestBuilder = createTreeRequest();
+
+    try {
+      treeRequestBuilder.sendRequest( null, new RequestCallback() {
+
+        public void onError( Request request, Throwable exception ) {
+          MessageDialogBox dialogBox =
```
tmorgner: This error handling code (and its second variation with a response-status) is replicated multiple times. Condense it into a single method?

user-console/source/org/pentaho/mantle/client/commands/AdhocRunInBackgroundCommand.java
```diff
@@ -0,0 +1,233 @@
...
+
+  public void setJobId( String jobId ) {
+    this.jobId = jobId;
+  }
+
+  public String getJobId() {
+    return jobId;
+  }
+
+  @Override public String getOutputLocationPath() {
```
tmorgner: You are overriding plain getters from the parent class. Is that necessary, given that the parent has no logic in either getter or setter? You should be able to just use the base-versions instead. Removes a few lines of duplicate code.

user-console/source/org/pentaho/mantle/client/commands/AdhocRunInBackgroundCommand.java
```diff
@@ -0,0 +1,233 @@
...
+        public void onError( Request request, Throwable exception ) {
+          MessageDialogBox dialogBox =
+            new MessageDialogBox( Messages.getString( "error" ), exception.toString(), false, false,
+              true );
+          dialogBox.center();
+        }
+
+        public void onResponseReceived( Request request, Response response ) {
+          if ( response.getStatusCode() == Response.SC_OK ) {
+            String folderId = null;
```
tmorgner: Could we extract this block into a "parseFolderId(Response response)" method? Makes the block a lot easier to read. Along with moving the error handling into extra methods (see above), this then becomes a trivially simple to read method. I love trivially simple - it matches my brain :)

EgorZhuk:
@tmorgner ok, I will refactor it. Not sure for getters and setters - there are some problems with GWT, but I will try to understand what is going on.

tmorgner:
If the getter and setter cause problems, then screw that part and play it safe. But we can at least make the main method a hell of a lot more readable via simple refactorings.

EgorZhuk:
@tmorgner in fact we needed only solutionPath inside child class

PR #2977, opened 2016-07-12
[BACKLOG-8799] - Removing pdfbox …

dkincade:
…as a dependency as it is not used (and not has a security vulnerability)


dkincade:
@pentaho-nbaker please review 😄

PR #2978, opened 2016-07-12
[BACKLOG-9273] Create a configuration property that controls the show…

pedrofvteixeira:
…/hide of the 'Tools' sub-menu. This sub-menu should be hidden by default

	- added a 'forceFocus' mechanism to ensure the 'kettle status' always get focus, when the menu item is clicked


pedrofvteixeira:
@pdesai16 @mracine @niehusa @pentaho-jlien

PR #2983, opened 2016-07-21
[BACKLOG-9340] - PIR: The wait dialog “Just a few moments please” app…

dzmitry-hryhoryeu:
[BACKLOG-9340] - PIR: The wait dialog “Just a few moments please” appears before the progress bar.

@tmorgner
please review

### DIFF
--- a/user-console/source/org/pentaho/mantle/client/ui/tabs/MantleTabPanel.java
+++ b/user-console/source/org/pentaho/mantle/client/ui/tabs/MantleTabPanel.java
@@ -17,11 +17,24 @@
 
 package org.pentaho.mantle.client.ui.tabs;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-
+import com.google.gwt.core.client.GWT;
+import com.google.gwt.dom.client.NodeList;
+import com.google.gwt.http.client.Request;
+import com.google.gwt.http.client.RequestBuilder;
+import com.google.gwt.http.client.RequestCallback;
+import com.google.gwt.http.client.RequestException;
+import com.google.gwt.http.client.Response;
+import com.google.gwt.json.client.JSONObject;
+import com.google.gwt.json.client.JSONParser;
+import com.google.gwt.user.client.Element;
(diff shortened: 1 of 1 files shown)


user-console/source/org/pentaho/mantle/client/ui/tabs/MantleTabPanel.java
```diff
@@ -818,7 +820,44 @@ public void onResponseReceived( Request request, Response response ) {
...
+    }
+  }
+
+  /**
+   * Queries the server once to see whether PIR async-mode is enabled. If it is enabled,
+   * the report-viewer provides it own loading screen and the default "please wait" screen interferes
+   * with the user experience of that. Therefore we disable the wait-screen that comes up when a new
+   * page opens. @see case BACKLOG-9340 for details.
+   */
+  public void fetchAsyncModeEnabledStatusForPIRPlugin() {
```
tmorgner: This code is almost identical to the other method. Refactor so that the changing bit (the static text appended to the URL) is provided as parameter. This will greatly reduce the amount of duplicate code you have in this class.

user-console/source/org/pentaho/mantle/client/ui/tabs/MantleTabPanel.java
```diff
@@ -818,7 +823,11 @@ public void onResponseReceived( Request request, Response response ) {
           if ( response.getStatusCode() == Response.SC_OK ) {
             JSONObject jsonObject = (JSONObject) JSONParser.parseLenient( response.getText() );
             if ( jsonObject != null ) {
-              isAsyncModeEnabled = jsonObject.get( "supportAsync" ).isBoolean().booleanValue();
+              if ( path.equals( REPORT_VIEWER_ASYNC_CONFIG_PATH ) ) {
```
tmorgner: Make the method return a boolean, and you can avoid this whole "if" statement. 

    return jsonObject.get( "supportAsync" ).isBoolean().booleanValue();

and then when you *call* the method, assign the return value to the correct value. Return false in the catch-block in case there are any errors.

dzmitry-hryhoryeu:
@tmorgner please review.

PR #3007, opened 2016-08-01
[BACKLOG-9732] Remove Execute permission from Authenticated role and …

pedrofvteixeira:
…grant that permission to the Power User role

	- leaving this one commented-out as a means to ease the documentation on how-to re-enable execute permission for the authenticated role
	- all unit tests regarding the adding of the execute permission to a list of roles function just as well nowadays


pedrofvteixeira:
@pentaho/rogueone please review.

To be merged alongside https://github.com/pentaho/pentaho-platform-ee/pull/949

PR #3008, opened 2016-08-01
[BISERVER-12069] MdxBaseComponent forces all catalog URLs to start wi…

mchen-len-son:
…th http: or solution:

- if the filename is ambiguous, mdxBaseComponent should check file existence in repository, and if file exists, add solution prefix, otherwise, treat the resource as from file system


pamval:
@mchen-len-son  coverage drop, please takee a look

pamval:
@mchen-len-son  I'm merging this so we can validate tomorrow and call code freeze. But please issue a pull request asap fixing the drop in code coverage

thanks

PR #3009, opened 2016-08-01
[SP-2791] Backport of BISERVER-12069 - MdxBaseComponent forces all ca…

mchen-len-son:
…talog URLs to start with http: or solution: (6.1 Suite)

- if the filename is ambiguous, mdxBaseComponent should check file existence in repository, and if file exists, add solution prefix, otherwise, treat the resource as from file system


pamval:
@mchen-len-son  I'm merging this so we can validate tomorrow and call code freeze. But please issue a pull request asap fixing the drop in code coverage

thanks

PR #3011, opened 2016-08-02
[BISERVER-12069] MdxBaseComponent forces all catalog URLs to start wi…

mchen-len-son:
…th http: or solution:

- added a unit test to increase the test coverage


mchen-len-son:
@pamval 
to clarify what was changed:
at previous PR (#3008) the method coverage dropped, since I added two methods (factory methods for testing) not covered - thats why it had -10% method coverage drop.

here I simple covered one of the newly added methods. That gives us +5%

pamval:
can we get the other 5%? :)

mchen-len-son:
this is one line factory method. it's too simple to break.
I'm not sure if we should work on it.

pamval:
ok, but the "too simple to break" will come back to bite you (it usually does).

PR #3012, opened 2016-08-02
[SP-2791][BISERVER-12069] MdxBaseComponent forces all catalog URLs to…

mchen-len-son:
… start with http: or solution:

- added a unit test to increase the test coverage


mbatchelor:
@mchen-len-son - please correct this PR. Thanks.

mchen-len-son:
those static imports left are legal ones

PR #3016, opened 2016-08-03
[BACKLOG-8389] - Removing stray antlr jar …

dkincade:
…since antlr-complete is already there and at the proper version


dkincade:
This is working locally. The BASE build failed for a compile error and the HEAD build failed with a JVM error ... re-running wingman...

dkincade:
This work was completed with another PR

PR #3021, opened 2016-08-09
[BACKLOG-8389] - Upgrade antlr to latest stable version across Pentaho product suite - changes for pentaho platform repository

alexzhk:
Set transitive to false for hibernate-core and explicitly added necessary dependencies to the core module;
Removed the exclude tag for antlr from the extensions module.


alexzhk:
Could you @dkincade @hudak please review my changes?

PR #3023, opened 2016-08-09
[TEST] Test coverage for AclPublish

chiltonj:
@pentaho/rogueone Should now run as part as regular build


pedrofvteixeira:
@chiltonj PR looks good and wingman has given us its :+1: . Merging.

PR #3024, opened 2016-08-10
[TEST] Rename test to conform with convention: "Test" as suffix

chiltonj:
@pentaho/rogueone Renamed with 'Test' as suffix.


pedrofvteixeira:
PR looks good and wingman has given us it :+1: . Merging.

PR #3027, opened 2016-08-11
[ANALYZER-3220] - Impossible to add into dashboard analyzer report wi…

apratkin:
…th prompt that contains special chars.

### DIFF
--- a/core/src/org/pentaho/platform/util/beans/BeanUtil.java
+++ b/core/src/org/pentaho/platform/util/beans/BeanUtil.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2013 Pentaho Corporation..  All rights reserved.
+ * Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
  */
 
 package org.pentaho.platform.util.beans;
@@ -103,8 +103,13 @@ public class BeanUtil {
    * @return <code>true</code> if the bean property can be written to
    */
   public boolean isWriteable( String propertyName ) {
-    return propUtil.isWriteable( bean, propertyName )
-        || ( propUtil.getResolver().isIndexed( propertyName ) && propUtil.isReadable( bean, propertyName ) );
+    try {
+      return propUtil.isWriteable( bean, propertyName )
+          || ( propUtil.getResolver().isIndexed( propertyName ) && propUtil.isReadable( bean, propertyName ) );
+    } catch ( IllegalArgumentException exception ) {
+      logger.debug( "Exception rose. Return false value", exception );
+      return false;
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


apratkin:
@pentaho-nbaker could you please review my commit?

mchen-len-son:
@rmansoor @rfellows please take a look.

mchen-len-son:
@rmansoor @rfellows @bennychow could you please take a look?
Thanks.

bennychow:
With this change, at least you can open the Analyzer report in the dashboard... however, does the prompt with special characters in its name actually work?  I.e. can you create a dashboard prompt and hook that up to the Analyzer filter prompt?

core/src/org/pentaho/platform/util/beans/BeanUtil.java
```diff
@@ -103,8 +103,12 @@ public Object getValue( String propertyName ) throws IllegalAccessException, Inv
    * @return <code>true</code> if the bean property can be written to
    */
   public boolean isWriteable( String propertyName ) {
-    return propUtil.isWriteable( bean, propertyName )
-        || ( propUtil.getResolver().isIndexed( propertyName ) && propUtil.isReadable( bean, propertyName ) );
+    try {
+      return propUtil.isWriteable( bean, propertyName )
+          || ( propUtil.getResolver().isIndexed( propertyName ) && propUtil.isReadable( bean, propertyName ) );
+    } catch ( IllegalArgumentException exception ) {
+      return false;
```
mbatchelor: I think there should be some logging here - at least debug-level logging. This is swallowing the exception completely.

apratkin:
@bennychow 
I tested with following steps:
1.	Rename some measure to ‘
2.	Add this measure
3.	Add some columns
4.	Filter by measure ‘
5.	Rename some column to ‘’ and add parametrized filter with that column(I did not found if we could create parametrized filter for measure)
6.	Added create report to new dashboard
7.	Change value of filter parameter

apratkin:
@bennychow 
I tested with following steps:
1.	Create new analysis report
2.	Create filter ‘contains’ with named parameter ‘[~!@#$%^&*(){}|.,]-=_+|;'"?<>~`:’ and the same value.
3.	Create new dashboard and add report. Verify no data displayed because of filter (it is expected)
4.	Change filter value to some normal value and verify it shows some result.

apratkin:
@mbatchelor 
About logging. I suppose it make sense add debug level logging. 
Inside PropertyUtilsBean. isWriteable method we have following code:
            try {
                nestedBean = getProperty(bean, next);
            } catch (IllegalAccessException e) {
                return false;
            } catch (InvocationTargetException e) {
                return false;
            } catch (NoSuchMethodException e) {
                return false;
            }

Based on this I decided do not log this exception.

apratkin:
Updated commit with debug logging

apratkin:
@pedrofvteixeira could you please take a look?

apratkin:
@pamval could you please take a look?

bennychow:
+1 Looks good to me now

mchen-len-son:
Checked locally.
Merging, since no objections were raised after the PR was updated.

PR #3031, opened 2016-08-16
[BACKLOG-10020] - Clear report content cache

EgorZhuk:
@tmorgner please review

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -3,8 +3,7 @@
   <component name="EntryPointsManager">
     <entry_points version="2.0" />
   </component>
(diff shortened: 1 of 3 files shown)


tmorgner:
Seems there is an API change in the platform base classes causing this build to fail.

tmorgner:
Egor, not sure this is an API change after all. The file in question (RepositoryFile) has not been changed in years. 

And the compile error happens in a separate part: ExportManifestEntity, which has not been changed since 2014. Weird.

EgorZhuk:
@tmorgner wingman's magic

PR #3039, opened 2016-08-25
BISERVER-13330 When using backup/restore repository from the import-export utility, all schedules are restored with a status of active regardless of original status

AliaksandrDrebenchuk:
will restored normal like normal and all other job states like paused


AliaksandrDrebenchuk:
@rfellows @pamval @rmansoor @pentaho-nbaker please take a look

PR #3042, opened 2016-08-25
SP-2863 L10N - Home folder should start with upper-case "H" letter when localized

AliaksandrDrebenchuk:
default node title and locale no longer linked


mchen-len-son:
Double-checked locally. Tests are passing, checkstyle is OK.
Merging.

PR #3048, opened 2016-08-30
[BISERVER-13309] Perminent JCR folder deletion leaving residual orpaned

denisprotopopov:
ACL Information that can reappear
 - now session is cleaning at the end of importfile
 - added unit tests

### DIFF
--- a/extensions/src/org/pentaho/platform/plugin/services/importer/RepositoryFileImportFileHandler.java
+++ b/extensions/src/org/pentaho/platform/plugin/services/importer/RepositoryFileImportFileHandler.java
@@ -69,10 +69,7 @@ public class RepositoryFileImportFileHandler implements IPlatformImportHandler {
   }
 
   public ImportSession getImportSession() {
-    if ( importSession.get() == null ) {
-      importSession.set( ImportSession.getSession() );
-    }
-    return importSession.get();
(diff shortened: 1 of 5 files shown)


mbatchelor:
@denisprotopopov - would like to see in the JIRA case what lead you to this fix. From the very explicit description, I expected to see a fix in DefaultDeleteHelper as it's claiming there are orphan files. Your fix affects import sessions, but doesn't handle the stranded file (if there is in fact an orphaned file).

denisprotopopov:
@mbatchelor - as i understood during investigation issue that problem arise during import file because of ImportSession could return old ACL from previuos import and this old ACL will use in current import of file

mchen-len-son:
@mbatchelor, there is no orphaned file left after importing, but the corresponding acl object is not deleted from thread-local variable in ImportSession, and can therefore be available (when it shouldn't be!) for subsequent requests, if the same thread is chosen by tomcat thread executor. Thus, we should clear import session when exiting.

mbatchelor:
Great - I copied your response into the BISERVER-13309 JIRA case. However, it claims you have a newly broken test. Can you please fix?

denisprotopopov:
all tests successful, looks like nondeterministic test

PR #3051, opened 2016-09-01
Fixed BISERVER-12730 Untranslated strings in Blockout time table. Add…

ArtsiomYurhevich:
BISERVER-12730 Untranslated strings in Blockout time table.  Added conf parameters in order to support localized dates.


ArtsiomYurhevich:
@mdamour1976  , @pentaho-nbaker, @rmansoor   Could you please take a look?

PR #3052, opened 2016-09-01
[BISERVER-13332] The ACLs selection disappear from all groups/roles on BA server after doing a RESTORE using the Import-Export utility

Advent51:
- fixed
@rmansoor 
@pentaho-nbaker
Wingman failed test, that is not connected with my changes (at first run).


mbatchelor:
Andrei,
In SP triage, we determined I shouldn't have merged this code. The fix may work for LDAP and Jackrabbit, but since our authentication/authorization platforms are pluggable, there is an infinite number of possible role providers available. This should use a bean and injection for deciding the strategy.
Please: 
1- Revert the commit 
2- Reimplement using beans and injection 
3- Make sure you also include a strategy for JDBC roles as that's also provided as an option for our customers today.

PR #3059, opened 2016-09-06
[BACKLOG-10534][BACKLOG-10159] Operation Permissions from plugins are…

pedrofvteixeira:
… not applied in due time to pentaho-server

	- Follow-up to the work done a while back for [BACKLOG-7363] Operation Permissions from plugins are not used in merged server
	- cacheManager gets populated with platform permissions *prior* to a plugin's systemListener.onReload() kicks in ( and while loaded = false )


pedrofvteixeira:
@pentaho/rogueone , @rmansoor please review

This ends up acting as a follow-up to the work done a few months ago for BACKLOG-7363 ( https://github.com/pentaho/pentaho-platform/pull/2893 ).

While this work does indeed solve the issue for slower platform startups ( such as when there's a need to first create users and roles in the repository ), subsequent server restarts are quicker and cacheManager gets filled in with the operation permissions list *before* a platform plugin has time to register its own IAuthorizationAction via IoC ( see https://github.com/pentaho/data-access/blob/6.1.0.5-R/package-res/plugin.spring.xml#L276-L282 )

PR #3061, opened 2016-09-06
[BISERVER-13323]-Fixed hardcode 'waitSeconds' timeout setting in PentahoWebContextFilter. Now…

ArtsiomYurhevich:
… it comes from the pentaho.xml

### DIFF
--- a/extensions/src/org/pentaho/platform/web/http/filters/PentahoWebContextFilter.java
+++ b/extensions/src/org/pentaho/platform/web/http/filters/PentahoWebContextFilter.java
@@ -21,6 +21,7 @@ import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.commons.lang.StringUtils;
 import org.owasp.esapi.ESAPI;
 import org.owasp.esapi.Encoder;
+import org.pentaho.platform.api.engine.ICacheManager;
 import org.pentaho.platform.api.engine.IPentahoRequestContext;
 import org.pentaho.platform.api.engine.IPluginManager;
 import org.pentaho.platform.api.engine.IPentahoSession;
@@ -50,9 +51,9 @@ import java.util.Map;
 
 /**
  * If the request is searching for a webcontext.js, it writes out the content of the webcontext.js
- * 
+ *
  * @author Ramaiz Mansoor
- * 
+ *
  */
 public class PentahoWebContextFilter implements Filter {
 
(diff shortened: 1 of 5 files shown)


ArtsiomYurhevich:
@mdamour1976 , @pentaho-nbaker, @pedrofvteixeira Could you please take a look?

mbatchelor:
Please fix the checkstyle violations and add a unit test for the variable.

extensions/src/org/pentaho/platform/web/http/filters/PentahoWebContextFilter.java
```diff
@@ -369,6 +368,20 @@ private void printResourcesForContext( String contextName, OutputStream out, Htt
 
   }
 
+  public void printRequireJsCfgStart( OutputStream out ) throws IOException {
+
+    int waitTime;
+
+    try {
```
mbatchelor: Since this is invoked a lot, I'd rather you read the setting from the pentaho.xml and store it in the cache (see CacheManager) and use it from there. If a customer needs to refresh the value, they can clear the cache.

ArtsiomYurhevich: org.pentaho.platform.engine.core.system.PentahoSystem uses org.pentaho.platform.engine.core.system.SystemSettings which already uses  cache for system variables. See  [https://github.com/pentaho/pentaho-platform/blob/0e9201a186de07d2e8693a4360613e14597a5b79/core/src/org/pentaho/platform/engine/core/system/SystemSettings.java#L80-L91] .  Then user may refresh variables clicking Tools->Refresh->System Settings  in user console. The refresh command [https://github.com/pentaho/pentaho-platform/blob/0e9201a186de07d2e8693a4360613e14597a5b79/core/src/org/pentaho/platform/engine/core/system/SystemSettings.java#L185-L187] only cleans the cache map and doesnt work with CacheManager. Whether it is necessary in this case to use CacheManager.

mbatchelor: It caches the whole pentaho.xml. Every time you read the setting, it's traversing the XML to get the discrete setting. So yes - it pulls the xml file out of the cache, and then parses it to get the discrete value. My preference is for this method to extract the discrete value from the cache if it's there, or read the setting if it's not.

ArtsiomYurhevich:
@mdamour1976 , @pentaho-nbaker, @pedrofvteixeira, @mbatchelor  Could you please take a look?

PR #3072, opened 2016-09-09
[PPP-3536] - Use of vulnerable component org.codehaus.groovy v.1.8.0 …

YuryBY:
…CVE-2015-3253

 - groovy is replaced by groovy-all

@mbatchelor, @pamval, could you please take a look?

### DIFF
--- a/extensions/ivy.xml
+++ b/extensions/ivy.xml
@@ -220,7 +220,7 @@
 
 
     <dependency org="com.lowagie" name="itext-rtf" rev="${dependency.itextrtf.revision}" changing="true" transitive="false"/>
-    <dependency org="org.codehaus.groovy"  name="groovy"  rev="2.4.7"  changing="true" transitive="false"/>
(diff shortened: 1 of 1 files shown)


lucboudreau:
I get that using groovy-all will override the downstream builds, but I think we should fix metaverse, which depends on gremlin, who in turn pulls the outdated groovy dependency.

http://cerberus.pentaho.com/job/master-snapshot/view/control/job/dependency-report/lastSuccessfulBuild/artifact/newDependencyMgmt/dependency-report.html#org.codehaus.groovy:groovy

YuryBY:
@lucboudreau 

mvn dependency:tree result is:

[INFO] +- com.tinkerpop:frames:jar:2.5.0:test
[INFO] |  \- com.tinkerpop.gremlin:gremlin-groovy:jar:2.5.0:test
[INFO] |     +- org.apache.ivy:ivy:jar:2.3.0:test
[INFO] |     +- org.codehaus.groovy:groovy:jar:1.8.9:test

So that I’st test-time dependency.

PR #3077, opened 2016-09-13
[BISERVER-13357] After Backup/Restore, Authenticated user can access other users' home folders and run reports - when using LDAP auth

Advent51:
-fixed


lucboudreau:
Please check if you can help make this test more reliable. 

Thanks.

mchen-len-son:
@Advent51,
i've checked the failing test - let's increase thread timeout to 400ms like you proposed, though I don't like it, otherwise we should completely re-work the test, which would be out of the scope of the current issue.

Advent51:
@mchen-len-son Ok, I made it worser, but let's try.
@rmansoor @pentaho-nbaker please review

Advent51:
In case of reverting #3052 this PR must be closed. This case will be reworked in new PR connected with BISERVER-13332

PR #3090, opened 2016-09-21
[BISERVER-13332] The ACLs selection disappear from all groups/roles on BA server after doing a RESTORE using the Import-Export utility

Advent51:
-reworked for using multiple auth providers.


Advent51:
@mbatchelor 
Marc, please, check it out. Is that what you asked me for?

PR #3103, opened 2016-09-27
[BACKLOG-9900] - 7.0 Localization for PUC.

dzmitry-hryhoryeu:
[BACKLOG-9900] - 7.0 Localization for PUC.


dzmitry-hryhoryeu:
@tmorgner please review

tmorgner:
The files are not supposed to be renamed, just updated. There is a difference between a locale selector "fr" and "fr_FR". The first covers all french dialects (including swiss-french and french speaking countries, like Canada) while the second only covers mainland France.

dzmitry-hryhoryeu:
@tmorgner Changed. Please review.

PR #3115, opened 2016-09-30
[BACKLOG-10623] Update springframework 4.3.2 and spring security to 4…

pdesai16:
….1.3

               - Added an exclude for spring-2.0 and jaxws-rt in
		 assembly/ixy.xml


pdesai16:
please review @pentaho/rogueone

PR #3119, opened 2016-10-03
[BACKLOG-11271] The applicationContext-spring-security-memory.xml fil…

pedrofvteixeira:
…e contains an incorrect default password for suzy

	- removed that 'value="' snippet


pedrofvteixeira:
@pentaho/rogueone

PR #3121, opened 2016-10-04
[BACKLOG-9649]-System is not able to delete the karaf folder created in the temporary folder location

stepanovdg:
* Used io deleteOnExit except of nio FileUtils
* Changed symlinks failure logginf from error to warns (as we still copy that files)


stepanovdg:
@rmansoor @tkafalas Could you please take a look?

PR #3130, opened 2016-10-05
[BACKLOG-10623] Upgrade springframework and spring security

pdesai16:
- Excluded the dependency streambuffer.jar that is pulled in by
	  jaxws-spring


pedrofvteixeira:
👍

PR #3082, opened 2016-09-16
[BISERVER-12715] L10N - "Created" and "Last modified" date of file's Properties ignores localization

Advent51:
-fixed

PR #3136, opened 2016-10-06
[BISERVER-12715] L10N - "Created" and "Last modified" date of file's …

mchen-len-son:
…Properties ignores localization

revert https://github.com/pentaho/pentaho-platform/pull/3082
regression was found [BACKLOG-11280]


mchen-len-son:
@pamval please take a look

PR #3152, opened 2016-10-10
PRD-5515 PRPT reports crashed if it has Cyrillic Characters on paramerters on JBoss Manual installation.

AliaksandrDrebenchuk:
URL parameters encoding is fixed

### DIFF
--- a/extensions/src/org/pentaho/platform/web/http/filters/PentahoPathDecodingFilter.java
+++ b/extensions/src/org/pentaho/platform/web/http/filters/PentahoPathDecodingFilter.java
@@ -1,6 +1,29 @@
+/*!
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License, version 2.1 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
+ * or from the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
+ */
+
 package org.pentaho.platform.web.http.filters;
 
-import org.apache.commons.lang.StringUtils;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
@@ -10,7 +33,17 @@ import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletRequestWrapper;
-import java.io.IOException;
+import javax.ws.rs.HttpMethod;
+import javax.ws.rs.core.MultivaluedMap;
+
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.http.NameValuePair;
+import org.apache.http.client.utils.URLEncodedUtils;
+
+import com.google.common.collect.Iterators;
+import com.sun.jersey.core.util.MultivaluedMapImpl;
 
 /**
  * This filter is designed to work around a limitation of some web servers which prohibit the encoded form of / and \
@@ -21,6 +54,9 @@ import java.io.IOException;
  * Created by nbaker on 4/8/14.
  */
 public class PentahoPathDecodingFilter implements Filter {
+
+  private static final Log LOGGER = LogFactory.getLog( PentahoPathDecodingFilter.class );
+
   @Override
   public void init( FilterConfig filterConfig ) throws ServletException {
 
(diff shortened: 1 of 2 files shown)


extensions/src/org/pentaho/platform/web/http/filters/PentahoPathDecodingFilter.java
```diff
@@ -40,8 +75,39 @@ public void destroy() {
   }
 
   private static final class EncodingAwareHttpServletRequestWrapper extends HttpServletRequestWrapper {
+
+    StringKeyStringValueIgnoreCaseMultivaluedMap parameterMap = null;
```
mbatchelor: I believe variables here will be extremely bad - this is a static final class, and every page runs though this filter. So these variables would have to minimally be threadlocal variables as one request after another would blast these things. This isn't threadsafe in any way that I can see.

extensions/src/org/pentaho/platform/web/http/filters/PentahoPathDecodingFilter.java
```diff
@@ -58,5 +124,40 @@ public String getRequestURI() {
...
+      List<String> values = parameterMap.get( name );
+      return values.toArray( new String[values.size()] );
+    }
+
+    @Override
+    public Enumeration<String> getParameterNames() {
+      if ( parameterMap == null ) {
+        return super.getParameterNames();
+      }
+      return new Vector<String>( parameterMap.keySet() ).elements();
```
mbatchelor: Why use Vector here ?

extensions/src/org/pentaho/platform/web/http/filters/PentahoPathDecodingFilter.java
```diff
@@ -40,8 +75,39 @@ public void destroy() {
...
     private EncodingAwareHttpServletRequestWrapper( HttpServletRequest request ) {
       super( request );
+
+      if ( HttpMethod.GET.equalsIgnoreCase( request.getMethod() ) ) {
+        String query = getQueryString();
+        if ( StringUtils.isNotBlank( query ) ) {
+          try {
+            List<NameValuePair> pairs =
+                URLEncodedUtils.parse( new URI( "/?" + query ), request.getCharacterEncoding() );
+            parameterMap = new StringKeyStringValueIgnoreCaseMultivaluedMap();
```
mbatchelor: I believe that URL parameters are required to be case sensitive. The RFC states:
6.2.2.1. Case Normalization

When a URI uses components of the generic syntax, the component syntax equivalence rules always apply; namely, that the scheme and host are case-insensitive and therefore should be normalized to lowercase. For example, the URI is equivalent to http://www.example.com/.

The other generic syntax components are assumed to be case-sensitive unless specifically defined otherwise by the scheme (see Section 6.2.3).

So I don't believe that the parameterMap should use a StringKeyStringValueIgnoreCaseMultivaluedMap.

extensions/src/org/pentaho/platform/web/http/filters/PentahoPathDecodingFilter.java
AliaksandrDrebenchuk: @mbatchelor For every new request we have a new rapper with new map inside. What the problem with threadsafe?

extensions/src/org/pentaho/platform/web/http/filters/PentahoPathDecodingFilter.java
AliaksandrDrebenchuk: @mbatchelor I didn't found how better to create the Enumeration from Set. But now i found the variant "Iterators.asEnumeration( parameterMap.keySet().iterator() );". Is it ok?

AliaksandrDrebenchuk:
@mbatchelor PR was updated

tmorgner:
This is wrong on so many levels. What happens if you have russian path names, or POST content?

The error is not in the code, the error is in how the JBoss server is configured. JBoss is a nasty piece of work at best of times, but if you use a off-the-shelf unconfigured JBoss to run our software, you are simply getting garbage. 

Pentaho and all modern browsers assume that Parameter are encoded as UTF-8. JBoss however assumes that parameters are provided as ISO-8859-1 encoded text. This works only if you are living in 1980 where Russian or Chinese people were not present on the Internet.

You are trying to fix a wrongly configured JBoss server. 

JBoss - unless configured explicitly - defaults to ISO-8859-1 encoding for parameters. (as documented here: https://docs.jboss.org/jbossweb/3.0.x/config/http.html ) This is what mangles the incoming content and breaks your requests.

Reconfigure your JBoss to use UTF-8, like everyone else in 2016, and your problem goes away.

If you do implement a fix like this, you are most likely breaking the world for everyone else. 

PS: Check out the default Tomcat configuration that we ship and you will find that there we actually do configure the URL-encoding explicitly to be UTF-8. 


So in recap: To fix this "bug" follow the following steps:

(1) Close this Pull-Request immediately.
(2) Learn how to configure JBoss to use UTF-8 as default encoding for URLs
(3) Document how to configure JBoss to use UTF-8 and attach that to the case.
(4) Close the case as "Not a bug" and hand it back to the customer service team, so that they can tell the customer how to configure their installation.

tmorgner:
Hint: I'm pretty sure that "domain/configuration/domain.xml" contains the connector element that you need to configure.

AliaksandrDrebenchuk:
@tmorgner Of cource the configurations it's the first what I viewed.

AliaksandrDrebenchuk:
@tmorgner 
Of course the configurations it's the first what I viewed. Like <Connector port="8090" URIEncoding="UTF-8"/>.
But much better the <Connector port="8090" useBodyEncodingForURI="true"> because in this case the web client can set another encoding.
Regarding my fix, I chose this way because:
- we can change the XML files only for major releases.
- filter is works as for jboss as for tomcat.
- filter uses encoding from header instead constant value.
If you still sure that XML changes will be better, please confirm your own decision.

tmorgner:
No, in that case you are friggin' breaking every standard conforming web-client out there. It is 2016, and all major browsers have agreed that parameter are encoded as UTF-8 (@see the relevant HTML5 standard here: https://dev.w3.org/html5/spec-LC/urls.html ). 

Heck, there are enough dangers and red-flags here to drive a bus loaded with bugs through. 

For starters: You are recoding URL parameter, but not posted Form-content. You are hoping that the character encoding is specified. If may not be at all. GET parameter had been defined to be ISO-8859-1 encoded in hte old days, so you would have to guess now: Is that request UTF or is it ISO-8859-1. An you better hope you get it right all the time. 

So please dont do it. If anything, teach the customer how to deploy their applications correctly and work with the Documentation team to amend the existing documentation to include these settings.

And for "We are not supposed to change configuration files": In a *manual* installation it is absolutely OK to ask them to change their *server's* configuration files as needed. This is part of the normal deployment process.

AliaksandrDrebenchuk:
@tmorgner Yes, I understand that the encoding header can be missed. But in this case my filter uses UTF-8 because the PentahoAwareCharacterEncodingFilter sets DEFAULT_CHAR_ENCODING (UTF-8). Regarding XML files, I will glad to change them when it necessary, it's normal to change the spring xml files often for example. Just my XML changes never was approved earlier.

tmorgner:
Sorry, but I am not going to ever merge this, as this should be fixed by altering the configuration of the JBoss server where the Pentaho platform is installed, and by documenting that UTF-8 URL-Encoding is a  built-in and hard-coded setting that the platform and all dependent code expect to work correctly. (So: Document that we adhere to standard HTML5 and expect customer installations to do so too. This is not an unreasonable request. )

Our pre-configured Tomcat that we ship as part of the default implementation works fine. A correctly configured JBoss (with the encoding set in the server's configuration) will work fine, a WebSphere and all other supported servers work fine too (if configured correctly).

The correct fix for a broken configuration/documentation experienced by a subset of customers (AFAIK one!) does not justify a dangerous and wide-ranging change on all other server configurations if that one customer can actually solve the problem by fixing their server configuration. 

If you dont aggree with any of that, then please feel free to reach out to your team-lead or the engineering leadership team to get this issue solved.

AliaksandrDrebenchuk:
@tmorgner I will agree with any your decision. Thanks for your descriptions about all nuances.

PR #3154, opened 2016-10-10
[BACKLOG-11242] - Pentaho server will fail to start if a non-pooled c…

andrei-viaryshka:
…onnection is defined for a dialect that is not present


andrei-viaryshka:
@mkambol Could you please take a look?

PR #3168, opened 2016-10-14
[BACKLOG-11592] Ensure all PDI plugins are in Pentaho Server

pedrofvteixeira:


pedrofvteixeira:
@smaring , @lgrill-pentaho please review. 

Updated as per your recommendation ( to move from one dependency build key in .properties to a generic dependency.kettle.revision )

CC @pentaho/rogueone

PR #3170, opened 2016-10-17
[BACKLOG-11618] - Incorrect Data source name

apratkin:


pamval:
@apratkin in the previous commit you introduced a method that appends .xmi to the name of the domain if it's not there. Now we seem to be introducing another method, called after the first one, that removes the .xmi from the name. Is that accurate?

apratkin:
@pamval yes, it is accurate. I am going to add JUnits and will write comment how it was tested.

apratkin:
@pamval

apratkin:
@pamval I added JUnits and made some improvements. 

I tested new fix with steps from BACKLOG-11618 and BISERVER-13273 (simple steps at comment)
Additional testing:
I created all possible kind of data sources (CSV, SQL query, db table). 
I edited each data source and verified no new additional data sources were created. 
Verified PIR displays valid names for data sources.
Run import-export.bat (backup). Remove all data sources. Run import-export.bat (restore)

PR #3171, opened 2016-10-17
[BACKLOG-11059] Introducing extension of the SE-JCR Spring Transactio…

pentaho-nbaker:
…nManager which captures information about the transaction being created to inform session creation later.


rmansoor:
wingman failed, so ran the test locally. Deployed the jar and xml file in the latest pentaho server build and it looks fine. processing the pr now

PR #3172, opened 2016-10-18
[BACKLOG-10623] Cleaning up unnecessary spring xsd version

pedrofvteixeira:
- these files have nothing spring about then, except for the header
	- they are free to remain unchanged / unaltered


pedrofvteixeira:
@CodeOnCoffee , @rmansoor , @pentaho/rogueone please review

pedrofvteixeira:
@CodeOnCoffee mentioned that we don't need to back this down. It's fine the way it is and we'll just let it be new, for new installations.

Closing this PR.

PR #3175, opened 2016-10-18
[BACKLOG-11639] Updated link for PUC documentation. Added support for…

carlosrusso:
… external documentation url starting with http(s):


carlosrusso:
Note: This solution does not remove the documentation that ships with the pentaho-server.

PR #3187, opened 2016-10-20
[BISERVER-12719] When creating a schedule the dialog window where thetype of schedule is selected isn't fully localized

Advent51:
-additional fix


Advent51:
@pamval @mbatchelor please review

PR #3192, opened 2016-10-20
[BISERVER-12719] When creating a schedule the dialog window where the type of schedule is selected isn't fully localized (7.0.0.0)

Advent51:
-additional fix
#3187


dkincade:
We need to revert this commit to 7.0.0.0 branch. This is not a 7.0.0.0 release blocker. This will need to wait until there is a formal 7.0 branch.

Advent51:
@dkincade it's an important PR, because it is impossible to filter schedules by status or type in localized environment, so @pamval and @mbatchelor know about it and give this PR a green light.

dkincade:
As an update ... the revert was reverted ... so this is IN the 7.0.0.0 branch

PR #3194, opened 2016-10-20
Revert "Revert "[BISERVER-12719] When creating a schedule the dialog window where the type of schedule is selected isn't fully localized (7.0.0.0)""

dkincade:
Reverts pentaho/pentaho-platform#3193


dkincade:
This issue is actually required for 7.0.0.0 GA as it is a regression in a 6.1 SP that is currently in progress. So it will go with the 6.1 SP in October. With this it must be included in 7.0.0.0 GA.

PR #3215, opened 2016-10-28
[BACKLOG-11964]-Bring the loading dialog back

EgorZhuk:
Reverts the changes from BACKLOG-7180  and BACKLOG-9340

@tmorgner  @pamval  @kcruzada


tmorgner:
Agreed. Its safer to revert than to hope'n'pray for a stable fix here.

PR #3228, opened 2016-10-31
[BACKLOG-12001] - Removed elasticsearch as a dependency. 

dkincade:
It was there only to support the Kettle elasticsearch loader which is a separate plugin and has its own version, and the version specific had a known security vulnerability. Since it was bad and not needed, it was removed


dkincade:
Replaced

PR #3252, opened 2016-11-04
[BACKLOG-12100]-Bring back refresh data cache endpoint

EgorZhuk:
Our main fix for the cache refresh works like a charm, but we forgot that the old refresh functionality now also uses the plugin endpoint.
So both refresh button and "Tools" work the same way - clean cache for one session :(
The fix is easy - bring back the old url for "Tools".

🙏 

@tmorgner @kcruzada @CodeOnCoffee @pamval

### DIFF
--- a/user-console/source/org/pentaho/mantle/client/commands/PurgeReportingDataCacheCommand.java
+++ b/user-console/source/org/pentaho/mantle/client/commands/PurgeReportingDataCacheCommand.java
@@ -35,9 +35,9 @@ public class PurgeReportingDataCacheCommand extends AbstractCommand {
   }
 
   protected void performOperation() {
-    final String url = GWT.getHostPageBaseURL() + "plugin/reporting/api/cache/clear"; //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


tmorgner:
Wait, isnt that the entry-point for the admin level "clear cache" function? I would assume that if the admin clears the cache, he wants it for all users, not just himself.  This would be in line with the other menu-entries on that menu - they all are global, server wide operations.

If we make this change here, we actually have no way in the platform to perform a global clearing of caches for all users.

EgorZhuk:
@tmorgner  actually now we "have no way in the platform to perform a global clearing of caches for all users." because both Tools and refresh button uses plugin endpoint

tmorgner:
Oh, ignore me. Its to early. Yes, we want this to be global and global means "/api/system/refresh .." it is.

tmorgner:
I need coffee :)

tmorgner:
Waiting for Wingman OK and then I merge.

PR #3253, opened 2016-11-04
[BACKLOG-12100]-Bring back refresh data cache endpoint

EgorZhuk:
#3252 port 

@tmorgner @kcruzada @pamval @CodeOnCoffee


tmorgner:
come on wingman, hurry up.

PR #3129, opened 2016-10-05
[BISERVER-12718] L10n - The name of sections are cropped for Japanese…

ivanpogodin:
… language in Dashboard

PR #3255, opened 2016-11-04
[SP-2960] Backport of BISERVER-12718 - L10n - The name of sections ar…

mchen-len-son:
…e cropped for Japanese language in Dashboard

 - fixed refresh button right aligning on the browser toolbar


mchen-len-son:
Browser toolbar is still breaking Japanese words in its header (in 6.1 only, in 7.0 and 7.1 everything works good). 
Simple removing add( GLUE ) method call (like we did for FilesToolbar in #3129) breaks refresh button aligning - so I replaced that method call with a more correct version IMHO. 

The change looks like a workaround, and should be placed in [Toolbar.java](https://github.com/pentaho/pentaho-commons-gwt-modules/blob/master/pentaho-gwt-widgets/src/org/pentaho/gwt/widgets/client/toolbar/Toolbar.java) of pentaho-commons-gwt-modules, but this would be a lower level change and should be tested before.

@pamval please review

PR #3259, opened 2016-11-04
[SP-2968] - Backport of BISERVER-13273 - Creation of duplicate dataso…

apratkin:
…urce in PUC (6.1 Suite)

PR #3263, opened 2016-11-05
[SP-2968] Backport of BISERVER-13273 - Creation of duplicate datasource in PUC

mchen-len-son:


mchen-len-son:
This is a fixed version of #3259 (fixed checkstyle violations).
Double-checked locally.
Merging.

PR #3266, opened 2016-11-05
[BACKLOG-12170]-Copy authorities to session roles when becoming user

EgorZhuk:
@pamval @pedrofvteixeira @kcruzada @tmorgner @CodeOnCoffee @kolinus


pedrofvteixeira:
👍

PR #3268, opened 2016-11-09
[BACKLOG-12212] - When runing in background or scheduling .prpti and …

dzmitry-hryhoryeu:
[BACKLOG-12212] - When runing in background or scheduling .prpti and prpt reports, output type is locked.

### DIFF
--- a/extensions/src/org/pentaho/platform/web/http/api/resources/FileResource.java
+++ b/extensions/src/org/pentaho/platform/web/http/api/resources/FileResource.java
@@ -89,6 +89,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.StringReader;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.nio.channels.IllegalSelectorException;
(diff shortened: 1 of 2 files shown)


dzmitry-hryhoryeu:
@tmorgner please review.

extensions/src/org/pentaho/platform/web/http/api/resources/FileResource.java
```diff
@@ -2228,9 +2228,10 @@ protected ByteArrayOutputStream getByteArrayOutputStream() {
     return new ByteArrayOutputStream();
   }
 
-  protected Document parseText( String text ) throws DocumentException {
+  protected Document parseText( ByteArrayOutputStream outputStream ) throws DocumentException {
```
tmorgner: We do not change public API in core classes for no good reason. A protected member can be used by plugins and other code and such changes are only allowed under very limited circumstances (a x.0 release, and the member has been deprecated before so that OEMs can adapt to the change.)

Stick with the string as parameter, then use a StringReader as parameter for the SAXReader. That will do the trick.

dzmitry-hryhoryeu:
@tmorgner please review.

dzmitry-hryhoryeu:
@tmorgner please review.

tmorgner:
Any chance we get a unit-test for that? Just to make sure that this change is covered and that we detect regressions on future changes. 

A simple: Give me a dummy XML, check that I get a parsed document  back should suffice.

Please make it a habit to include tests in all your changes.

dzmitry-hryhoryeu:
@tmorgner please review.

PR #3276, opened 2016-11-23
PDI-15515- Removing legacy ESAPI encoder

ArtsiomYurhevich:
Removing legacy ESAPI encoder from projects and replacing it by org.owasp.encoder


ArtsiomYurhevich:
Please, do not merge it!

PR #3281, opened 2016-11-25
[SP-3057][PDI-15689] Log level set for schedules do not preserve after doing a RESTORE using the import-export utility

AliaksandrShuhayeu:
[SP-3057][PDI-15689] Log level set for schedules do not preserve after doing a RESTORE using the import-export utility

- fixed problem restoring job parameters as pdi parameters
- added backup pdi parameters
- added restoring pdi parameters


AliaksandrShuhayeu:
@mdamour1976 Could you please review and merge it?
There are problems only in static imports in jUnit tests.

PR #3301, opened 2016-12-06
[SERVER-13407]-ValidationQuery and Pooling options not sent to DB whe…

denisprotopopov:
…n datasource is defined in Manage DataSources Perspective

-added processing for pooling ds case
-added unit test


denisprotopopov:
should be merge after https://github.com/pentaho/pentaho-connections/pull/16

PR #3307, opened 2016-12-09
BISERVER-13392 - Fixed processing favorites files' names with special…

ArtsiomYurhevich:
… characters in js. Added validation for file's name in createFile method in REST FileResource


ArtsiomYurhevich:
@mbatchelor, @mdamour1976   Could you please take a look?

PR #3312, opened 2016-12-18
BISERVER-13531 - Admins are unable to copy/paste into a hidden folder

ArtsiomYurhevich:
The issue is that when we click on 'paste' in a hidden folder , first, gwt paste command makes a request for the children of the folder without url parameter "showHidden" regardless of the fact whether "show hidden" option is selected or not. But since the folder is hidden it is unavailable without request param 'showHidden'. That causes NPE.


ArtsiomYurhevich:
@mdamour1976,@mbatchelor , @pamval Could you take a look please.

PR #3030, opened 2016-08-15
[BISERVER-13264] When using a custom AccessVoter, cut/paste does not …

mchen-len-son:
…check for write permission on destination
- added accessVoter consulting to file moving method
- added unit-test

PR #3317, opened 2016-12-21
[SP-3080][BISERVER-13264] When using a custom AccessVoter, cut/paste …

mchen-len-son:
…does not check for write permission on destination

- added accessVoter consulting to file moving method
- added unit-test


mchen-len-son:
@pamval @mbatchelor @rfellows please take a look.

This is a backport of #3030.

PR #3319, opened 2016-12-21
[BACKLOG-12212] - When runing in background or scheduling .prpti and …

YuryBY:
…prpt reports, output type is locked.

@pamval, @mchen-len-son, could you please take a look?


pamval:
can we change the commit message to read SP-2093?

YuryBY:
@pamval, the message is updated

PR #3321, opened 2016-12-22
[BISERVER-13379] Xaction Output Destination File throws error on PUC …

ivanpogodin:
…it fails on / (slash) Separator but writes files folder


ivanpogodin:
@rmansoor , could you review it, please?

PR #3344, opened 2017-01-13
[PPP-3629] - UploadService limit file extensions

mbatchelor:
[PPP-3630] - UploadService limit tmp upload


mbatchelor:
@pamval - could you please review and merge? Thanks.

PR #3361, opened 2017-01-20
[WIP][BACKLOG-13736] Inject OSGi requirejs manager configurations script

nantunes:
@pentaho/millenniumfalcon an early look.


nantunes:
Work changed to the context of BACKLOG-14145. Will do a new PR when appropriate.

PR #3365, opened 2017-01-23
[PPP-3410]-Add an API to remove glasspane listeners

EgorZhuk:
Depends on https://github.com/pentaho/pentaho-commons-gwt-modules/pull/510 https://github.com/pentaho/data-access/pull/834

### DIFF
--- a/user-console/src/main/java/org/pentaho/mantle/client/MantleApplication.java
+++ b/user-console/src/main/java/org/pentaho/mantle/client/MantleApplication.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2013 Pentaho Corporation..  All rights reserved.
+ * Copyright (c) 2002-2017 Pentaho Corporation..  All rights reserved.
  */
 
 package org.pentaho.mantle.client;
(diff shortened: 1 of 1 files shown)


user-console/source/org/pentaho/mantle/client/MantleApplication.java
```diff
@@ -117,7 +117,11 @@ public native void setupNativeHooks( MantleApplication mantle, LoginCommand logi
               $wnd.addDataAccessGlassPaneListener(callback);
           }
           //CHECKSTYLE IGNORE LineLength FOR NEXT 1 LINES
-          mantle.@org.pentaho.mantle.client.MantleApplication::addGlassPaneListener(Lcom/google/gwt/core/client/JavaScriptObject;)(callback);
+          return mantle.@org.pentaho.mantle.client.MantleApplication::addGlassPaneListener(Lcom/google/gwt/core/client/JavaScriptObject;)(callback);
+      }
+
+      $wnd.removeGlassPaneListener = function (uuid) {
```
tmorgner: Given that you are not actually accepting a GlassPaneListener object, and to keep it in line with the GWT-commons change, would it be better to name this function (and the private counterpart a few lines below) "removeGlassPaneListenerById"? 

Otherwise the next guy seeing that function will try to pass a listener-instance in and will be surprised by the errors he gets. A lack of a proper type system makes these pitfalls even more deadly.

tmorgner:
@EgorZhuk can you have a look at this?

EgorZhuk:
@tmorgner sure, after the SPs as long as we have a code feeze for them on Monday
By the way this one still need some work as far as we have another instance of glass pane in data access wich still causes leaks

tmorgner:
That glasspane needs double glazing and some sealant :)

tmorgner:
As sane as it could be. Data access bit is merged, so lets get on with it ..

PR #3370, opened 2017-01-23
[SP-3181][BISERVER-13444] - Scheduled PRPTs and PIR Reports are not r…

EgorZhuk:
…ecorded to pro_audit table.


EgorZhuk:
fixed dates

PR #3371, opened 2017-01-23
[BISERVER-12721] Profile pop up window has wrong size for translated "Log out" string

PashkovichAndrei:
Changed padding & text-align property values.


PashkovichAndrei:
@mbatchelor fix has been applied to other themes

PR #3367, opened 2017-01-23
[SP-3189] Backport of BISERVER-13379 - Xaction Output Destination Fil…

ivanpogodin:
…e throws error on PUC it fails on / (slash) Separator but writes files folder (6.1 Suite)

cherry-pick:
        [BISERVER-13379] Xaction Output Destination File throws error on PUC it fails on / (slash) Separator but writes files folder
cherry-pick:
        [BISERVER-13379] Xaction Output Destination File throws error on PUC it fails on / (slash) Separator but writes files folder

PR #3382, opened 2017-01-26
Sp 3189

ivanpogodin:
Backport of BISERVER-13379 - Xaction Output Destination File throws error on PUC it fails on / (slash) Separator but writes files folder (6.1 Suite)
Instead of https://github.com/pentaho/pentaho-platform/pull/3367


mchen-len-son:
Local build is ok. Some tests are failing locally, but they look unrelated to the changes.

PR #3384, opened 2017-01-27
[BACKLOG-12101] fix artifactId(s) on CE assemblies

smaring:
@rfellows @pedrofvteixeira

### DIFF
--- a/assembly/pentaho-server/src/main/descriptors/assembly.xml
+++ b/assembly/pentaho-server/src/main/descriptors/assembly.xml
@@ -7,17 +7,17 @@
   <baseDirectory>pentaho-server</baseDirectory>
   <files>
     <file>
-      <source>${project.root.build.directory}/${project.artifactId}/pentaho-solutions/system/dialects/hsqldb/applicationContext-spring-security-hibernate.properties</source>
+      <source>${project.root.build.directory}/pentaho-server/pentaho-solutions/system/dialects/hsqldb/applicationContext-spring-security-hibernate.properties</source>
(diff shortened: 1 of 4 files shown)


assembly/pentaho-server/src/main/descriptors/assembly.xml
```diff
@@ -7,48 +7,48 @@
   <baseDirectory>pentaho-server</baseDirectory>
   <files>
     <file>
-      <source>${project.root.build.directory}/${project.artifactId}/pentaho-solutions/system/dialects/hsqldb/applicationContext-spring-security-hibernate.properties</source>
+      <source>${project.root.build.directory}/pentaho-server/pentaho-solutions/system/dialects/hsqldb/applicationContext-spring-security-hibernate.properties</source>
```
pedrofvteixeira: http://jira.pentaho.com/browse/BACKLOG-10262

One of the requirements of the work of merging BA Server and DI Server into a single Pentaho Server was also that we would no longer get a differentiation between CE and EE distributables embedded in the folder-name. To sum it up, the folder is meant to be named "pentaho-server" ( regardless of it being CE or EE )  

Doing this would mean that the unzipped artifact would have a folder structure of <unzip-dir>/pentaho-server*-ce*/pentaho-solutions/(...)

smaring: no, that's why I set it so pentaho-server instead of the artifactId

pedrofvteixeira: Ugh... misread the change ... actually read it the other way around... yep, this PR is good by me 👍

PR #3398, opened 2017-02-03
[SP-3269] Backport of MONDRIAN-2511 - Remove JdbcPassword from Mondri…

stanislau-strelchanka:
[SP-3269] Backport of MONDRIAN-2511 - Remove JdbcPassword from MondrianException (7.0 Suite)

PR #3399, opened 2017-02-03
Sp 3270

stanislau-strelchanka:


scheduler/src/org/pentaho/platform/scheduler2/quartz/QuartzScheduler.java
```diff
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2013 Pentaho Corporation..  All rights reserved.
+ * Copyright (c) 2002-2017 Pentaho Corporation..  All rights reserved.
```
mbatchelor: #3398 and #3397 didn't update QuartzScheduler but this one does. Seems like a regression waiting to happen.

stanislau-strelchanka: my bad, grabbed extra commits

PR #3403, opened 2017-02-06
[BACKLOG-14145] Inject OSGi requirejs manager configurations script

nantunes:
@pentaho/millenniumfalcon please review.


extensions/src/main/java/org/pentaho/platform/web/http/filters/PentahoWebContextFilter.java
```diff
@@ -158,6 +160,16 @@ public void doFilter( ServletRequest request, ServletResponse response, FilterCh
         printReservedCharsDisplay( out );
         printReservedRegexPattern( out );
 
+        boolean noOsgiRequireConfig = "true".equals( request.getParameter( "noOsgiRequireConfig" ) );
+        if ( !noOsgiRequireConfig ) {
+          final String useFullyQualifiedUrlParameter = httpRequest.getParameter( "fullyQualifiedUrl" );
```
dcleao: Don' see why not let the value determined according to local logic, `shouldUseFullyQualifiedUrl`, pass through.

nantunes: Currently the same logic is being applied on the _other side_, so no practical effect. But allows for different interpretations in the future... Can't imagine what, but no reason to lock it.

PR #3411, opened 2017-02-08
BISERVER-12605 - Testing a connection fails when we have same name for Named Datasource and JNDI datasource

ArtsiomYurhevich:


core/src/main/java/org/pentaho/platform/engine/services/connection/datasource/dbcp/NonPooledOrJndiDatasourceService.java
```diff
@@ -50,10 +48,8 @@ protected DataSource retrieve( String dsName ) throws DBDatasourceServiceExcepti
       if ( databaseConnection != null && !databaseConnection.getAccessType().equals( DatabaseAccessType.JNDI ) ) {
         ds = resolveDatabaseConnection( databaseConnection );
         // Database does not have the datasource, look in jndi now
-      } else if ( databaseConnection == null ) {
-        ds = getJndiDataSource( dsName );
       } else {
-        ds = getJndiDataSource( databaseConnection.getDatabaseName() );
+        ds = getJndiDataSource( dsName );
```
mbatchelor: Not sure I agree with removing the getJndiDataSource( databaseConnection.getDatabaseName() ) part. I think this will cause regressions. Suggest trying the getJndiDataSource( dsName ) and if that fails to return a datasource, then use the other.

ArtsiomYurhevich:
@mbatchelor , Reworked. Review please

PR #3415, opened 2017-02-10
[BISERVER-13523]-Create a .locale bundle entry, when uploading a repo…

EgorZhuk:
…rt from a zip file

In 7.0 the method is refactored, so the logic wasn't moved under an if on line 74.
Backporting the refactoring and the test.


EgorZhuk:
I believe we need an SP case to include it in patch.

PR #3419, opened 2017-02-13
[BISERVER-12916]-Add a filter to handle MS Office pre-flight requests

EgorZhuk:
MS Excel behaves wierd, when opening a file from IE. It sends OPTION requests ( OPTION+HEAD requests in case of XHR ) before opening a file. Pentaho responds with 401 and the file is not opened.
Also, sometimes Excel thinks that the file is broken. In such case it tries to load the content one more time ( another GET ) and repair it. When failed,  it will prompt you a warning message but will open the file anyway. We can't do anything with this particular case without setting a session cookie. And we definitely don't want to do so.

The fix is in returning 200 status for OPTION and HEAD requests for the report content urls.
It adds not so big overhead ( 2 string equals in most cases ).
Unfortunately we won't be able to use it as is in SP cases. We can't change a web.xml. I made the method static and package private to be able to use it in another filter ( most likely WebappRootForwardingFilter ).

I know it looks like a hack, but it tends to match a general developement aproach for MS products :)

### DIFF
--- a/assembly/package-res/biserver/pentaho-solutions/system/pentaho.xml
+++ b/assembly/package-res/biserver/pentaho-solutions/system/pentaho.xml
@@ -154,4 +154,7 @@
 
    </file-upload-defaults>
   <default-theme>sapphire</default-theme>
+
+  <!-- Excel pre-flight request filter pattern -->
+  <pre-flight-pattern>^.*(reporting/api/jobs|api/repos/.*\\.prpt).*$</pre-flight-pattern>
 </pentaho-system>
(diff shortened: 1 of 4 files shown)


tmorgner:
Your comment made my day!

tmorgner:
There are other endpoints in the platform that may want to have the same filter. Analyzer produces Excel files, and given that Excel is so much fun, there are probably other places as well.

Can you make the pattern configurable via a filter-config entry or via a configuration entry in "/pentaho-solutions/system/pentaho.xml" (and then access it via ISystemSetting#getSystemSetting(name, default)?
If nothing is set, use the pattern you already have there.

This way it becomes configurable and support can point OEMs and customers to the settings file instead of having to create a code patch.

EgorZhuk:
@tmorgner now it's configurable. Checked locally + added a unit test.

tmorgner:
.. aaaaaaand merged.

PR #3432, opened 2017-02-21
[BISERVER-13484] 7.0 Manual install on Tomcat 8.0.39- Browse files is not working

Advent51:
Unsafe character - pipe line "|" replaced with "%7C".


Advent51:
@mbatchelor please review

PR #3434, opened 2017-02-21
[Test] Refactoring of RuntimeContextTest to use mocked objects

pentaho-mzur:


pentaho-mzur:
@chiltonj Hey John, please take a look and let me know what you think...

PR #3435, opened 2017-02-21
[BACKLOG-14860] Removed use of activeByDefault for profile activation…

pminutillo:
…s in pom


pminutillo:
@smaring @pentaho/2-1b @emartin-pentaho Please review when you have a chance

PR #2939, opened 2016-05-11
[BISERVER-12815] - Manage Data Sources: User can import an .xmi using "Import Analysis"

andrei-viaryshka:
Updated version of https://github.com/pentaho/pentaho-platform/pull/2929.

PR #3436, opened 2017-02-22
Sp 3100

mchen-len-son:


mchen-len-son:
@mbatchelor @rmansoor can you please take a look?

This is a backport of #2939, a part of [BISERVER-12815](http://jira.pentaho.com/browse/BISERVER-12815) fix, which is needed to fix [SP-3100](http://jira.pentaho.com/browse/SP-3100).

If mondrian schema import caused an exception, then we need to remove that damaged file from repository, in order not to break the entire Data Sources component.

mchen-len-son:
@mbatchelor @rmansoor @pamval can you please take a look?

mchen-len-son:
@mbatchelor @rmansoor @pamval can you please take a look?
thanks.

PR #3438, opened 2017-02-23
[TEST] Improved code coverage for StringUtil and VersionInfo

pentaho-amessier:
Additionally,  fixed test checkstyles where appropriate and converted StringUtilTest to use junit annotations.

### DIFF
--- a/core/src/test/java/org/pentaho/platform/util/StringUtilTest.java
+++ b/core/src/test/java/org/pentaho/platform/util/StringUtilTest.java
@@ -12,29 +12,46 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2013 Pentaho Corporation..  All rights reserved.
+ * Copyright (c) 2002-2017 Pentaho Corporation..  All rights reserved.
  */
 
 package org.pentaho.platform.util;
 
-import junit.framework.Assert;
-import junit.framework.TestCase;
+import org.junit.Assert;
+import org.junit.Test;
 
-public class StringUtilTest extends TestCase {
+/**
+ * Unit tests for the {@link org.pentaho.platform.util.StringUtil} class.
+ */
+public class StringUtilTest {
 
+  @Test
   public void testTokenStringToArray() {
 
-    String[] array = StringUtil.tokenStringToArray( "This is a test to convert this string into an array", " " ); //$NON-NLS-1$ //$NON-NLS-2$
-    StringBuffer buffer = new StringBuffer();
+    String token = " "; //$NON-NLS-2$
+    String tokenString = "This is a test to convert this string into an array"; //$NON-NLS-2$
+    String[] array = StringUtil.tokenStringToArray( tokenString, token );
+    StringBuilder buffer = new StringBuilder();
 
     for ( String element : array ) {
-      buffer.append( element.toString() );
-      buffer.append( " " ); //$NON-NLS-1$
+      buffer.append( element );
+      buffer.append( token );
     }
+    Assert.assertEquals( buffer.toString().trim(), tokenString );
 
-    Assert.assertEquals( buffer.toString().trim(), "This is a test to convert this string into an array" ); //$NON-NLS-1$
+    // when tokenizedString is null, the result should be null
+    Assert.assertNull( StringUtil.tokenStringToArray( null, token ) );
+    Assert.assertNull( StringUtil.tokenStringToArray( null, null ) );
+
+    // when no token is provided, we should get back an array containing the entire input token as a single string
+    tokenString = "some text"; //$NON-NLS-2$
+    array = StringUtil.tokenStringToArray( tokenString, null );
+    Assert.assertNotNull( array );
+    Assert.assertEquals( 1, array.length );
+    Assert.assertEquals( tokenString, array[ 0 ] );
   }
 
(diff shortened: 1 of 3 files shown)


pentaho-amessier:
The remaining test failures were caused by an unrelated commit.

Please review at your convenience.

@neeta-rathi @pdesai16 @pentaho-mracine @chiltonj @niehusa @pentaho-mzur @pentaho-esager @Kyrrui @diogofscmariano

core/src/test/java/org/pentaho/platform/util/StringUtilTest.java
```diff
@@ -75,14 +92,20 @@ public void testDoesPathContainParentPathSegment() {
...
-  public static void main( final String[] args ) {
-    StringUtilTest test = new StringUtilTest();
+  @Test
+  public void testMain() {
     try {
-      test.testTokenStringToArray();
-      test.testDoesPathContainParentPathSegment();
-
-    } finally {
+      Assert.assertTrue( true );
```
pentaho-mzur: What is this for? Feels like some unintended checked in code?

core/src/test/java/org/pentaho/platform/util/StringUtilTest.java
```diff
@@ -75,14 +92,20 @@ public void testDoesPathContainParentPathSegment() {
     }
   }
 
-  public static void main( final String[] args ) {
-    StringUtilTest test = new StringUtilTest();
+  @Test
+  public void testMain() {
     try {
-      test.testTokenStringToArray();
```
pentaho-mzur: It seems that the testDoesPathContainParentPathSegment test was removed. Is this intentional?

core/src/test/java/org/pentaho/platform/util/TestAll.java
```diff
@@ -43,7 +42,6 @@ public static TestSuite testAll() {
     suite.addTestSuite( ParameterHelperTest.class );
     suite.addTestSuite( PasswordServiceTest.class );
     suite.addTestSuite( PentahoSystemExceptionTest.class );
-    suite.addTestSuite( StringUtilTest.class );
```
pentaho-mzur: Why was StringUtilTest removed from the testAll suite?

core/src/test/java/org/pentaho/platform/util/VersionInfoTest.java
```diff
@@ -0,0 +1,263 @@
...
+    Assert.assertEquals( expectedProductID, info.getProductID() );
+    Assert.assertEquals( expectedTitle, info.getTitle() );
+    Assert.assertEquals( expectedVersionMajor, info.getVersionMajor() );
+    Assert.assertEquals( expectedVersionMinor, info.getVersionMinor() );
+    Assert.assertEquals( expectedVersionRelease, info.getVersionRelease() );
+    Assert.assertEquals( expectedVersionMilestone, info.getVersionMilestone() );
+    Assert.assertEquals( expectedVersionBuild, info.getVersionBuild() );
+    Assert.assertEquals( expectedVersionNumber, info.getVersionNumber() );
+
+    final StringBuffer buffer = new StringBuffer();
```
pentaho-mzur: Do we want/need to have a test on to versionInfo.toString()? Not sure what the answer is... it depends if this method must return a string in an standard format. I.e. do we have a contract on the format of toString()? If not, I would recommend removing this test.

core/src/test/java/org/pentaho/platform/util/StringUtilTest.java
pentaho-amessier: Yes - following the preferred convention, unit tests should use annotations and not extend the TestCase class. For that reason, the main method is not needed and the class cannot be added to the test quite within TestAll, because addTestSuite accepts a TestCase argument.

core/src/test/java/org/pentaho/platform/util/TestAll.java
pentaho-amessier: See above

core/src/test/java/org/pentaho/platform/util/StringUtilTest.java
pentaho-amessier: Same comment as below

core/src/test/java/org/pentaho/platform/util/VersionInfoTest.java
pentaho-amessier: I don't have a strong opinion on this, depends on how toString is used. I will defer this to some of the more seasoned pentaho developers and ask for their opinion... 

@pdesai16 @neeta-rathi  @niehusa @chiltonj

core/src/test/java/org/pentaho/platform/util/StringUtilTest.java
pentaho-amessier: Sorry - missed a line above, means to call StringUtil.main(null) to make sure it doesn't throw an exception

mbatchelor: Can't the testMain() method simply be removed? All it does is test that asserting true works - that's not testing anything in StringUtil at all.

pentaho-amessier: This was a mistake on my part - I means to smoke test the StringUtil.main method to ensure that it does not throw any exceptions, this line should be added above the assert:
StringUtil.main( null );

The only reason for this test is to achieve 100% code coverage. I will be updating the PR with this change shortly.

PR #3454, opened 2017-03-03
[BISERVER-13436] Scheduled PRPT Report Parameters are not Referenced …

ivanpogodin:
…when Editing BA Server Schedule


ivanpogodin:
@tmorgner , could you review it, please?
This is done due to https://github.com/pentaho/pentaho-commons-gwt-modules/pull/531#issuecomment-279331629
See also https://github.com/pentaho/pentaho-commons-gwt-modules/pull/548

ivanpogodin:
@pamval , @tmorgner , could you review it, please?
The code coverage decrease is in extensions module.

tmorgner:
I care little about reports from randomized coverage results. If you add a test-case, and as a result of that some totally unrelated code gets reported as decreased in coverage - that is just weird. I dont listen to weird :)

Code itself looks sane.

PR #3457, opened 2017-03-03
[Tests] added class Base64PasswordServiceTest so that the corresponding class has 100% code coverage

pentaho-esager:
Please review my pull request.

### DIFF
--- a/core/src/test/java/org/pentaho/platform/util/Base64PasswordServiceTest.java
+++ b/core/src/test/java/org/pentaho/platform/util/Base64PasswordServiceTest.java
@@ -1,0 +1,84 @@
+/*
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU General Public License, version 2 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/gpl-2.0.html
+ * or from the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ *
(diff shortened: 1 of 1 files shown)


pentaho-esager:
The two test failures wingman is referring to are not from my commit, as I did not touch the code in those classes.  The only class I changed is Base64PasswordServiceTest.

core/src/test/java/org/pentaho/platform/util/Base64PasswordServiceTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Test
+  public void testEncryptValidPassword( ) throws Exception {
+    res = service.encrypt( "asdfghjkldieukancmvaiodfweewjrkwer" );
+    assertEquals( res, "YXNkZmdoamtsZGlldWthbmNtdmFpb2Rmd2Vld2pya3dlcg==" );
+  }
+
+  @Test
+  public void testDecryptNullPassword( ) throws Exception {
+    res = service.decrypt( null );
```
pentaho-amessier: What if you encrypt or decrypt an empty string or a string containing only white spaces? Is it worth testing for that special case?

core/src/test/java/org/pentaho/platform/util/Base64PasswordServiceTest.java
```diff
@@ -0,0 +1,58 @@
...
+  @Test
+  public void testEncryptNullPassword( ) throws Exception {
+    res = service.encrypt( null );
+    assertNull( res );
+  }
+
+  @Test
+  public void testEncryptValidPassword( ) throws Exception {
+    res = service.encrypt( "asdfghjkldieukancmvaiodfweewjrkwer" );
+    assertEquals( res, "YXNkZmdoamtsZGlldWthbmNtdmFpb2Rmd2Vld2pya3dlcg==" );
```
pentaho-amessier: It might be worth decrypting here and verifying that we get back the original string.

core/src/test/java/org/pentaho/platform/util/Base64PasswordServiceTest.java
pentaho-esager: Updated and repushed

pentaho-esager:
Again, some checks done by wingman refer to classes I did not modify, the only class I changed was Base64PasswordServiceTest, which shows an increase in coverage and passing tests.

PR #3460, opened 2017-03-06
[BACKLOG-11933][PDB-1938] Encode URI for the tab name when user input contains special characters

niehusa:
@pentaho/rogueone 
Please review. 
The fix is to do a URI encoding on file name user entered. So that special characters won't break succeeding JSON response.


niehusa:
@pentaho/rogueone 
Folks, Please review this fallback solution. If everything is alright, please merge the PR so that it can be in today's nightly build.

chiltonj:
If we change the double quote to two single quotes then we are effectively disallowing double quotes. If we are disallowing double quotes then it is better to have customer decide on what to replace them with -- not make that decision for them without their input.

niehusa:
Totally agreed with @chiltonj . Maybe we should remove double quote from the supported special character list, simply disallowing it.

PR #3462, opened 2017-03-09
[BACKLOG-14605] Shutdown system listeners in reversed order

niehusa:
@pentaho/rogueone 
Please review. Changed Iterator to be typed ListIterator so that list elements can be traversed backward.


pedrofvteixeira:
👍

PR #3464, opened 2017-03-10
[BACKLOG-14990] Fix OSGi 'and' filter invalid syntax

nantunes:
The 'and' filter syntax only has a `&` at the begining, not between operands. Was causing `InvalidSyntaxException`.

Also changed the util function so it returns a simple filter if only one property exists.

@pentaho/millenniumfalcon and @pentaho-nbaker please review.


nantunes:
The broken test has nothing to do with this. They all pass locally.

PR #3467, opened 2017-03-13
[SP-3282]-[BACKLOG-14587] Menu Home has all options align center in theme Sapphire

PashkovichAndrei:


PashkovichAndrei:
@mchen-len-son, could you please do review?

PR #3481, opened 2017-03-21
[BACKLOG-14605] added a simple test case to ensure the shutdown order…

pentaho-esager:
… occurs in reverse

### DIFF
--- a/core/src/test/java/org/pentaho/platform/engine/core/system/TestShutdownOrderListener.java
+++ b/core/src/test/java/org/pentaho/platform/engine/core/system/TestShutdownOrderListener.java
@@ -1,0 +1,42 @@
+package org.pentaho.platform.engine.core.system;
+
+import org.pentaho.platform.api.engine.IPentahoSession;
+import org.pentaho.platform.api.engine.IPentahoSystemListener;
+
+/*
+ * This program is free software; you can redistribute it and/or modify it under the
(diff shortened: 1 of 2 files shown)


pentaho-esager:
The wingman comments are talking about code I didn't modify.

pentaho-mracine:
@pentaho-esager Wingman does that sometimes, I believe it has something to do with other wingman jobs running concurrently. As long as you don't have any "newly broken tests" I believe you should be fine.

core/src/test/java/org/pentaho/platform/engine/core/system/TestShutdownOrderListener.java
```diff
@@ -0,0 +1,42 @@
+package org.pentaho.platform.engine.core.system;
```
pentaho-mracine: Maybe put this class inside the PentahoSystemShutdownOrderTest.java file?

pentaho-esager: I did it this way because I saw several other tests of different types of system listeners set up as separate classes, and I wanted to be consistent with other tests.  I am happy to change it if you think it's overall better to have it in one class though!

pentaho-mracine: You're right. If the existing test code does it this way then it's probably best to just keep as is then.

PR #3484, opened 2017-03-23
[BISERVER-13473] JMS Connection for destination 'snmpQueue' error after switching to LDAP authentication on Pentaho 7

Advent51:
Fixed regression which was caused by switching to the spring-ldap-2.1.0

### DIFF
--- a/assemblies/pentaho-solutions/src/main/resources/pentaho-solutions/system/applicationContext-spring-security-ldap.xml
+++ b/assemblies/pentaho-solutions/src/main/resources/pentaho-solutions/system/applicationContext-spring-security-ldap.xml
@@ -12,7 +12,7 @@
        						http://www.pentaho.com/schema/pentaho-system http://www.pentaho.com/schema/pentaho-system.xsd
        						http://www.springframework.org/schema/util http://www.springframework.org/schema/util/spring-util-4.3.xsd" default-lazy-init="true">
 
-  <bean class="org.pentaho.platform.plugin.services.security.userrole.ldap.DefaultLdapAuthenticationProvider">
+  <bean id="ldapAuthenticationProvider" class="org.pentaho.platform.plugin.services.security.userrole.ldap.DefaultLdapAuthenticationProvider">
     <constructor-arg>
       <ref bean="authenticator" />
     </constructor-arg>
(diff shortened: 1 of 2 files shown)


Advent51:
@mbatchelor @mchen-len-son please review

Advent51:
@mbatchelor @mchen-len-son please review

pamval:
@pedrofvteixeira please review

pedrofvteixeira:
@duarteteixeira @mbatchelor @dkincade should this fix get merged: is there a need to capture this somewhere in the documentation ? Thinking of the "Upgrade path from 6.x to 7.1" ( https://help.pentaho.com/Documentation/7.0/0T0/Upgrade_Pentaho_Servers_and_Design_Tools )

pedrofvteixeira:
@Advent51 reading through this jira's comments, quote,

> This is actually happens if you switch to any other security besides Jackrabbit. Not only LDAP

Which means that other supported auth. methods such as JDBC and/or SAML ( dunno if SAML is fully supported by now, TBH ) are also suffering from the same issue.

This fix would only address LDAP.

In the jira, I can also read that, quote,

> 2. Open the pentaho-solutions/system/karaf/etc/org.apache.karaf.features.cfg
> 3. Find and remove the following: "pentaho-monitoring-to-snmp,pentaho-monitoring-jms-broker

would actually address the issue.

Don't know if we need to have the monitoring capabilities enabled by default. 
If we don't, then this would no longer be an issue.

And on some documentation on "How to enable Pentaho SNMP Monitoring", we could add a Troubleshooting section for anyone experiencing this 'snmpQueue' error after switching to LDAP, where we would then mention the manual step of replacing _"DefaultLdapAuthenticationProvider"_ with "ClassloaderSwitchLdapAuthenticationProvider"

The proposed solution above would:

1. Prevent us for doing a change on the default LDAPProvider we're setting nowadays
2. Isolate such a change for anyone that wants to explicitly enable SNMP traps with LDAP Auth.
3. Have a class similar to "ClassloaderSwitchLdapAuthenticationProvider" for each of the other authentications out these  that also suffer of this issue ( such as JDBC / SAML )


My $0.02c, @CodeOnCoffee @mbatchelor @pamval @dkincade @duarteteixeira

extensions/src/main/java/org/pentaho/platform/plugin/services/security/userrole/ldap/ClassloaderSwitchLdapAuthenticationProvider.java
```diff
@@ -0,0 +1,75 @@
...
+   * the next <code>AuthenticationProvider</code> that supports the presented
+   * <code>Authentication</code> class will be tried.
+   *
+   * @throws AuthenticationException if authentication fails.
+   */
+
+  @Override
+  public Authentication authenticate( Authentication authentication ) throws AuthenticationException {
+    ClassLoader currentThreadContextClassLoader = Thread.currentThread().getContextClassLoader();
+    Thread.currentThread().setContextClassLoader( this.getClass().getClassLoader() );
```
mbatchelor: I need this to be in a try/finally block.

pentaho-nbaker:
I would rather address this by wrapping the active AuthenticationProvider in a Spring Proxy Bean which sets the TCCL to that of the target bean for all method calls. This way all AuthenticationProviders will get the fix and it's centralized in one place:
https://github.com/pentaho/pentaho-ee/blob/master/data-integration/assemblies/client/platform/package-res/biserver/pentaho-solutions/system/applicationContext-spring-security.xml#L40

See: org.springframework.aop.framework.ProxyFactoryBean

We can also build this TCCL fix behavior into the <pen:bean fix-tccl="true">. I'd probably go this route if this were on a major release, but that's changing a component used in many more places.
https://github.com/pentaho/pentaho-platform/blob/master/core/src/main/java/org/pentaho/platform/engine/core/system/objfac/spring/BeanBuilder.java#L160

PR #3486, opened 2017-03-23
[SP-3394] Backport of PPP-3420 - Mondrian cache is not cleared when c…

mchen-len-son:
…learing it via another Admin user.


mchen-len-son:
@pamval @mbatchelor @duarteteixeira please take a look.
thanks.

PR #3491, opened 2017-03-27
[BACKLOG-15600] - Fix licenses for user-console, repository, extensio…

rfellows:
…ns, and scheduler sub-modules.


rfellows:
@arunimagupta @mbatchelor please review

PR #3492, opened 2017-03-29
[BISERVER-12808]-There is no error message if you run import-export.bat file with invalid command

PashkovichAndrei:
* Created validation for charset parameter

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/plugin/services/importexport/CmdParser.java
+++ b/extensions/src/main/java/org/pentaho/platform/plugin/services/importexport/CmdParser.java
@@ -1,0 +1,124 @@
+/*
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU General Public License, version 2 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/gpl-2.0.html
+ * or from the Free Software Foundation, Inc.,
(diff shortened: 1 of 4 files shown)


PashkovichAndrei:
@mchen-len-son, could you please do review?

PashkovichAndrei:
Solution was reworked. 
Also refactored class according TODO.

@mbatchelor, please review a fix

mchen-len-son:
@mbatchelor can please take a look?

pamval:
@PashkovichAndrei does this address Marc's comments in JIRA  about unknown options?

pamval:
yeah, a simple test showing that behavior would be good.

PashkovichAndrei:
@pamval, new tests was created.

But there is one possible danger place.
Let's imagine that we have a list of options: "abc" (-a, --abc); "bcd" (-b, --bcd); "cde" (-c, --cde). Then if we pass "-cde=123" to the command line, that lib (which is used at command line processor) will parse that as short option -c with value "de=123". That's how that lib works. 

My fix based on value validation after parsing option. So does that fix is acceptable for that issue or should I rework parsing of options and throw exception during the parsing?

pamval:
that looks like a change in behavior that we should avoid. If this breaks existing behavior, I think you should rework your change

PashkovichAndrei:
ok, I'll rework my solution

PashkovichAndrei:
@pamval, @mbatchelor could you please do review of reworked solution?

PashkovichAndrei:
@pamval, @mbatchelor could you please do review of reworked solution?

PR #3494, opened 2017-03-30
[BACKLOG-15653] In KarafBoot, while looking for etc folder based on c…

pdesai16:
…lientType, if the

clientType is null, the folder is getting defaulted to etc-default. Since this
folder does not exist in PentahoServer, added a check to ensure the folder
exist, before using it to set the path in system property.


pdesai16:
@CodeOnCoffee 
@pentaho/rogueone 

Please review

PR #3497, opened 2017-04-03
[PPP-3686] Duplicate logging for session start

Advent51:
- catching the InteractiveAuthenticationSuccessEvent was removed


Advent51:

PR #3506, opened 2017-04-10
[BISERVER-12677]-Replace periodic checks with a "just in time" check

EgorZhuk:
A simple and effective improvement.
Why to check for the expired session periodically if the expiration time can only be increased?
Let's shedule a check to the current expiration time. If there was no activity - OK, the session is dead.
If there was - create a new timer, and so on.


EgorZhuk:
@pamval @mbatchelor @tmorgner could you please take a look?

PR #3509, opened 2017-04-11
[BISERVER-12883]  Inconsistent behavior with generated/uploaded filen…

alina-ipatina:
…ames

- do not remove extension from files, except files with known extensions

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/plugin/services/importer/RepositoryFileImportFileHandler.java
+++ b/extensions/src/main/java/org/pentaho/platform/plugin/services/importer/RepositoryFileImportFileHandler.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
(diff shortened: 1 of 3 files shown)


alina-ipatina:
@mchen-len-son please review.

alina-ipatina:
@mchen-len-son please review.

mchen-len-son:
Thank you.

mchen-len-son:
Seems like some tests are non-deterministic - latest wingman's report shows different coverage, but the only change was about fields reordering.

extensions/src/main/java/org/pentaho/platform/plugin/services/importer/RepositoryFileImportFileHandler.java
```diff
@@ -49,10 +50,13 @@
  */
 public class RepositoryFileImportFileHandler implements IPlatformImportHandler {
 
+  private static final List<String> KNOWN_EXTENSIONS = Arrays.asList( "prpt", "prpti", "xdash",
+          "xaction", "xanalyzer", "kjb", "ktr", "xcdf", "xmi", "cda", "xreportspec" );
```
pamval: This should get read from the system configurations, not hardcoded here. Right? I mean, the accepted extensions are defined in importExport.xml or whatever that is. So we need to query the system to know which file extensions are known.

alina-ipatina:
@mchen-len-son @pamval @mbatchelor please review if you have a chance.
Thank you.

PR #3520, opened 2017-04-14
PRD-5865 - Pentaho Platform has inconsistent methods to connect to Mo…

YuryBY:
…ndrian, creating multiple caches, of which only one can be cleared.

@pamval, @tmorgner, @lucboudreau, could you please take a look?

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/plugin/action/olap/impl/OlapServiceImpl.java
+++ b/extensions/src/main/java/org/pentaho/platform/plugin/action/olap/impl/OlapServiceImpl.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright 2016 Pentaho Corporation.  All rights reserved.
+ * Copyright (c) 2002-2017 Pentaho Corporation.  All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


YuryBY:
@pamval, @tmorgner, @lucboudreau, the PR was updated tests pass locally.

YuryBY:
@pamval, @tmorgner, @lucboudreau, any updates?

YuryBY:
@pamval, @tmorgner, @lucboudreau, any updates?

extensions/src/main/java/org/pentaho/platform/plugin/action/olap/impl/OlapServiceImpl.java
```diff
@@ -504,6 +504,12 @@ public void flushAll( IPentahoSession session ) {
       resetCache( session );
 
       flushHostedAndRemote( session );
+
+      MondrianServer mondrianServer =
```
tmorgner: You will need to clear both server instances. The bug states that there are inconsistent methods for actually acquiring a mondrian instance used by the various tools. Your patch here does not address that, it only implements the workaround outlined in the case.

As there are still multiple instances alive, you have to call the flushSchemaCache() line on both the active mondrian server (this.server) and the one with Id "null". You cant just clear one, but not the other.

YuryBY:
@tmorgner, the PR is updated. Thanks.

PR #3523, opened 2017-04-17
[CLEANUP] - Removing unnecessary files. We are no longer using ant/ivy

rfellows:


rfellows:
@dkincade please review

PR #3531, opened 2017-04-18
[BACKLOG-15987] for this point onward, 'Inherits folder permissions' …

pedrofvteixeira:
…will always be disabled for any folder within /home ( i.e. home folders )


pedrofvteixeira:
@pentaho/rogueone , @dkincade , @kcruzada , @mbatchelor please review. 

This change is made as per agreed upon: to solve BACKLOG-15986 and BACKLOG-15987, the rule that needs to be in place is:

- any folder within /home ( i.e. user home folders ) should *not* have the 'Inherits folder permission' checkbox enabled
- files in user home folders, as well as anything in /public should be the ones where 'Inherits folder permission' checkbox comes into play

pedrofvteixeira:
Rebased with wingman's checkstyle violations addressed

PR #3534, opened 2017-04-20
[BISERVER-12883] Inconsistent behavior with generated/uploaded filenames

alina-ipatina:
- add null check


alina-ipatina:
@mchen-len-son @mbatchelor @pamval please review.

PR #3539, opened 2017-04-21
[PDI-16124]- PDI Schedule API - Parameters defined in the xml input is not being passed to the job

ArtsiomYurhevich:


ArtsiomYurhevich:
@mbatchelor , @pedrofvteixeira Could you please take a look?

PR #3541, opened 2017-04-25
[BACKLOG-15987] Revert: disabling 'inherits folder permission' checkb…

pedrofvteixeira:
…ox for root and home folders

- As per @kcruzada , @bkemper-pentaho: reverting this change, and focusing on an alternate solution


pedrofvteixeira:
@pentaho/rogueone , @kcruzada , @bkemper-pentaho , @pamval , @dkincade

PR #3545, opened 2017-04-26
[BISERVER-13579]- REST API /userrolelist/getRolesForUser and /userrolelist/getUsersInRole don't return JSON

ArtsiomYurhevich:
-XML hard-coded response is removed.  Now the answer type (XML or JSON) depends on Accept header.
-The previous type of xml aswer (as in pentaho documentation) is  retained

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/RolesWrapper.java
+++ b/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/RolesWrapper.java
@@ -1,0 +1,48 @@
+/*!
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License, version 2.1 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
+ * or from the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * Copyright (c) 2017-2017 Pentaho Corporation..  All rights reserved.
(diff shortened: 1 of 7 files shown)


ArtsiomYurhevich:
@mbatchelor , @pedrofvteixeira Could you please take a look?

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/RolesWrapper.java
```diff
@@ -0,0 +1,48 @@
+/*!
```
mbatchelor: Why not use existing RoleListWrapper.java?

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/UsersWrapper.java
```diff
@@ -0,0 +1,48 @@
+/*!
```
mbatchelor: Why not use existing UserListWrapper.java ?

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/RolesWrapper.java
ArtsiomYurhevich: @mbatchelor I did it because we cant change the format of existing xml response.  Please look at https://help.pentaho.com/Documentation/7.0/0R0/070/010/0D0/0N0#path__userrolelist_getRolesForUser.html. The method should return`<roles>
<role>Role1</role>
<role>Role2</role>
</roles>`. But if we use RoleListWrapper it will return `<roleList>
<roles>Role1</roles>
<roles>Role2</roles>
</roleList>` . You may look at example https://help.pentaho.com/Documentation/7.0/0R0/070/010/0D0/0N0#path__userrolelist_extraRoles.html . These two rest endpoints ( getRolesForUser and getUsersInRole) have a different format of xml from  other endpoints in this class.

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/UsersWrapper.java
ArtsiomYurhevich: The same as above. The current format -
 https://help.pentaho.com/Documentation/7.0/0R0/070/010/0D0/0N0#path__userrolelist_getUsersInRole.html . The example of format we will get if we use UserListWrapper - https://help.pentaho.com/Documentation/7.0/0R0/070/010/0D0/0N0#path__userrolelist_users.html

PR #3546, opened 2017-04-26
[PPP-3715] Need to encode inputs to Javascript Rule

Advent51:
- Encoding was added
- Test was added

### DIFF
--- a/core/src/main/java/org/pentaho/platform/engine/services/MessageFormatter.java
+++ b/core/src/main/java/org/pentaho/platform/engine/services/MessageFormatter.java
@@ -13,7 +13,7 @@
  * See the GNU General Public License for more details.
  *
  *
- * Copyright 2006 - 2013 Pentaho Corporation.  All rights reserved.
+ * Copyright 2006 - 2017 Pentaho Corporation.  All rights reserved.
  */
 
 package org.pentaho.platform.engine.services;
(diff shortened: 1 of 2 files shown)


Advent51:
@mbatchelor @mchen-len-son please review.
Marc, I couldn't find any way to verify this issue except writing this simple unit test. Do you have any idea how to do that?
I have tried to verify it through writing an `xaction`, but I didn't reach a success.

Advent51:
@mbatchelor please review

dkincade:
Let's get @mbatchelor to review this since it involves encoding...

extensions/src/main/java/org/pentaho/platform/plugin/action/javascript/JavascriptRule.java
```diff
@@ -266,6 +267,7 @@ protected Object executeScript( final ScriptableObject scriptable, final Scripta
         results.setResultSet( (IPentahoResultSet) inputValue );
         wrapper = Context.javaToJS( inputValue, results );
       } else {
+        inputValue = Encode.forHtml( Context.toString( inputValue ) );
```
mbatchelor: I think encoding for html could cause issues here. Suggest killing the encode here as I couldn't demonstrate any ability to subvert the Javascript system based on how we bind the variables. Suggest doing the encode in the success message and calling it a day.

Advent51:
@mbatchelor please review

PR #3548, opened 2017-04-28
[PPP-3700]-Schedule's "Start time" is changing after using "Execute now" function

PashkovichAndrei:
* Removed contradictory code
* Fixed integration test


PashkovichAndrei:
@mbatchelor, @mchen-len-son could you please do a review?

PR #3549, opened 2017-05-01
[BACKLOG-15986] Reset the state of add and remove button to the value

pdesai16:
before the confirmation dialog, if the user cancel in the "Inherit folder
permission" dialog.


pdesai16:
@pentaho/rogueone  Please review

PR #3553, opened 2017-05-02
[BACKLOG-15986] Added the call to setInheritsAcls() in cancel handler…

pdesai16:
… of the

confirmation dialog. This was to avoid any side effects as
refreshPermissions are being called from lots of places.


pdesai16:
@pentaho/rogueone  please review

PR #3574, opened 2017-05-08
[BACKLOG-11933][REVERT] Encode URI for the tab name when user input c…

pedrofvteixeira:
…ontains special characters

- [REVERT] https://github.com/pentaho/pentaho-platform/pull/3460/commits/20117fcba1a625243a39644fafa590514fd4478d
- Focus will be on actually allowing a double-quote on the first save


pedrofvteixeira:
@pamval , @dkincade , @kcruzada , @mdamour1976 please review.

CC'ing @pentaho/rogueone

PR #3579, opened 2017-05-10
[BACKLOG-11933] adding fix that escapes the id of the tab for safe co…

befc:
…nversion to JSON, when report is saved with a name containing double quote (") character(s)

This happens because on save, not only the tab name is changed but also the tab id. In case the name contains double quotes tab it causes problems when "fireCloseTab( frameId )" is called due to conversion to JSON. Ex. for tab name a"b:
* #a"b1494409287116 need to be escaped to #a\\"b1494409287116.

### DIFF
--- a/user-console/src/main/java/org/pentaho/mantle/client/ui/tabs/MantleTabPanel.java
+++ b/user-console/src/main/java/org/pentaho/mantle/client/ui/tabs/MantleTabPanel.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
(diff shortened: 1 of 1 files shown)


pedrofvteixeira:
👍

pedrofvteixeira:
@pentaho/rogueone

pentaho-jlien:
@befc @pedrofvteixeira Just curious...Should we be including in the java file (in CE) a comment with mention of a specific Backlog Item? Won't this comment be visible to the Open Source community?

pedrofvteixeira:
@pentaho-jlien good catch. I was the one that recommended us leaving a _"why we're doing this"_ comment. But yep, you're right, community can't see BACKLOG-type cases. 

@befc please issue a quick follow-up [CLEANUP] PR where we strip that _"[BACKLOG-11933]"_ bit from the comment ( keeping the comment, ofc ). Thanks @pentaho-jlien , @befc.

befc:
@pentaho-jlien, @pedrofvteixeira, I submitted the cleanup PR: https://github.com/pentaho/pentaho-platform/pull/3584

PR #3583, opened 2017-05-11
[BACKLOG-11933] adding fix that escapes the id of the tab for safe co…

befc:
…nversion to JSON


befc:
@dkincade, @kcruzada, @pamval: https://github.com/pentaho/pentaho-platform/pull/3579 already dev validated in 8.0-QAT-100
CC: @pentaho/rogueone

PR #3589, opened 2017-05-12
[BISERVER-13385] IE specific: Shape of hidden buttons "Edit Content",…

mchen-len-son:
… "Save", "Save As" is visible

- changed disabled state of the buttons to invisible, per UX recommendation


mchen-len-son:
@pamval @pedrofvteixeira can you please take a look?
thanks.

PR #3594, opened 2017-05-16
[BACKLOG-16489] Pentaho Server CE - Folder 'plugin-samples' is hidden

pedrofvteixeira:
- CE plugin-samples's exportManifest.xml has the 'isHidden' flag set to 'true'
  - https://github.com/pentaho/pentaho-platform/blob/7.1.0.0/assemblies/pentaho-plugin-samples/src/main/resources/plugin-samples/exportManifest.xml#L26 

- EE also holds a plugin-samples's exportManifest.xml, that also has the 'isHidden' flag set to 'true'
  - https://github.com/pentaho/pentaho-platform-ee/blob/7.1.0.0/assemblies/pentaho-solutions-ee/src/main/resources/pentaho-solutions/plugin-samples/exportManifest.xml#L26

We switch CE plugin-samples's 'isHidden' flag to 'false', leaving EE's the way it is.

### DIFF
--- a/assemblies/pentaho-plugin-samples/src/main/resources/plugin-samples/exportManifest.xml
+++ b/assemblies/pentaho-plugin-samples/src/main/resources/plugin-samples/exportManifest.xml
@@ -23,7 +23,7 @@
 	
     <ExportManifestEntity path="public/plugin-samples">
         <ExportManifestProperty>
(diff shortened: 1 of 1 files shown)


pedrofvteixeira:
@pamval please review

CC'ing @pentaho/rogueone

PR #3608, opened 2017-05-22
[PPP-3734] Use of vulnerable component webservices-rt-2.1- CVE-2013-5…

stanislau-strelchanka:
[PPP-3734] Use of vulnerable component webservices-rt-2.1- CVE-2013-5816 and CVE-2013-2172

 - remove jaxb-impl from class path as 'com.sun.xml.bind' package already implemented in webservices-rt

### DIFF
--- a/assemblies/pentaho-war/pom.xml
+++ b/assemblies/pentaho-war/pom.xml
@@ -940,11 +940,6 @@
       <version>${jackson.version}</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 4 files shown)


mchen-len-son:
@mbatchelor @rmansoor @pamval can you please take a look?
this is a cure for regression in the last QAT, when you cannot log in to PUC.

stanislau-strelchanka:
@mbatchelor @rmansoor @pamval can you please take a look?

PR #3612, opened 2017-05-24
[SP-3534]-[PPP-3700]-Schedule's "Start time" is changing after using …

PashkovichAndrei:
…"Execute now"

function
* Removed contradictory code
* Fixed integration test


mchen-len-son:
The failing test is non-deterministic. the failure is not related to the changes

PR #3616, opened 2017-05-24
[BACKLOG-16728] - Using relative context path in WebContext

rmansoor:
- Return false to shouldUseFullyQualifiedUrl to make sure CONTEXT_PATH is relative

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/web/http/filters/PentahoWebContextFilter.java
+++ b/extensions/src/main/java/org/pentaho/platform/web/http/filters/PentahoWebContextFilter.java
@@ -197,15 +197,16 @@ public class PentahoWebContextFilter implements Filter {
     }
   }
 
+
   private boolean shouldUseFullyQualifiedUrl( HttpServletRequest httpRequest ) {
     final String useFullyQualifiedUrlParameter = httpRequest.getParameter( "fullyQualifiedUrl" );
     if ( useFullyQualifiedUrlParameter != null ) {
       return "true".equals( useFullyQualifiedUrlParameter );
     } else {
-      final String referer = httpRequest.getHeader( "referer" );
-      final String serverAddress = httpRequest.getScheme() + "://" + httpRequest.getServerName() + ":" + httpRequest.getServerPort();
-
-      return referer != null && !referer.startsWith( serverAddress );
+      // Returning false for now. The smart way of determining whether we should use the fully
+      // qualified url did not work behind the proxy server and this case
+      // http://jira.pentaho.com/browse/BACKLOG-16728 was created.
+      return false;
     }
   }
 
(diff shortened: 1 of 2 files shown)


rmansoor:
@nantunes , @pentaho-nbaker please review

extensions/src/main/java/org/pentaho/platform/web/http/filters/PentahoWebContextFilter.java
```diff
@@ -197,16 +197,10 @@ public void doFilter( ServletRequest request, ServletResponse response, FilterCh
...
-    if ( useFullyQualifiedUrlParameter != null ) {
-      return "true".equals( useFullyQualifiedUrlParameter );
-    } else {
-      final String referer = httpRequest.getHeader( "referer" );
-      final String serverAddress = httpRequest.getScheme() + "://" + httpRequest.getServerName() + ":" + httpRequest.getServerPort();
 
-      return referer != null && !referer.startsWith( serverAddress );
-    }
+  private boolean shouldUseFullyQualifiedUrl( HttpServletRequest httpRequest ) {
+    // Will return false for now.
```
nantunes: If not removing it completely, I would keep the code for the explicit use of the `fullyQualifiedUrl` parameter and just always return `false` inside the `else` clause (so it doesn't try to be _"smart"_).

Also include the BACKLOG-16728 reference in the comment, so in the future is it easier to find out why it is returning false for now.

extensions/src/test/java/org/pentaho/platform/web/http/filters/PentahoWebContextFilterTest.java
```diff
@@ -162,54 +162,6 @@ public void testDoGetWithNoOsgiRequireConfigDefault() throws ServletException, I
     assertTrue( this.requirejsManagerInitIsCalled( response, null ) );
   }
 
-  @Test
-  public void testDoGetWithFullyQualifiedUrlTrue() throws ServletException, IOException {
```
nantunes: Recover the first two tests, that deal with the `fullyQualifiedUrl` parameter.

You can also modify the other two to tests to verify it uses the relative path by default.

dcleao:
The `fullyQualifiedUrl` parameter is an important feature that enables **embeddability** of whatever we do in the web/JavaScript side of the Pentaho platform. This includes the Visualization API. CDF can also start using this new capability to be able to drop the custom `cdf-embed.js`/`cde-embed.js` endpoints it had to develop in the past.

nantunes:
@dcleao right. It might, probably, fail the same way it is failing on the proxy scenario (although it did work in our sandboxed embedded scenario).
We need a story for figuring out what plugins aren't handling the full URL correctly.

But keeping the parameter is harmless, so only the change that modified existing default behaviour must be reverted.

dcleao:
@nantunes Good point.

rmansoor:
@nantunes . I have updated the PR with the changes you have suggested. I have  tested with the changes and all is working well. Please review the updated PR now

PR #3623, opened 2017-05-25
[BISERVER-13633] Schema cache refresh takes too long

mkambol:
Removing the logic which iterates each hosted catalog,
creating a connection and flushing individually.  That had
the effect of loading each schema, flushing it out, and
then loading again.  Big perf hit, and unnecessary, since
we were also invoking CacheControl.flushSchemaCache

http://jira.pentaho.com/browse/BISERVER-13633
http://jira.pentaho.com/browse/BISERVER-13631


mkambol:
@lucboudreau 
Coverage drop is due to removed code.

PR #3626, opened 2017-05-26
[BACKLOG-16711,BACKLOG-15332,BISERVER-13266] Changed filtering of system folders from FileService …

befc:
[BACKLOG-16711] Changed filtering of system folders from FileService & DefaultUnifiedRepositoryWebService to JcrRepositoryUtils. Added includeSysDirs parameter to the getTree rest call

[BACKLOG-15332,BISERVER-13266] Protect RepositoryFileAdapter.toFileDto() against NullPointerException while reading file attributes. When Exception is found, returns null

- **RepositoryRequest:** Added boolean parameter for the includeSystemFolder option. The default value is true because PDI uses DefaultUnifiedRepositoryWebService.getTreeFromRequest() and expects to receive system folders unless the user is not admin;
- **FileResource:** Expose the _includeSysDirs_ parameter in the REST interface; **Note:** This didn't exist before, system folders were never included in the response tree object
- **FileService:** Add call for doGetTree including the _includeSystemFolders_ flag; Remove the for cycle that was filtering system folders at this point;
- **DefaultUnifiedRepository:** Checkstyle cleanup;
- **JcrRepositoryFileUtils:** Before applying Collections.sort() to the list of children, remove all null values from the tree, this was sometimes causing 500 errors on the tree call as well; Include _includeSystemFolders_ and _rootPath_ in the getTreeByNode() and checkNodeForTree() recursive calls to apply the system folders filtering within; While iterating over the tree nodes apply a filter that  skips system folders at root level if includeSystemFolders is false; Add private method isSystemFolder() to verify if a file node is a system folder
- **RepositoryFileProxy:** Replaced all e.printStackTrace(); calls with getLogger().error( "RepositoryException was found: ", e ); **Note:** This is not required for this particular fix, but was identified as a best practice since was hiding some of the real errors behind the 500 errors being found;
- **DefaultUnifiedRepositoryWebService:** Removed the filtering of system folders when the call is made by non admin users; Included the necessary logic to pass down this filtering to _JcrRepositoryFileUtils_;
- **RepositoryFileAdapter:** Added a try catch block in the toFileDto() method to return null in case it returns NPE while getting the file parameters. Since this means that the file no longer exists.
- **RepositoryFileTreeAdapter:** Add null value handling for when null is returned from RepositoryFileAdapter.toFileDto()

### DIFF
--- a/api/src/main/java/org/pentaho/platform/api/repository2/unified/RepositoryRequest.java
+++ b/api/src/main/java/org/pentaho/platform/api/repository2/unified/RepositoryRequest.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2002-2015 Pentaho Corporation..  All rights reserved.
+ * Copyright (c) 2002-2017 Pentaho Corporation..  All rights reserved.
  */
 
 package org.pentaho.platform.api.repository2.unified;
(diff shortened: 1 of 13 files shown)


pedrofvteixeira:
@dkincade , @mbatchelor , @CodeOnCoffee , @rmansoor , @pentaho/rogueone please review

PR #3628, opened 2017-05-29
[SP-3396] Backport of BISERVER-12949 - BA Server throws Invalid Bundl…

mchen-len-son:
…eContext error during shutdown


mchen-len-son:
@pamval @pedrofvteixeira can you please take a look?
thanks.

PR #3633, opened 2017-06-01
[PDI-14912]-Within PDI, visualize table output on geomap does not work

stepanovdg:


mbatchelor:
PR closed due to PDI-14912 being closed as Won't Fix.

PR #3644, opened 2017-06-08
[PDI-14912]-Within PDI, visualize table output on geomap does not work

stepanovdg:


kolinus:
@rmansoor  could you please take a look

PR #3646, opened 2017-06-08
[BISERVER-13621] All the command line parameters for import-export.sh/.bat are showing the localized parameters names.

Advent51:
- parameters were pulled out of the message bundle;
- only English parameters are allowed now.


Advent51:
@mchen-len-son @mbatchelor @pamval please review

kolinus:
@mbatchelor @pamval could you please review

PR #3648, opened 2017-06-09
[COMP-6163] in relation to [BACKLOG-16372] Upgrading version of httpcore

pentaho-mzur:
The currently used httpcore-4.4.3 was never officially officially approved but httpcore-4.4.4 is already approved and used in other parts of the product. The jar is pulled in as a run time dependency from exenstions project and assembled by pentaho-war.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -199,7 +199,7 @@
     <rome.version>1.0</rome.version>
     <javatar.version>2.5</javatar.version>
     <elasticsearch.version>0.16.3</elasticsearch.version>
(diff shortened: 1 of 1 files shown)


pentaho-mzur:
@diogofscmariano please review - I am not sure if we want to upgrade or not yet, but putting this up for review in case we decide to. 

Also adding @lucboudreau and please add someone from build team as needed.

pentaho-mzur:
Closing this pull request since it was decided to revert the use of this library.

PR #3652, opened 2017-06-13
[BISERVER-13603] - IE - PDF pane remains active when glass pane was a…

YuryBY:
…ctivated, which blocks user from accepting/declining action in appeared dialog box.

@pamval, @mbatchelor, could you please take a look?


kolinus:
@rmansoor could you please take a look

YuryBY:
@rmansoor, @pamval, @mbatchelor could you please take a look?

PR #3657, opened 2017-06-19
[BISERVER-10319] adapted for using from other projects

stepanovdg:


stepanovdg:
@mchen-len-son @pamval, @mbatchelor, Could you please take a look?

PR #3658, opened 2017-06-19
[BISERVER-12456] "New Data Source" button is available on Opened pers…

stanislau-strelchanka:
[BISERVER-12456] "New Data Source" button is available on Opened perspective for user without appropriate permissions


stanislau-strelchanka:
@mdamour1976 @e-cuellar @pentaho-nbaker @mbatchelor @mchen-len-son please review

PR #3659, opened 2017-06-19
[TEST] add tests for org.pentaho.platform.engine.services.SoapHelper

pentaho-whartman:
This PR adds unit tests for SoapHelper.java

Some unusual behavior in SoapHelper was found:
- `SoapHelper.createSoapResponseDocument(String, String)` adds `"\"http://pentaho.org\""` as an activity response. All other versions of the method add `"http://pentaho.org"`.
- Calls to `SoapHelper.createSoapFaultElement()` result in a `org.dom4j.IllegalAddException`, no matter what arguments are passed. 

Given that this behavior has been present for four years, these potential issues were ignored in test cases. They may be worth addressing in the future.


pentaho-whartman:
Updated the names. Let me know if there's anything else.

PR #3661, opened 2017-06-20
[BACKLOG-17078] Renaming /runInBackground to /invoke and providing a …

pedrofvteixeira:
…new param 'async' ( default fallback to 'true' )

To be merged alongside https://github.com/pentaho/pentaho-platform-ee/pull/1095

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/ActionResource.java
+++ b/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/ActionResource.java
@@ -35,6 +35,7 @@ import org.pentaho.platform.util.ActionUtil;
 import org.pentaho.platform.util.StringUtil;
 
 import javax.ws.rs.Consumes;
+import javax.ws.rs.DefaultValue;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.QueryParam;
@@ -42,6 +43,7 @@ import javax.ws.rs.core.Response;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.Map;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
@@ -66,7 +68,7 @@ public class ActionResource {
    * @return a {@link Response}
    */
   @POST
-  @Path( "/runInBackground" )
+  @Path( "/invoke" )
   @Consumes( { TEXT_PLAIN } )
   @StatusCodes( {
       @ResponseCode( code = 200, condition = "Action invoked successfully." ),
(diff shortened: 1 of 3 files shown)


pedrofvteixeira:
@pentaho/rogueone , @pentaho/rebelalliance please review

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/ActionResource.java
```diff
@@ -66,22 +65,44 @@
...
+
+      // default scenario for execution
+      executorService.submit( createRunnable( actionId, actionClass, actionUser, actionParams ) );
+      httpStatus = HttpStatus.SC_ACCEPTED;
+
+    } else {
+
+      try {
+
+        IActionInvokeStatus status = internalInvokeAction( actionId, actionClass, actionUser, actionParams );
```
pentaho-mzur: If I understand correctly, the purpose of introducing the internalInvokeAction() methods is to be able to get back the status of the run, that we don't have in Runnable. Instead, why not consider refactoring createRunnable to be a createCallable that returns a Callable.

Seems cleaner, and the asych vs sync will just invoke directly on the callable...like so:

    if (isAsyncExecution) {
    .... submit( createCallable(...) );
    }
.... else {
    ... IActionInvokeStatus status = createCallable(...).call();
}

pedrofvteixeira: I would consider this the simplest approach, having the least amount of impact: 
1. keep the "old" form ( a Runnable for async ) as it has always been ( given that it has been tried and tested )
2. Refactor the code, to accommodate for a sync execution.
3. Re-use existing logic ( refactored to go from within the _run()_ to an _internalInvokeAction()_ )

We can, at some later point in time, redo this code to have a Callable, but if this is an accepted solution for now, I would prefer we go with it.

pentaho-mzur: Well, I understand that it will save some time on refactoring the unit test maybe, but it is not really a big deal to change to Callable...why create this technical debt in the first place (that will probably never be addressed in the future).

pedrofvteixeira: Very well. I'll do the switch to Callable. Please stand by.

pedrofvteixeira:
@pentaho/rogueone @pentaho/rebelalliance please review. 
Rebased and squashed, now with the addition of having replaced Runnable with Callable.

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/ActionResource.java
```diff
@@ -139,30 +163,41 @@ IAction createActionBean( final String actionClass, final String actionId ) thro
...
-          logger.info( Messages.getInstance().getRunningInBgLocallySuccess( action.getClass().getName(), params ),
-            status.getThrowable() );
+        IActionInvokeStatus status = actionInvoker.invokeAction( action, user, params );
+
+        if ( status != null && status.getThrowable() == null ) {
+          getLogger().info( Messages.getInstance().getRunningInBgLocallySuccess( action.getClass().getName(), params ),
+                  status.getThrowable() );
         } else {
-          logger.error( Messages.getInstance().getCouldNotInvokeActionLocally( action.getClass().getName(), params ),
-            status.getThrowable() );
```
pentaho-mzur: This will result an NPE if status==null.

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/ActionResource.java
```diff
@@ -139,30 +163,41 @@ IAction createActionBean( final String actionClass, final String actionId ) thro
...
+          getLogger().error( Messages.getInstance().getCouldNotInvokeActionLocally( action.getClass().getName(), params ),
+                  status.getThrowable() );
```
pentaho-mzur: This will result an NPE if status==null.

PR #3471, opened 2017-03-14
[BACKLOG-11940] - Preview button not working (Dashboard-Parameterized…

Aliaksandr-Kastenka:
… query with no default)

PR #3668, opened 2017-06-22
[SP-3618] Backport of PDB-1898 - PDD - Metadata Query Editor preview throws error when condition default is null (6.1 Suite)

Advent51:
- backport of #3471


Advent51:

PR #3673, opened 2017-06-25
[BACKLOG-17201] Add ability to change log level at runtime.

niehusa:
Team, please review. This is to add a REST Endpoint so that we can change log level at runtime without restart.
This PR should be merged along with https://github.com/pentaho/pentaho-platform-ee/pull/1103

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/Log4jResource.java
+++ b/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/Log4jResource.java
@@ -1,0 +1,99 @@
+/*!
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License, version 2.1 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
+ * or from the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * Copyright (c) 2017 Pentaho Corporation..  All rights reserved.
+ */
+
+package org.pentaho.platform.web.http.api.resources;
+
+import org.apache.commons.lang.StringUtils;
+import org.apache.log4j.Level;
+import org.apache.log4j.LogManager;
+import org.apache.log4j.Logger;
+import org.apache.log4j.helpers.Loader;
+import org.apache.log4j.xml.DOMConfigurator;
+import org.codehaus.enunciate.jaxrs.ResponseCode;
+import org.codehaus.enunciate.jaxrs.StatusCodes;
+
+import javax.ws.rs.FormParam;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.Enumeration;
+
+
+@Path( "/logconfig" )
+public class Log4jResource {
+
+  private static final Logger LOGGER = Logger.getLogger( Log4jResource.class );
+  private static final String CONFIG = "log4j.xml";
+
+  @PUT
+  @Path ( "/reload" )
+  @StatusCodes( {
+    @ResponseCode( code = 200, condition = "Successfully reload from configuration." )
+    } )
+  @Produces( { MediaType.TEXT_PLAIN } )
+  public Response reloadConfiguration() throws Exception {
+    LOGGER.setLevel( Level.INFO );
+    LOGGER.info( "Reloading configuration..." );
+
+    DOMConfigurator.configure( Loader.getResource( CONFIG ) );
+    return Response.ok( "Done" ).build();
+  }
+
+  @PUT
+  @Path ( "/update" )
+  @StatusCodes( {
+    @ResponseCode( code = 200, condition = "Successfully update log level." ),
+    @ResponseCode( code = 304, condition = "Log level is not modified." )
+    } )
+  @Produces( { MediaType.TEXT_PLAIN } )
+  public Response updateLogLevel( @FormParam(  "level" ) String targetLevel, @FormParam( "category" ) String category ) throws Exception {
+    LOGGER.setLevel( Level.INFO );
+
+    if ( StringUtils.isBlank( targetLevel ) && StringUtils.isBlank( category ) ) {
+      return Response.notModified( "No parameter provided, log level not modified." ).build();
+    }
+
+    Logger root = LogManager.getRootLogger();
+
+    if ( StringUtils.isNotBlank( targetLevel ) ) {
+      LOGGER.info( "Request to set log level: " + targetLevel );
+
+      if ( StringUtils.isNotBlank( category ) ) {
+        LOGGER.info( "Request to set log level for package: " + category );
+        Logger catLog = LogManager.exists( category );
+        if ( catLog != null ) {
+          catLog.setLevel( Level.toLevel( targetLevel, root.getLevel() ) );
+          return Response.ok( "Setting log level for: '" + catLog.getName() + "' to be: " + catLog.getLevel() ).build();
+        }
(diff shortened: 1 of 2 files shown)


extensions/src/main/java/org/pentaho/platform/web/http/api/resources/Log4jResource.java
```diff
@@ -0,0 +1,103 @@
...
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.Enumeration;
+
+
+/**
+ * Created by on 6/23/2017.
```
pentaho-amessier: We probably don't need this message

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/Log4jResource.java
```diff
@@ -0,0 +1,103 @@
...
+
+  @PUT
+  @Path ( "/reset" )
+  @StatusCodes( {
+    @ResponseCode( code = 200, condition = "Successfully reset log level." )
+    } )
+  @Produces( { MediaType.TEXT_PLAIN } )
+  public Response resetLogLevel() throws Exception {
+    LOGGER.setLevel( Level.INFO );
+    LOGGER.info( "Resetting Log Level..." );
```
pentaho-amessier: Any message above "debug" needs to be externalized

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/Log4jResource.java
```diff
@@ -0,0 +1,103 @@
...
+
+
+/**
+ * Created by on 6/23/2017.
+ */
+@Path( "/logconfig" )
+public class Log4jResource {
+
+  private static final Logger LOGGER = Logger.getLogger( Log4jResource.class );
+  private static final String CONFIG = "log4j.xml";
```
pentaho-amessier: this makes sense as the default value, but should we allow this file name to be configurable through spring, just in case it changes?

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/Log4jResource.java
```diff
@@ -0,0 +1,103 @@
...
+
+  @PUT
+  @Path ( "/reset" )
+  @StatusCodes( {
+    @ResponseCode( code = 200, condition = "Successfully reset log level." )
+    } )
+  @Produces( { MediaType.TEXT_PLAIN } )
+  public Response resetLogLevel() throws Exception {
+    LOGGER.setLevel( Level.INFO );
```
pentaho-amessier: Can we assume that by "reset" we always want to set the log level to "info"? It might make sense to have the "default level" be configurable, so the user can control what "resetting" really means

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/Log4jResource.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @PUT
+  @Path ( "/update" )
+  @StatusCodes( {
+    @ResponseCode( code = 200, condition = "Successfully update log level." ),
+    @ResponseCode( code = 304, condition = "Log level is not modified." )
+    } )
+  @Produces( { MediaType.TEXT_PLAIN } )
+  public Response updateLogLevel( @FormParam(  "level" ) String targetLevel, @FormParam( "cat" ) String category ) throws Exception {
```
pentaho-amessier: any reason not to call the query string parameter the fill name "category" rather than "cat"?

extensions/src/test/java/org/pentaho/platform/web/http/api/resources/Log4jResourceTest.java
```diff
@@ -0,0 +1,90 @@
...
+  @Test
+  public void resetLogLevel() throws Exception {
+    Logger root = LogManager.getRootLogger();
+    Level initialLevel = root.getLevel();
+    assertNotEquals( initialLevel, Level.ALL );
+    root.setLevel( Level.ALL );
+    assertEquals( Level.ALL, root.getLevel() );
+
+    Response res = target.resetLogLevel();
+    assertNotEquals( Level.ALL,  LogManager.getRootLogger().getLevel() );
```
pentaho-amessier: rather than checking that the level is not equal to ALL, wouldn't it be better to check that it's equal to the expected value?

extensions/src/test/java/org/pentaho/platform/web/http/api/resources/Log4jResourceTest.java
```diff
@@ -0,0 +1,90 @@
...
+public class Log4jResourceTest {
+
+  private static final String CONFIG = "log4j.xml";
+  private Log4jResource target = new Log4jResource();
+
+  @Test
+  public void resetLogLevel() throws Exception {
+    Logger root = LogManager.getRootLogger();
+    Level initialLevel = root.getLevel();
+    assertNotEquals( initialLevel, Level.ALL );
```
pentaho-amessier: If the out of the box level is ever changes to ALL, this test will fail, will it not? Do we really need this assert statement? Instead, wouldn't it be better to pick a couple of different log levels, and verify that after setting to that level and resetting we once again have the original log level. By picking two different values we guarantee that this feature is being sufficiently testing, even if the original log level is changed to one of those two values

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/Log4jResource.java
```diff
@@ -0,0 +1,103 @@
...
+      LOGGER.info( "Request to set log level: " + targetLevel );
+
+      if ( StringUtils.isNotBlank( category ) ) {
+        LOGGER.info( "Request to set log level for package: " + category );
+        Logger catLog = LogManager.exists( category );
+        if ( catLog != null ) {
+          catLog.setLevel( Level.toLevel( targetLevel, root.getLevel() ) );
+          return Response.ok( "Setting log level for: '" + catLog.getName() + "' to be: " + catLog.getLevel() ).build();
+        }
+        return Response.notModified( "Category: '" + category + "' not found, log level not modified." ).build();
```
pentaho-amessier: is it possible to add a category if it doesn't exist? I think it would be immensely helpful. Currently, to see logs from the dispatching code, we have to add a number of categories that don't currently exist. We may want to just add them to log4j.xml and ship with the added category, but I think, if at all possible, It would go a long way to be able to add categories on the fly.

chiltonj:
Johnny, can you give a quick description of the Rest Apis?

niehusa:
@chiltonj 
curl -X PUT -d 'level=TRACE&cat=org.pentaho' --user admin:password http://localhost:8080/pentaho/api/logconfig/update

curl -X PUT --user admin:password http://localhost:8080/pentaho/api/logconfig/reset

chiltonj:
What does reset do? Do we need it? the caller should be able use update to reset.

niehusa:
Actually I just changed "reset" to be "reload", it's reloading from config file, log4j.xml.
I think we can add new category/package to log4j.xml, whenever it's ready, then call "reload" to have the new category loaded.

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/Log4jResource.java
niehusa: See my comment below. instead of adding category through API, I think add to log4j.xml then have the API reload it will do the trick.

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/Log4jResource.java
niehusa: It's just an attribute in the submitted form. package name can be long so I made the attribute name short. No particular reason.

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/Log4jResource.java
diogofscmariano: I would not call this a priority

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/Log4jResource.java
diogofscmariano: Since it is just a "word" I propose use category like @pentaho-amessier

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/Log4jResource.java
pentaho-amessier: I don't see any comment "below" :) , but as long as there's a way to accomplish this without a server reboot - which I believe is what you are describing - I think that's sufficient

chiltonj:
Coverage drops unrelated to these changes.

PR #3675, opened 2017-06-26
[BACKLOG-17085] created the capability to run actions (work-items) on startup of the pentaho server via a systemListener from json files

pentaho-esager:
- Unit Tests will come in a subsequent PR 
Summary of changes: 

ActionInvokerSystemListener: a System Listener which on startup of the pentaho server consumes json files  and runs the Work Item described by the json.  The json files are can be read from a configurable location as a property on the spring bean in SystemListeners.xml or the SystemListener will check the default-content folder if no location is set. 

WorkItemRequestPayload: Responsible for creating a work item from the json files.  Requires an actionId, actionClass, actionUser and actionParams in order for the payload to be valid. 

ActionResource and ActionResourceUtil:  Refactored existing code so that the SystemListener could reuse pre-existing code

ActionInvokerListenerTest: 1 successful unit test is included, more unit tests will come in a subsequent PR, code has been validated via by manual tests in which the server starts up and consumes different json files in various test cases.  

environment.json: A sample of a valid json file which will become a Work Item.

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
+++ b/extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
@@ -1,0 +1,171 @@
+/*!
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License, version 2.1 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
+ * or from the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * Copyright (c) 2017 Pentaho Corporation..  All rights reserved.
+ */
+package org.pentaho.platform.plugin.action;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+import org.pentaho.platform.api.engine.IPentahoSession;
+import org.pentaho.platform.api.engine.IPentahoSystemListener;
+import org.pentaho.platform.engine.core.system.PentahoSystem;
+import org.pentaho.platform.util.ActionUtil;
+import org.pentaho.platform.web.http.api.resources.ActionResource;
+
+import javax.ws.rs.core.Response;
+import java.util.HashMap;
+import java.util.Map;
+import java.io.FileNotFoundException;
+import java.io.File;
+import java.io.IOException;
+import java.io.FileReader;
+import java.io.FileFilter;
+
+public class ActionInvokerSystemListener implements IPentahoSystemListener {
+
+  private String environmentVariablesFilepath;
+  private static final Log logger = LogFactory.getLog( IPentahoSystemListener.class );
+  private static final String WORK_ITEM_FILE_EXTENSION = ".json";
+  private static final String DEFAULT_CONTENT_FOLDER = "system/default-content";
+
+  private Map<String, JSONObject> parseJsonFiles( )  {
+    Map<String, JSONObject> dotJsons = new HashMap<>();
+    final String solutionPath = PentahoSystem.getApplicationContext().getSolutionPath( DEFAULT_CONTENT_FOLDER );
+    if ( StringUtils.isEmpty( solutionPath ) ) {
+      logger.info( getClass().getName() + " Empty solution path" );
+    } else {
+      File[] jsonFiles = listFiles( new File( solutionPath ), WORK_ITEM_FILE_EXTENSION );
+      if ( jsonFiles != null ) {
+        for ( File jsonFile : jsonFiles ) {
+          JSONObject jsonObject = getJsonFromFile( jsonFile );
+          if ( jsonObject != null ) {
+            dotJsons.put( jsonFile.getAbsolutePath(), jsonObject );
+          }
+        }
+      }
+    }
+    return dotJsons;
+  }
+
+  private JSONObject getJsonFromFile( final File file )  {
+    FileReader fr;
+    try {
+      fr = new FileReader( file );
+    } catch ( FileNotFoundException e ) {
+      logger.error( e.getMessage() );
+      return null;
+    }
+    try {
+      JSONParser jsonParser = new JSONParser();
+      JSONObject jsonObject = (JSONObject) jsonParser.parse( fr );
+      String jsonString = jsonObject.toJSONString();
+      fr.close();
+      return (JSONObject) jsonParser.parse( jsonString );
+    } catch ( ParseException e ) {
+      logger.error( "File: " + file.getAbsolutePath() + " is not valid json, and will not be processed" );
+      try {
+        fr.close();
+      } catch ( IOException ex ) {
+        ex.getMessage();
+      }
+      //rename the file here and allow the chronos container to die because of bad input
+      renameFile( file.getAbsolutePath() );
+    } catch ( IOException e ) {
+      logger.error( e.getMessage() );
+      renameFile( file.getAbsolutePath() );
+    }
+    return null;
+  }
+
+  private File[] listFiles( final File folder, final String fileExtension ) {
+    if ( folder.isDirectory() && folder.canRead() ) {
+      return folder.listFiles( new FileFilter() {
+        @Override
+        public boolean accept( File f ) {
+          return f.isFile() && f.getName().toLowerCase().endsWith( fileExtension );
+        }
+      } );
+    } else {
+      logger.error( folder.getAbsolutePath() + " is not a valid directory" );
+    }
+    return null;
+  }
+
+  private void renameFile( final String filename ) {
+    File dotJson = new File( filename );
+    File dotJsonDotCurTime = new File( filename + "." + System.currentTimeMillis() );
+    dotJson.renameTo( dotJsonDotCurTime );
+  }
+
+  /**
+   * System listener looks for JSON files at start up and runs them as work items
+   * @param session
+   * @return
+   */
+  @Override
+  public boolean startup( IPentahoSession session ) {
+    Response status;
+    Map<String, JSONObject> jsonFiles = new HashMap<>(); //map between the json objects and the filenames from which they were consumed
+    if ( environmentVariablesFilepath != null ) {
+      JSONObject environmentVariables = getJsonFromFile( new File( environmentVariablesFilepath ) );
+      jsonFiles.put( environmentVariablesFilepath, environmentVariables );
+    } else {
+      jsonFiles = parseJsonFiles();
+    }
+    for ( String filename : jsonFiles.keySet() ) {
+      JSONObject json = jsonFiles.get( filename );
+      WorkItemRequestPayload payload = new WorkItemRequestPayload( json );
+      if ( !payload.isValid() ) {
+        logger.error( "File: " + filename + " is missing required parameters, skipping" );
+        renameFile( filename );
+      } else {
+        try {
+          status = issueRequest( payload );
+          if ( status == null ) {
+            logger.error( "File: " + filename + " has parameters formatted in an unparsable way. Could not process work item." );
+          }
+          //todo- sometype of status handling once WI lifecycle is defined
+        } catch ( Exception e ) {
+          logger.error( "Error processing " + filename + " skipping" );
+        } finally {
+          renameFile( filename );
+        }
+      }
+    }
+    return true;
+  }
+
+  private Response issueRequest( final WorkItemRequestPayload payload ) {
+    return new ActionResource().invokeAction(
+        ActionUtil.INVOKER_SYNC_VALUE,
+        payload.getActionId(),
+        payload.getActionClass(),
+        payload.getActionUser(),
+        payload.getActionParams( ) );
+  }
+
+  @Override
+  public void shutdown( ) {
+
+  }
+  public void setEnvironmentVariablesFilepath( String environmentVariablesFilepath ) {
+    this.environmentVariablesFilepath = environmentVariablesFilepath;
+  }
+}
--- a/extensions/src/main/java/org/pentaho/platform/plugin/action/WorkItemRequestPayload.java
+++ b/extensions/src/main/java/org/pentaho/platform/plugin/action/WorkItemRequestPayload.java
@@ -1,0 +1,73 @@
+/*!
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License, version 2.1 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
+ * or from the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * Copyright (c) 2017 Pentaho Corporation..  All rights reserved.
+ */
+
+package org.pentaho.platform.plugin.action;
+
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.json.simple.JSONObject;
+import org.pentaho.platform.api.engine.IPentahoSystemListener;
+import org.pentaho.platform.util.ActionUtil;
+import java.util.Arrays;
+
+public class WorkItemRequestPayload {
+
+  private String actionId;
+  private String actionClass;
+  private String actionParams;
+  private String actionUser;
+  private static final Log logger = LogFactory.getLog( IPentahoSystemListener.class );
+  private boolean isValid;
+
+  public WorkItemRequestPayload( JSONObject jsonEnvironmentVariables ) {
+    isValid = buildPayload( jsonEnvironmentVariables );
+  }
+
+  public String getActionId( ) {
+    return actionId;
+  }
+  public String getActionUser( ) {
+    return actionUser;
+  }
+  public String getActionClass( ) {
+    return actionClass;
+  }
+  public String getActionParams( ) {
+    return actionParams;
+  }
+
+  private boolean buildPayload( JSONObject json ) {
+    try {
+      actionClass = (String) json.get( ActionUtil.INVOKER_ACTIONCLASS );
+      actionId = (String) json.get( ActionUtil.INVOKER_ACTIONID );
+      actionParams = (String) json.get( ActionUtil.INVOKER_ACTIONPARAMS );
+      actionUser = (String) json.get( ActionUtil.INVOKER_ACTIONUSER );
+      return !StringUtils.isEmpty( actionClass )
+          && !StringUtils.isEmpty( actionId )
+          && !StringUtils.isEmpty( actionParams )
+          && !StringUtils.isEmpty( actionUser );
+    } catch ( Exception e ) {
+      logger.error( Arrays.toString( e.getStackTrace( ) ) );
+      //if any exception occurs json was not in valid format and thus return false
+      return false;
+    }
+  }
+  public boolean isValid( ) {
+    return isValid;
+  }
+}
--- a/extensions/src/test/resources/ActionInvokerListenerTestResources/environment.json
+++ b/extensions/src/test/resources/ActionInvokerListenerTestResources/environment.json
@@ -1,0 +1,6 @@
+{
+  "actionId" : "prpt.backgroundExecution",
+  "actionClass" : "org.pentaho.reporting.platform.plugin.SimpleReportingAction",
+  "actionUser" : "admin",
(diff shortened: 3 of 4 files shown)


pentaho-esager:
@pentaho/rogueone  @pentaho/rebelalliance

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -0,0 +1,155 @@
...
+
+  private String environmentVariablesFilepath;
+  private static final Log logger = LogFactory.getLog( IPentahoSystemListener.class );
+  private static final String WORK_ITEM_FILE_EXTENSION = ".json";
+  private static final String DEFAULT_CONTENT_FOLDER = "system/default-content";
+
+  private Map<String, JSONObject> parseJsonFiles() throws IOException {
+    Map<String, JSONObject> dotJsons = new HashMap<>(  );
+    final String solutionPath = PentahoSystem.getApplicationContext().getSolutionPath( DEFAULT_CONTENT_FOLDER );
+    if ( solutionPath == null || solutionPath.isEmpty() ) {
```
pentaho-amessier: You can use StringUtils.isEmpty (...) as a convenience method that does both of these checks

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -0,0 +1,155 @@
...
+      logger.info( getClass().getName() + " Empty solution path" );
```
pentaho-amessier: INFO messages need to be externalized - you can do it in a separate PR, but perhaps add a "TODO" comment above, so we don't forget

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -0,0 +1,155 @@
...
+          if ( jsonObject != null ) {
+            dotJsons.put( jsonFile.getAbsolutePath(), jsonObject );
+          }
+        }
+      }
+    }
+    return dotJsons;
+  }
+
+  private JSONObject getJsonFromFile( File file ) throws IOException {
```
pentaho-amessier: It's good practice to declare parameters as final - less change of a developer error or reassigning the value - the developer might think that doing so changes the value of the parameter as seen by the caller, which is not the case.

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -0,0 +1,155 @@
...
+    FileReader fr = new FileReader( file );
+    try {
+      JSONParser jsonParser = new JSONParser();
+      JSONObject jsonObject = (JSONObject) jsonParser.parse( fr );
+      String jsonString = jsonObject.toJSONString();
+      fr.close();
+      return (JSONObject) jsonParser.parse( jsonString );
+    } catch ( ParseException e ) {
+      logger.error( "File: " + file.getAbsolutePath() + " is not valid json, and will not be processed" );
```
pentaho-amessier: ERROR messages need to be externalinzed (anything other than DEBUG, really)

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -0,0 +1,155 @@
...
+      fr.close();
+      renameFile( file.getAbsolutePath() );
```
pentaho-amessier: Why rename only in case of exception? In general, some more comments would be helpful

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+    return null;
+  }
+
+  private File[] listFiles( final File folder, final String fileExtension ) throws IOException {
+    if ( folder.exists() && folder.isDirectory() && folder.canRead() ) {
```
pentaho-amessier: the exists() check is redundant, isDirectory() is sufficient, it will also return false if the file does not exist

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionResourceUtil.java
```diff
@@ -0,0 +1,72 @@
...
+import org.pentaho.platform.util.StringUtil;
+import org.slf4j.MDC;
+
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.Map;
+import java.util.Optional;
+
+public class ActionResourceUtil {
```
pentaho-amessier: Do we need a separate utility class for this? Why not use ActionUtils? Which incidentally contains some of the methods declared here (for instance createActionBean)

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionResourceUtil.java
```diff
@@ -0,0 +1,72 @@
...
+
+  private static IAction createActionBean( final String actionClass, final String actionId ) throws Exception {
+    return ActionUtil.createActionBean( actionClass, actionId );
+  }
+
+  private static Map<String, Serializable> deserialize( IAction action, final String actionParams ) throws IOException, ActionInvocationException {
+    return ActionParams.deserialize( action, ActionParams.fromJson( actionParams ) );
+  }
+
+  public static IActionInvokeStatus runWorkItemRequest( String actionClass, String actionId, String actionParams, String user, IActionInvoker actionInvoker, Map< String, String > mdcContextMap ) {
```
pentaho-amessier: I would call this invokeAction for consistency with all other code

extensions/src/main/java/org/pentaho/platform/plugin/action/WorkItemRequestPayload.java
```diff
@@ -0,0 +1,101 @@
...
+  private final String actionId = "actionId";
+  private final String actionUser = "actionUser";
+  private final String actionClass = "actionClass";
+  private final String actionParams = "params";
+  private final String environmentVariablesKey = "environmentVariables";
+  private static final Log logger = LogFactory.getLog( IPentahoSystemListener.class );
+  private Map<String, String> environmentVariables;
+  private boolean isValid;
+
+  public WorkItemRequestPayload( JSONObject jsonEnvironmentVariables ) {
```
pentaho-amessier: Should an invalid object even be able to exist? It might make sense to throw an exception here and not even let the object be created (IllegalArgumentException or some other specialized exception)

extensions/src/main/java/org/pentaho/platform/plugin/action/WorkItemRequestPayload.java
```diff
@@ -0,0 +1,101 @@
...
+import org.pentaho.platform.api.engine.IPentahoSystemListener;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+public class WorkItemRequestPayload {
+
+
+  private final String actionId = "actionId";
```
pentaho-amessier: We already have constants for these in the ActionUtils class. Please use those instead.

extensions/src/main/java/org/pentaho/platform/plugin/action/WorkItemRequestPayload.java
```diff
@@ -0,0 +1,101 @@
...
+  private final String actionUser = "actionUser";
+  private final String actionClass = "actionClass";
+  private final String actionParams = "params";
+  private final String environmentVariablesKey = "environmentVariables";
```
pentaho-amessier: These are constants, they should be final. Also, once they're final, they should be ALL_CAPS with underscores for word separation - standard practice.

extensions/src/main/java/org/pentaho/platform/plugin/action/WorkItemRequestPayload.java
```diff
@@ -0,0 +1,101 @@
...
+    boolean actionIdSet = false, actionClassSet = false, actionUserSet = false, actionParamsSet = false;
+
+    try {
+      JSONArray jsonArray = (JSONArray) json.get( environmentVariablesKey );
+      if ( json == null || StringUtils.isEmpty( jsonArray.toJSONString() ) ) {
+        return false;
+      }
+      for ( Object var : jsonArray ) {
+        JSONObject environmentVariable = (JSONObject) var;
+        String name = (String) environmentVariable.get( "name" );
```
pentaho-amessier: magic strings, these should be constants

extensions/src/main/java/org/pentaho/platform/plugin/action/WorkItemRequestPayload.java
```diff
@@ -0,0 +1,101 @@
...
+    return environmentVariables.get( actionUser );
+  }
+  public String getActionClass( ) {
+    return environmentVariables.get( actionClass );
+  }
+  public String getActionParams( ) {
+    return environmentVariables.get( actionParams );
+  }
+
+  private boolean buildPayload( JSONObject json ) {
```
pentaho-amessier: Seems like a lot of work just to verify that the json is valid. Perhaps there is some 3rd party utility to do that for us? Something to consider

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/ActionResource.java
```diff
@@ -131,11 +130,11 @@ protected CallableAction createCallable( final String actionId, final String act
    *
    * @return the {@link IActionInvoker}
    */
-  IActionInvoker getActionInvoker() {
+  IActionInvoker getActionInvoker( ) {
```
pentaho-amessier: Why the change? The original code did not violate checkstyle. Also, in general, it is best to avoid checkstyle changes unrelated to the code you are changing directly, which making functional code changes, because it makes the review longer and increases the possibility of merge conflicts.

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/ActionResource.java
```diff
@@ -149,12 +148,15 @@ DefaultActionInvoker getDefaultActionInvoker() {
...
-    CallableAction() {
+    private Map< String, String > mdcContextMap = MDC.getCopyOfContextMap();
+
+    CallableAction( ) {
+
     }
 
     public CallableAction( final ActionResource resource, final String actionId, final String actionClass, final
-      String user, final String actionParams ) {
+    String user, final String actionParams ) {
```
pentaho-amessier: this is a checkstyle violation - the original indentation was correct

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/ActionResource.java
```diff
@@ -166,46 +168,25 @@ IAction createActionBean( final String actionClass, final String actionId ) thro
...
+      // instantiate the DefaultActionInvoker directly to force local invocation of the action
+      final IActionInvoker actionInvoker = resource.getDefaultActionInvoker();
+      return ActionResourceUtil.runWorkItemRequest( actionClass, actionId, actionParams, user, actionInvoker, mdcContextMap );
     }
-  }
 
-  public static Log getLogger() {
-    return logger;
   }
+    public static Log getLogger( ) {
```
pentaho-amessier: this will be a checkstyle violations - indentations are off

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
pedrofvteixeira: Don't see the purpose for this method, try this one-liner instead

`JSONObject jsonObj = new JSONObject ( IOUtils.toString( new FileInputStream( file ) ) );`

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionResourceUtil.java
pedrofvteixeira: Agreed. Don't see a purpose for this ActionResourceUtil class. ActionUtil should already be giving us all we need.

extensions/src/test/resources/ActionInvokerListenerTestResources/environment.json
```diff
@@ -0,0 +1,8 @@
+{
+  "environmentVariables": [
```
befc: This **environmentVariables** object is used to pass the work item parameters from the main pentaho server to Chronos, and in my opinion, shouldn't be used here, as it's more complex for reading and not so straightforward to rebuild from the env. variables available inside the docker image.

For example, a much more simple approach would be: 

`{
  "actionId":"prpt.backgroundExecution",
  "actionClass":"org.pentaho.reporting.platform.plugin.SimpleReportingAction",
  "actionUser":"admin",
  "actionParams":"{ \"serializedParams\": [\"java.util.HashMap\", { \"actionUser\": \"admin\", \"user_locale\": [\"java.util.Locale\", \"en_US\"],\"outputFilePattern\": \"/home/admin/Buyer Report (sparkline report).*\", \"line\": \"Trains\", \"renderMode\": \"PARAMETER\", \"accepted-page\": 0, \"inputFile\": \"/public/Steel Wheels/Buyer Report (sparkline report).prpt\", \"uniqueFileName\": true, \"output-target\": \"table/html;page-mode=page\", \"maximum-query-limit\": 0, \"htmlProportionalWidth\": \"false\", \"query-limit-ui-enabled\": \"true\", \"lineage-id\": \"1bb60f2b-1c0d-43ac-ba5d-4003b6c7e785\", \"actionId\": \"prpt.backgroundExecution\", \"showParameters\": \"true\", \"query-limit\": 0}] \"paramsToRecreate\": [\"::session\", \"streamProvider\"]}"`

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
pentaho-esager: I am using a different JSON library where that operation is a little more complicated, so I abstracted it to a different method for organizational/readability purposes

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/ActionResource.java
```diff
@@ -166,46 +168,25 @@ IAction createActionBean( final String actionClass, final String actionId ) thro
       return ActionUtil.createActionBean( actionClass, actionId );
     }
 
-    Map<String, Serializable> deserialize( final IAction action, final String actionParams )
-            throws IOException, ActionInvocationException {
+    Map< String, Serializable > deserialize( final IAction action, final String actionParams )
```
befc: checkstyle violation, the original was correct

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionResourceUtil.java
pentaho-esager: ActionUtil is in core, whereas this is in extensions, I think it will be helpful as most of the code for WN is in extensions, this code does not need to be accessed by all of core, so I created a second file, but I am happy to move it if ActionUtil is the better place

extensions/src/main/java/org/pentaho/platform/plugin/action/WorkItemRequestPayload.java
pentaho-esager: ok - will change

extensions/src/main/java/org/pentaho/platform/plugin/action/WorkItemRequestPayload.java
pentaho-esager: ok-- will change

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionResourceUtil.java
pentaho-amessier: extensions code has access to core, so there're really no need for a separate class. I understand that the code you wrote specifically may not be accessed outside of extensions, but that's not enough of a reason to create a separate class and limit access to this functionality - just my two cents :)

extensions/src/test/resources/ActionInvokerListenerTestResources/environment.json
pentaho-esager: Yes, I agree that would be more simple.  However, I was under the impression that the parameters were going to be passed in an outer environmentVariables object from docker?  If this is not the case I can change it.

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
pentaho-esager: it gets renamed later if this exception doesn't occur. renaming it at this point in the code would cause the chronos container to die which we only want to occur if we are at an end stage (exception or success).

pentaho-amessier: ok sure, a comment to clarify this would be helpful, I think

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
pentaho-esager: ok thanks, I did not know that, I can change it.

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionResourceUtil.java
pentaho-esager: I named it work item request to avoid confusion because the method contract is different than the API invokeAction

pentaho-amessier: Why is the method contract different? Maybe you can document it to avoid confusion.

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -0,0 +1,156 @@
...
+        if ( !payload.isValid() ) {
+          logger.error( "File: " + filename + " is missing required parameters, skipping" );
+        } else {
+          status = issueRequest( payload );
+          if ( status == null ) {
+            logger.error( "File: " + filename + " has parameters formatted in an unparsable way. Could not process work item." );
+          }
+          //todo- sometype of status handling once WI lifecycle is defined
+        }
+        renameFile( filename );
```
pedrofvteixeira: Suggestion: adding a try-finally clause inside the for loop, so that whatever ends up happening ( successful execution, NPE or some other unchecked exception ), we'll always be renaming the file, appending it the timestamp:

`for ( String filename : jsonFiles.keySet() ) {     
     
     try {
         JSONObject json = jsonFiles.get( filename );
         WorkItemRequestPayload payload = new WorkItemRequestPayload( json );
         if ( !payload.isValid() ) {
           logger.error( "File: " + filename + " is missing required parameters, skipping" );
         } else {
           status = issueRequest( payload );
           if ( status == null ) {
             logger.error( "File: " + filename + " has parameters formatted in an unparsable way. Could not process work item." );
           }
           //todo- sometype of status handling once WI lifecycle is defined
         }
      } finally {
         renameFile( filename );
      }
}`

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -0,0 +1,156 @@
...
+      }
+    } catch ( ParseException | IOException | SecurityException e ) {
+      logger.error( e.getMessage(), e );
+    }
+    return true;
+  }
+
+  private Response issueRequest( final WorkItemRequestPayload payload ) throws IOException, ParseException {
+    return new ActionResource().invokeAction(
+        ActionUtil.INVOKER_SYNC_VALUE,
```
pedrofvteixeira: Just a thought: 

don't think we need an extra constant in ActionUtil for this. If we don't wan't to stick a magic number here ( which is the correct approach ), we can either

a)  pass a String.valueOf( true )
b)  declare it as a final static string within this systemListener class

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -0,0 +1,170 @@
...
+          renameFile( filename );
+        }
+      }
+    }
+    return true;
+  }
+
+  private Response issueRequest( final WorkItemRequestPayload payload ) {
+    return new ActionResource().invokeAction(
+        ActionUtil.INVOKER_SYNC_VALUE,
```
pedrofvteixeira: just a thought:

Don't think we need another ActionUtil constant for this. 
If we don't care about sticking some magic number here ( which is the correct approach ), we can:

a) declare it as a constant of this systemListener class
b) pass a `String.valueOf( true )`
c) pass a `!Boolean.parseValue( ActionUtil.INVOKER_DEFAULT_ASYNC_EXEC_VALUE )`

pedrofvteixeira:

neeta-rathi:
Look good!

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -0,0 +1,171 @@
...
+    try {
+      fr = new FileReader( file );
+    } catch ( FileNotFoundException e ) {
+      logger.error( e.getMessage() );
+      return null;
+    }
+    try {
+      JSONParser jsonParser = new JSONParser();
+      JSONObject jsonObject = (JSONObject) jsonParser.parse( fr );
+      String jsonString = jsonObject.toJSONString();
```
pentaho-mzur: Why get a string from the json object, just so parse it back into a json?

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -0,0 +1,171 @@
...
+      fr.close();
+      return (JSONObject) jsonParser.parse( jsonString );
+    } catch ( ParseException e ) {
+      logger.error( "File: " + file.getAbsolutePath() + " is not valid json, and will not be processed" );
+      try {
+        fr.close();
```
pentaho-mzur: It seems fr.close() is being closed twice...??

extensions/src/main/java/org/pentaho/platform/plugin/action/WorkItemRequestPayload.java
```diff
@@ -0,0 +1,73 @@
...
+import org.pentaho.platform.util.ActionUtil;
+import java.util.Arrays;
+
+public class WorkItemRequestPayload {
+
+  private String actionId;
+  private String actionClass;
+  private String actionParams;
+  private String actionUser;
+  private static final Log logger = LogFactory.getLog( IPentahoSystemListener.class );
```
pentaho-mzur: IPentahoSystemListeener...??

extensions/src/main/java/org/pentaho/platform/plugin/action/WorkItemRequestPayload.java
```diff
@@ -0,0 +1,73 @@
...
+      actionClass = (String) json.get( ActionUtil.INVOKER_ACTIONCLASS );
+      actionId = (String) json.get( ActionUtil.INVOKER_ACTIONID );
+      actionParams = (String) json.get( ActionUtil.INVOKER_ACTIONPARAMS );
+      actionUser = (String) json.get( ActionUtil.INVOKER_ACTIONUSER );
+      return !StringUtils.isEmpty( actionClass )
+          && !StringUtils.isEmpty( actionId )
+          && !StringUtils.isEmpty( actionParams )
+          && !StringUtils.isEmpty( actionUser );
+    } catch ( Exception e ) {
+      logger.error( Arrays.toString( e.getStackTrace( ) ) );
```
pentaho-mzur: you can just pass 'e' to error() and it will log the stack trace for you...

PR #3677, opened 2017-06-27
[BACKLOG-17293] Included config for 'pentaho/context' module

davidmsantos90:
@pentaho/millenniumfalcon please review

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/web/http/filters/PentahoWebContextFilter.java
+++ b/extensions/src/main/java/org/pentaho/platform/web/http/filters/PentahoWebContextFilter.java
@@ -44,20 +44,22 @@ import javax.servlet.http.HttpServletResponse;
 
 import java.io.IOException;
 import java.io.OutputStream;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 
 /**
  * If the request is searching for a webcontext.js, it writes out the content of the webcontext.js
- *
- * @author Ramaiz Mansoor
- *
  */
 public class PentahoWebContextFilter implements Filter {
 
   public static final String WEB_CONTEXT_JS = "webcontext.js"; //$NON-NLS-1$
+
+  private static final String REQUIREJS_LOCATION = "content/common-ui/resources/web/require.js";
+  private static final String REQUIREJS_CONFIG_LOCATION = "content/common-ui/resources/web/require-cfg.js";
+  private static final String REQUIREJS_INIT_LOCATION = "osgi/requirejs-manager/js/require-init.js";
+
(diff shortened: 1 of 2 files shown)


davidmsantos90:
@pentaho/millenniumfalcon please review.

included your reviews and also unit tests

extensions/src/main/java/org/pentaho/platform/web/http/filters/PentahoWebContextFilter.java
```diff
@@ -44,20 +44,22 @@
...
- *
- * @author Ramaiz Mansoor
- *
  */
 public class PentahoWebContextFilter implements Filter {
 
   public static final String WEB_CONTEXT_JS = "webcontext.js"; //$NON-NLS-1$
+
+  private static final String REQUIREJS_LOCATION = "content/common-ui/resources/web/require.js";
+  private static final String REQUIREJS_CONFIG_LOCATION = "content/common-ui/resources/web/require-cfg.js";
```
graimundo: It seems odd to me that we need direct dependency on resources from common-ui. Is there no other way around this?

davidmsantos90:
@pentaho/millenniumfalcon please review

The failing unit test already fails in previous PRs and it is being investigated

extensions/src/main/java/org/pentaho/platform/web/http/filters/PentahoWebContextFilter.java
dcleao: The pentaho platform project itself could provide the require.js and initializer file...
Or, maybe when we split common-ui?

davidmsantos90:
@dcleao did the change to only calculate the values one time and not use the values cache.
@pentaho/millenniumfalcon please review

extensions/src/main/java/org/pentaho/platform/web/http/filters/PentahoWebContextFilter.java
davidmsantos90: @graimundo @dcleao Regarding `[require-cfg.js](https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/assemblies/platform-plugin/src/main/resources/resources/web/require-cfg.js)` file, it is defining `pen.require` and `pen.define`. If I remember correctly this is something that we want to remove eventually because it is legacy, but it is still being used by pentaho projects.

davidmsantos90:
@pentaho/millenniumfalcon wingman seems to be drunk again. The `+` (checkstyle)[https://github.com/pentaho/pentaho-platform/pull/3677/files#diff-d18f1e7f5b58064245f057641e8d5ae4R501] error isn't correct, as there is no `+` on that line. And the rest of the errors is the usual random behaviour

PR #3680, opened 2017-06-27
[Unit tests] Added unit tests to WorkItemRequestPayload

pentaho-esager:


pdesai16:
The coverage drop shown by wingman is not touched in the change. Merging the change

PR #3681, opened 2017-06-27
[TEST] Adds test cases for the ActionInvokerSystemListener

pentaho-mracine:
@pentaho/rogueone

Tests for the ActionInvokerSystemListener. We don't really have a way to verify all codepaths are being hit, but we can trigger them with these JSON files. Please let me know if there's any improvements I can make.

I'm also a little concerned at the runtime of this test, not sure if there's a way around it.


pentaho-mracine:
Closing as we have more changes incoming.

PR #3684, opened 2017-06-29
[BACKLOG-17085] Adapted System Listener to handle URL encoded parameters

pentaho-esager:
This PR is to handle action params. Allows for a .json file to be consumed with URL encoded values during the startup of the pentaho server.


extensions/src/main/java/org/pentaho/platform/plugin/action/Payload.java
```diff
@@ -0,0 +1,73 @@
+/*!
```
pedrofvteixeira: Instead of creating a new file, could we have `Payload` declared as an inner class of `ActionInvokerSystemListener` ?

pentaho-esager:
Unit tests will be in a different PR

pdesai16:
The broken test in XmlHelperTest is not related to the changes in this commit.

pentaho-esager:

PR #3686, opened 2017-06-30
[Cleanup] ActionParams auto serialization/deserialization.

pentaho-mzur:
This PR should be reviewed and merged together with https://github.com/pentaho/pentaho-platform-ee/pull/1117.

The purpose of this PR is to add a JAXRS message-body provider  for jackson (faster xml) json serialization. This provider will handle POJOs decorated with the JsonRootName annotation (just like JAXB POJOs are decorated with XmlRootElement). The reason for this so the provider does not interfere with the "main path" of handling json, but only when asked explicitly (by the virtue of using this decoration). Note the use of this annotation in the ActionParams class. 

This change allows a proper signature for the REST resources (not only the ActionResource) moving foward instead of failing back to plain text and having the user code in the resource be responsible for writing it themselves.  It will also allows us to move to using jackson faster xml which has more to offer in terms of json serialization.  (cleaner, and easier to to maintain code).

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
+++ b/extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
@@ -145,9 +145,10 @@ public class ActionInvokerSystemListener implements IPentahoSystemListener {
       buildPayload( json, LocaleHelper.UTF_8 );
     }
 
-    public Response issueRequest( ) {
+    public Response issueRequest( ) throws IOException {
       ActionResource actionResource = new ActionResource();
-      return actionResource.invokeAction( ActionUtil.INVOKER_SYNC_VALUE, actionId, actionClass, actionUser, actionParams );
+      return actionResource.invokeAction( ActionUtil.INVOKER_SYNC_VALUE, actionId, actionClass, actionUser,
+        ActionParams.fromJson( actionParams ) );
     }
   }
 }
--- a/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/ActionResource.java
+++ b/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/ActionResource.java
@@ -23,12 +23,9 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.codehaus.enunciate.jaxrs.ResponseCode;
 import org.codehaus.enunciate.jaxrs.StatusCodes;
-import org.pentaho.platform.api.action.ActionInvocationException;
 import org.pentaho.platform.api.action.IAction;
 import org.pentaho.platform.api.action.IActionInvokeStatus;
 import org.pentaho.platform.api.action.IActionInvoker;
-import org.pentaho.platform.workitem.WorkItemLifecycleEvent;
-import org.pentaho.platform.workitem.WorkItemLifecyclePhase;
 import org.pentaho.platform.engine.core.system.PentahoSessionHolder;
 import org.pentaho.platform.engine.core.system.PentahoSystem;
 import org.pentaho.platform.plugin.action.ActionParams;
@@ -36,8 +33,10 @@ import org.pentaho.platform.plugin.action.DefaultActionInvoker;
 import org.pentaho.platform.plugin.action.messages.Messages;
 import org.pentaho.platform.util.ActionUtil;
 import org.pentaho.platform.util.StringUtil;
-import org.slf4j.MDC;
+import org.pentaho.platform.workitem.WorkItemLifecycleEvent;
+import org.pentaho.platform.workitem.WorkItemLifecyclePhase;
 import org.pentaho.platform.workitem.WorkItemLifecyclePublisher;
+import org.slf4j.MDC;
 
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DefaultValue;
@@ -45,7 +44,6 @@ import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Response;
-import java.io.IOException;
 import java.io.Serializable;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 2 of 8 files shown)


extensions/src/main/java/org/pentaho/platform/web/http/api/resources/ActionResource.java
```diff
@@ -88,14 +86,14 @@ public Response invokeAction(
     @QueryParam( ActionUtil.INVOKER_ACTIONID ) String actionId,
     @QueryParam( ActionUtil.INVOKER_ACTIONCLASS ) String actionClass,
     @QueryParam( ActionUtil.INVOKER_ACTIONUSER ) String actionUser,
-    final String actionParams ) {
+    final ActionParams actionParams ) {
```
pedrofvteixeira: This change will cause the ActionInvokerSystemListener to fail, as it's calling for the `ActionResource().invokeAction( String , String , String, String )`

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -145,9 +145,10 @@ public Payload( String json ) {
       buildPayload( json, LocaleHelper.UTF_8 );
     }
 
-    public Response issueRequest( ) {
+    public Response issueRequest( ) throws IOException {
       ActionResource actionResource = new ActionResource();
-      return actionResource.invokeAction( ActionUtil.INVOKER_SYNC_VALUE, actionId, actionClass, actionUser, actionParams );
+      return actionResource.invokeAction( ActionUtil.INVOKER_SYNC_VALUE, actionId, actionClass, actionUser,
+        ActionParams.fromJson( actionParams ) );
```
pdesai16: Do we need to call this method here ?

pentaho-mzur: Since now the REST resource signature now accepts ActionParams object (which is already de-serialized) we need to de-serialize here before calling the resource. Any problem with this? (or am I missing your point?)

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/ActionResource.java
pentaho-mzur: ActionResource.invokeAction signature was changed to ActionResource().invokeAction( String , String , String, ActionParams). This is the whole point of this PR...??

pentaho-mzur: Meant to say: ActionResource().invokeAction( String , String , ActionParams)

pentaho-esager: Yes this will cause the systemListener to fail as it was, although I see you made a change to adapt the listener to this change.   Before merging this please test that the system listener still behaves as expected. A concern with this change is that it forces the caller of this method to know how to create an instance of ActionParams.  The three methods which return actionParams from this class are very confusing.  

**public static ActionParams serialize( final IAction action, final Map<String, Serializable> params ) :** 
a method on the class called serialize which does not return a serialized object but rather an instance of actionParams. 

**public static ActionParams fromJson( final String params ):** 
this method does not accept a normal json string, a json object nor a json file, but rather escaped json which can only be generated from ActionParams.toJson().    I would like a method on this class that can go between normal (unescaped) json and the escaped json need by Jackson so that we don't run into issues with escape characters. 
**empty constructor:** 
I understand this has to be here for Jackson to work correctly but it would return a completely unusable version of ActionParams if called from an outside class. 

All this logic makes it very confusing for the caller of this class to (correctly) create an instance of ActionParams.

pentaho-mzur: I am not sure why you are saying it will break the listener as the listener now was changed to pass in a different 4th argument to action invoker of type ActionParams instead of String. Why is this an issue? I am not familiar with chronos testing yet but with R1 help I can verify this and see if it breaks anything.

Secondly - the serialize()/fromJson() methods they have separate concerns, the latter is related to json serialization deserialization we are talking about here. The former is related to converting to the Map required used during action invocation (maybe the name can be refactored if that is confusing to you - something like getParamsMap()  etc.). 

Thirdly why creating an instance of ActionParams is a concern? (especially when the conversion method is a black box to you). Remember that the listener is nothing but an adopter to invoke an action, so there is no concern if it "adopts" - if you want I can add another method to the resource that will just call this internally - but I don't think this is cleaner - this is a REST resource that accepts JSON and ideally the signature should look like it is now.

I understand that you guys had a lot of issues with the fact that chronos did not deal well with the json produced by ActionParams, But I am not sure it is invalid Json, maybe more "sophisticated" that requires escaping. Now, we could remove that special type information added to the json but then we introduce the risk of not being able to recreate the params on the worker now I think ...??

chiltonj: I am not sure SystemListener should be invoking ResourceAction explicitly like that. I think it should call to an interface and not the rest endpoint directly.

chiltonj: Also, I find String to me more opaque  then an explicit data model.  A string has no defined structure making it hard to know what correct value is. Since its a Rest endpoint that accepts JSON it makes sense to let the Jax layer convert json to the POJO.

chiltonj: But I agree that ActionParam class seems confusing.

pentaho-mzur: It's fine if ActionParams seems confusing this can be addressed in a separate (cleanup) PR.

pentaho-mzur: If there are any concerns still regarding this breaking the listener - just verified/tested it and it works fine....

pedrofvteixeira: > If there are any concerns still regarding this breaking the listener - just verified/tested it and it works fine....

Then I am good with the changes 👍

pedrofvteixeira: > I am not sure SystemListener should be invoking ResourceAction explicitly like that. I think it should call to an interface and not the rest endpoint directly.

Absolutely correct. 
In tine ( cleanup process ), we will step away from calling for `new ActionResource().invokeAction( String , .. )`, and into programming to interfaces, via IActionInvoker invoker = PentahoSystem.get( IActionInvoker.class ); invoker.invokeAction( String, ... )

PR #3687, opened 2017-06-30
[CLEANUP] Cleanup related to BACKLOG-17100: Improved WorkItemLifecyclePhase enum and configurability of the work-item-status log message

pentaho-amessier:
Relevant unit tests pass and local smoke tests have been performed. I'm still running more thorough tests, and will update this PR description, once those are done, but wanted to get the PR our sooner so people can start looking into it.

### DIFF
--- a/core/src/main/java/org/pentaho/platform/util/ObjectSerializationUtil.java
+++ b/core/src/main/java/org/pentaho/platform/util/ObjectSerializationUtil.java
@@ -1,0 +1,334 @@
+/*!
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License, version 2.1 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
+ * or from the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * Copyright (c) 2017 Pentaho Corporation. All rights reserved.
+ */
+
+package org.pentaho.platform.util;
+
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.util.ClassUtils;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A specialized utility class that converts a complex {@link Object} into a {@link Map} that can be safely
+ * serialized. Many serialization mechanisms do not handle complex objects well and even throw exceptions. This
+ * utility can be used to "normalize" any Object, turning into a {@link Map} of easily serializable values, where
+ * they map keys correspond to the Object's declared fields and methods, and the value is either a String, primitive
+ * or another "normalized" {@link Map}.<br><br>
+ * <p>
+ * This utility class also provides convenience methods for reading object "properties", where given a "propertyName",
+ * an attempt is made to read that value directly via reflection, or call one of the most likely corresponding getter
+ * methods, or simply use the property name as a key, if the object being read is a {@link Map}.<br><br>
+ * <p>
+ * Example: Let's assume we have the following classes:<br>
+ * <pre>
+ *  public class Company {
+ *    private String name;
+ *    private Map<String, Employee> employees = new HashMap<String, Employee>();
+ *    public Company( final String name ) {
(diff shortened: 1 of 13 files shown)


chiltonj:
Can we just use log4j patterns to control log output?

pentaho-amessier:
@chiltonj regarding your comment on using log4j patterns, it doesn't provide the fine grained control we require. We want to be able to log only specific pieces of information from the event object. Log for j only goes as far as letting us print a "message", we need to provide the message with the content we require, and this code is all about formatting the contents of that message.


https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/PatternLayout.html

chiltonj:
I am not sure we have a requirement to make this configurable. I do think something could be done using log4j MDC to control what gets written. I am sure this is necessary at this juncture since any additional will need to be tested and maintained. Also we only see a few pieces of metadata to track a lifecycle.

pentaho-mzur:
I think I understand the motivation for this solution - i.e. wanting to have/be able to configure fields to be output-ed (for events) but I am not sure I buy-in to this specific solution. We are going "out of our way" to provide this configuration ability when it feels we don't really need it. It feels like complicating things unnecessarily.  But I don't mind being convinced otherwise - maybe I don't see the benefits fully. The utility for serialization seems like a "infrastructure" class that needs to be tightened up. E.g the DEPTH level etc.

pentaho-amessier:
@pentaho-mzur and @chiltonj - the current message includes the entire params map, which I've been told is undesirable. Removing the map from output removes all details that identify the work item, and it will be impossible to distinguish one work item from another when multiple run at the same time. This work is to address this problem and to allow enough configurability to pick individual pieces form the map. The solution may be overkill, but it's generic.

chiltonj:
Why doesn't  a unique Id allow differentiation between multiple work items?

pentaho-amessier:
Unique Id (randomly generated) allows us to group status logs for a given work item, but how do we know which work item the logs belong to? The params contain information like 'inpufFile' which help us with that (tells us which report we tried to generate or which transformation we just ran). Without the params map, the log is just a bunch of timestamps, status and the unique ID. Granted, if you fire 20 requests to execute the same exact report, inputFile won't necessarily help you identify each run.

chiltonj:
Sounds like the problem is: how do we uniquely identify a single work item across multiple runs beyond using an opaque UID? What pieces of meta data can be used to create a common prefix? Is the actionId of any value here? I don't think making logs configurable is correct solution. We may need to add in a human readable token that provides a prettier label in conjunction w/ the UID.

pentaho-mzur:
If we wanted to be "economical" and light-weight while handling events (e.g. not introspecting and creating transformations/serializations) , could such an id (with a human readable aspect to it) not eventually be used to correlate or find the full details about work item executed? (e.g. from the logs, which is a complementary drill down tool) Do we track/provide this information today unrelated to wn?

pentaho-amessier:
actionId is something like  prpt.runInBackground or prpt.rurInBackground- it only tells us what kind of work item we ran (prpt -> simple report, ktr -> transformation). The information that can identify the work item uniquely is in the map, and we need a way to read specific pieced from the map. I'm happy to hard code something to read the pieces we know about today (inputFile for instance), but I thought a generic/configurable solution would be more desirable

pentaho-amessier:
@pentaho-mzur we can remove the params map from the log message entirely and use the work item uid to correlate the status to other logs that contain all the details, but then this goes against the spirit of this entire endevour, which was to provide all the necessary details within a single location. "Human readable" is the tricky part, how to we get it without hard coding? The params map is what contains these human readable components...

pentaho-amessier:
Anyway, looks like this solution is a no go :), but for completeness, here's a sample output from the new code, configured to pull out the inputFile from the map:

1498859402391|dc51befd-92bc-4708-9bf4-5b38bd51decc|Submitted|null|1498859402391|PC0FV9LS.pentaho.com|192.168.33.1|null|null
1498859402850|dc51befd-92bc-4708-9bf4-5b38bd51decc|Dispatched|null|1498859402850|PC0FV9LS.pentaho.com|192.168.33.1|admin|/public/Steel Wheels/Buyer Report (sparkline report).prpt
1498859404210|dc51befd-92bc-4708-9bf4-5b38bd51decc|Received|null|1498859404210|PC0FV9LS.pentaho.com|192.168.33.1|admin|/public/Steel Wheels/Buyer Report (sparkline report).prpt
1498859404451|dc51befd-92bc-4708-9bf4-5b38bd51decc|In progress|null|1498859404451|PC0FV9LS.pentaho.com|192.168.33.1|admin|/public/Steel Wheels/Buyer Report (sparkline report).prpt
1498859407217|dc51befd-92bc-4708-9bf4-5b38bd51decc|Succeeded|null|1498859407217|PC0FV9LS.pentaho.com|192.168.33.1|null|/public/Steel Wheels/Buyer Report (sparkline report).prpt

chiltonj:
@mzur Thats correct-- UID used a common key to correlate with other entries that may contain additional details.

chiltonj:
If there is a piece of info in that map that is guaranteed to be there for all WI then we could use that bit. Check w/ R1 since I believe they looked into that.

pentaho-mzur:
In fact - if there is a "lack of identification" of a work item executing, the proper solution to it would be to introduce one - this could and should probably taken up to the gui/fronted and then used to track the work item.

pentaho-amessier:
I don't believe there is a single piece of information do that identify a work item uniquely. The actions we currently support (SimpleReportingAction and PdiAction) have an inputFile, which can be used, but other actions? or non-actions, for which we will need to add support in the future, who knows? The action data is given to us in a map (params map).

pentaho-amessier:
This solution has been rejected, closing PR

PR #3693, opened 2017-07-05
[CLEANUP] fixed test error

mchen-len-son:


mchen-len-son:
@pamval @mbatchelor @rmansoor please take a look.

This is a fix for the failing test. For some reason this test was not executed in this wingman run. you can see the report with the error e.g. in #3690.

PR #3695, opened 2017-07-05
[BACKLOG-17396] Added a new method to IActionInvoker interface to allow

pdesai16:
passing the credential when invoking an action
- Added default implemention of this new method
- Added default implementation of existing invokeAction() to call the new
  method with null credential
- Added interface to provide user credential

### DIFF
--- a/api/src/main/java/org/pentaho/platform/api/action/IActionInvoker.java
+++ b/api/src/main/java/org/pentaho/platform/api/action/IActionInvoker.java
@@ -35,6 +35,25 @@ public interface IActionInvoker {
    * @return the {@link IActionInvokeStatus} object containing information about the action invocation
    * @throws Exception if the action cannot be run for some reason
    */
-  IActionInvokeStatus invokeAction( IAction action, final String user, final Map<String, Serializable> params )
-    throws Exception;
+  default IActionInvokeStatus invokeAction( IAction action, final String user, final Map<String, Serializable> params )
+    throws Exception {
+    return invokeAction( action, user, params, null );
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


pdesai16:
@pentaho/rogueone @pentaho/rebelalliance  - please review

api/src/main/java/org/pentaho/platform/api/action/IActionInvoker.java
```diff
@@ -35,6 +35,25 @@
...
    */
-  IActionInvokeStatus invokeAction( IAction action, final String user, final Map<String, Serializable> params )
-    throws Exception;
+  default IActionInvokeStatus invokeAction( IAction action, final String user, final Map<String, Serializable> params )
+    throws Exception {
+    return invokeAction( action, user, params, null );
+  }
+
+  /**
+   * Invokes the {@link IAction} {@code action} with teh credential. The default implementation is provided
```
pentaho-mzur: typo (teh)

api/src/main/java/org/pentaho/platform/api/action/IWorkerNodeCredential.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Copyright (c) 2017 Pentaho Corporation..  All rights reserved.
+ */
+
+package org.pentaho.platform.api.action;
+
+/**
+ * Interface that provide the access to user credentials
+ */
+public interface IWorkerNodeCredential {
```
pentaho-mzur: Is this a worker-node-specific interface as the name suggests? The javadoc line says 'user credentials' so why not use IUserCredentials?

PR #3696, opened 2017-07-05
[BACKLOG-17085] added unit tests and some refactoring to make the Act…

pentaho-esager:
…ionInvokerSystemListener more unit-testable


pedrofvteixeira:
PR is 👍  ( simply address the missing copyright header on `ActionInvokerListenerTest.java` ).

Was wondering: as we're already doing this work to create unit tests and refactoring the `ActionInvokerSystemListener` class to make it _unit-testable_, could we take this opportunity to replace the way we call for the `.invokeAction()`, so that we start calling for the interface instead of the `ActionResource` class ? 

In other words, start having 

```
IActionInvoker invoker = PentahoSystem.get( IActionInvoker.class , null );
invoker.invokeAction( actionBean, user, params );
```

instead of

```
ActionResource actionResource = getActionResource();
return actionResource.invokeAction( ActionUtil.INVOKER_SYNC_VALUE, actionId, actionClass, actionUser, actionParams );
```

If we feel this switch would best be handled on a separate, subsequent PR, I'm good with that as well.

pdesai16:
The wingman is reporting coverage drop on class that is not changed in this commit. Will merge

PR #3704, opened 2017-07-14
[BACKLOG-17598] Improved formatting of work item status timestamps, and other related cleanup and unit tests

pentaho-amessier:
- BACKLOG-17598: Timestamps in the work-item-status log file need to be formatted to be human readable
  -- removed the WorkItemLifecycleEvent.targetTimestamp variable - it is not needed, the current timestamp can be logged by the logging framework directly

- added WorkItemLifecyclePhase.REDIRECTED - although it may not be used in practice for now, it guards against the future and is added for the sake of completeness
- added missing work item status publishing calls to ActionInvokerSystemListener
- fixed minor issues (content and formatting) with the message logged by WorkItemLifecycleEventWriter and added more robust tests
- added check to the BlockingQuartzJob for LoggingJobExecutionException, to prevent the job from being re-submitted when there's no need
- changed failed work item status messages to use Exception.toString() instaed of Exception.getLocalizedMessage(), which isn't descriptive enough

@pentaho-esager please take a look at my changes to ActionInvokerSystemListener.java

related pentaho-platform-ee PR:
https://github.com/pentaho/pentaho-platform-ee/pull/1136

@pentaho/rebelalliance  @pentaho/rogueone


pentaho-amessier:
Sample work status log after these changes:
2017-07-14 15:58:20,418 eecdc70e-b5e4-418f-8898-47bae17b2df0|Submitted  |14 Jul 2017 15:58:20,418|aio-ps-db-wn|127.0.0.1
2017-07-14 15:58:20,690 eecdc70e-b5e4-418f-8898-47bae17b2df0|Dispatched |14 Jul 2017 15:58:20,690|aio-ps-db-wn|127.0.0.1|/public/Steel Wheels/Buyer Report (sparkline report).prpt[admin]
2017-07-14 15:58:28,785 eecdc70e-b5e4-418f-8898-47bae17b2df0|Received   |14 Jul 2017 15:58:28,000|6488cf979665|172.17.0.3|/public/Steel Wheels/Buyer Report (sparkline report).prpt[admin]
2017-07-14 15:58:28,857 eecdc70e-b5e4-418f-8898-47bae17b2df0|In progress|14 Jul 2017 15:58:28,000|6488cf979665|172.17.0.3|/public/Steel Wheels/Buyer Report (sparkline report).prpt[admin]
2017-07-14 15:58:37,380 eecdc70e-b5e4-418f-8898-47bae17b2df0|Succeeded  |14 Jul 2017 15:58:37,000|6488cf979665|172.17.0.3|/public/Steel Wheels/Buyer Report (sparkline report).prpt

core/src/test/java/org/pentaho/platform/workitem/WorkItemLifecyclePhaseTest.java
```diff
@@ -17,42 +17,64 @@
...
+    assertEquals( "The work item execution has been restarted",
       WorkItemLifecyclePhase.RESTARTED.getDescription() );
+    assertEquals( "The work item execution has been redirected",
+      WorkItemLifecyclePhase.REDIRECTED.getDescription() );
+  }
+
+  @Test
+  public void testGetMaxNameLength() {
+    // our longest name is that of IN_PROGRESS
+    assertEquals( WorkItemLifecyclePhase.IN_PROGRESS.getName().length(), WorkItemLifecyclePhase.getMaxNameLength() );
```
pdesai16: Will this code fails if more descriptive name gets added in future

core/src/main/resources/org/pentaho/platform/util/messages/messages.properties
```diff
@@ -71,4 +71,6 @@ LifecyclePhase.FAILED=Failed
 LifecyclePhase.FAILED_DESC=The work item execution has failed
 LifecyclePhase.RESTARTED=Restarted
 LifecyclePhase.RESTARTED_DESC=The work item execution has been restarted
+LifecyclePhase.REDIRECTED=Redirected
```
chiltonj: I don't think the phases should be localized.  It makes it much more difficult to grep and for tools like kibana / ES to perform any metrics around a particular state.

core/src/test/java/org/pentaho/platform/workitem/WorkItemLifecyclePhaseTest.java
pentaho-amessier: Yes, the test will fail, so it will need to be added, but the code, which pads names based on the max name length won't have to change

core/src/main/resources/org/pentaho/platform/util/messages/messages.properties
pentaho-amessier: Are you suggesting we simply use the all-caps values in the logs?

chiltonj: I think that would preferable to having them localized in the logs

pentaho-amessier: I don't feel strongly about this, so I can remove all the localization, if you think it's appropriate

chiltonj: Thanks -- It will be more tool friendly

core/src/main/java/org/pentaho/platform/workitem/WorkItemLifecyclePhase.java
```diff
@@ -17,72 +17,72 @@
...
-  }
-
-  public String getDescription() {
-    return this.description;
-  }
+  RESTARTED,
+  /**
+   * The work item execution has been restarted
+   */
+  REDIRECTED;
```
pdesai16: In what case would you use this state - Redirected

core/src/test/java/org/pentaho/platform/workitem/WorkItemLifecyclePhaseTest.java
pdesai16: This doesn't feel right. The existing test case should not break when we just added another state

core/src/main/java/org/pentaho/platform/workitem/WorkItemLifecyclePhase.java
```diff
@@ -17,72 +17,72 @@
...
-      .append( "description", this.getDescription() )
-      .toString();
+  /**
+   * A convenience method that pads the string representation of this {@link WorkItemLifecyclePhase} such that all
+   * padded names are of the same length. This allows us to line up values within logs and makes the logs slightly
+   * more readable.
+   *
+   * @return {@link String} containing a padded string representation of this {@link WorkItemLifecyclePhase}
+   */
+  public String getPaddedName() {
```
chiltonj: Not sure this belongs in this class -- seems specific to logging / how the value is to be rendered..

core/src/test/java/org/pentaho/platform/workitem/WorkItemLifecyclePhaseTest.java
pentaho-amessier: The existing test case works with existing code. Whenever we change code, we re-run tests to ensure that they pass. If a test checks the content of some string, for instance, and the content of that string changes, the test will fail and therefore needs to be updated. I'm not sure why this would be any different. Do you suggest that we remove the test altogether?

core/src/main/java/org/pentaho/platform/workitem/WorkItemLifecyclePhase.java
pentaho-amessier: I'll move it to the WorkItemLifecycleventFileWriter

core/src/main/java/org/pentaho/platform/workitem/WorkItemLifecyclePhase.java
pentaho-amessier: I added this for sake of completeness, we don't currently redirect...

core/src/test/java/org/pentaho/platform/workitem/WorkItemLifecyclePhaseTest.java
pentaho-amessier: At any rate, I moved some code around, and this test is no more :)

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -77,21 +79,24 @@ public boolean startup( IPentahoSession session ) {
...
         String encoded = IOUtils.toString( fileInputStream );
         fileInputStream.close();
-        Payload payload = new Payload( encoded, LocaleHelper.UTF_8 );
+        payload = new Payload( encoded, LocaleHelper.UTF_8 );
         logger.info( "Issuing Work Item request" );
         try {
           IActionInvokeStatus status = issueRequest( payload );
         } catch ( Exception e ) {
-          //send status to chronos
+          // TODO: send status to chronos
```
chiltonj: is this event worthy?

pentaho-amessier: Already taken care of indirectly: issueRequest > payload.issueRequest >  getActionInvoker().invokeAction( action, actionUser, actionMap ) - any failures would be reported inside invokeAction

scheduler/src/main/java/org/pentaho/platform/scheduler2/quartz/BlockingQuartzJob.java
```diff
@@ -50,6 +50,9 @@ public void execute( final JobExecutionContext jobExecutionContext ) throws JobE
             "Job '" + jobExecutionContext.getJobDetail().getName()
                 + "' attempted to run during a blockout period.  This job was not executed" );
       }
+    } catch ( ActionAdapterQuartzJob.LoggingJobExecutionException le ) {
+      // thrown by the execution code - if execution fails, there's nothing we can do, no point in trying to execute
```
chiltonj: Should a log entry be written out instead of silently failing?

pentaho-amessier: We already have log entries internal to the code that would cause the exceptions - this is purely to prevent a job retry

PR #3705, opened 2017-07-17
[PDI-14764] Jaro algorithm glitch

stanislau-strelchanka:


mchen-len-son:
com.wcohen.secondstring is now shipped with Pentaho server, since kettle-core is depending on it.
the obsolete secondstring can be removed.

PR #3706, opened 2017-07-17
[BACKLOG-17598 & BACKLOG-17599] Improved formatting of work item status timestamps, fixed work item status timestamps coming from worker nodes, and other related cleanup and unit tests

pentaho-amessier:
- BACKLOG-17598: Timestamps in the work-item-status log file need to be formatted to be human readable
-- removed the WorkItemLifecycleEvent.targetTimestamp variable - it is not needed, the current timestamp can be logged by the logging framework directly
- BACKLOG-17599: Incorrect time stamp in messages coming from worker node
- fixed minor issues (content and formatting) with the message logged by WorkItemLifecycleEventWriter and added more robust tests
- added check to the BlockingQuartzJob for LoggingJobExecutionException, to prevent the job from being re-submitted when there's no need
- changed failed work item status messages to use Exception.toString() instaed of Exception.getLocalizedMessage(), which isn't descriptive enough
- Per request, removed the unnecessary WorkItemLifecyclePhase localization

For reference, this is related to a PR that was accidentally closed: https://github.com/pentaho/pentaho-platform/pull/3704


pentaho-amessier:
All comments have been addressed and the changes have been dev tested

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -166,6 +182,7 @@ private void buildPayload( String urlEncodedJson, String enc ) {
         String actionClass = URLDecoder.decode( jsonVars.getString( ActionUtil.INVOKER_ACTIONCLASS ), enc );
         String actionId = URLDecoder.decode( jsonVars.getString( ActionUtil.INVOKER_ACTIONID ), enc );
         actionUser = URLDecoder.decode( jsonVars.getString( ActionUtil.INVOKER_ACTIONUSER ), enc );
+        workItemUid = URLDecoder.decode( jsonVars.getString( ActionUtil.WORK_ITEM_UID ), enc );
```
pentaho-esager: workItemUid needs to be included in the job passed from the main pentaho server and also in the scripts which create the JSON input, otherwise this payload will be looking for key (workItemUid) that doesn't exist.   Now that I'm thinking on it, perhaps we should create some object that allows us to put "extras" in the payload without modifying the scripts.

pentaho-amessier:
Temporarily closing to test chronos related changes that @pentaho-esager  pointed out

pentaho-amessier:
Temporarily closing to test chronos related changes that @pentaho-esager  pointed out

pentaho-amessier:
Validated the changes - PR is ready for merge

PR #3707, opened 2017-07-17
[BACKLOG-17100] Add needed support for spring to properly honor both s…

pentaho-mzur:
…ync and async events, and correctly honor Async annotation.

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/spring/ApplicationEventDispatcher.java
+++ b/extensions/src/main/java/org/pentaho/platform/spring/ApplicationEventDispatcher.java
@@ -28,7 +28,7 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 
 public class ApplicationEventDispatcher extends AbstractApplicationEventMulticaster {
-  public final static String ASYNC_ANNOTATION_QUALIFIER = "dispatcher";
+  public static final String ASYNC_ANNOTATION_QUALIFIER = "dispatcher";
 
   private ApplicationEventMulticaster asyncMulticaster;
   private ApplicationEventMulticaster defaultMulticaster;
@@ -67,7 +67,7 @@ public class ApplicationEventDispatcher extends AbstractApplicationEventMulticas
     asyncMulticaster.multicastEvent( event );
   }
 
-  public void multicastEvent(final ApplicationEvent event, final ResolvableType type ) {
+  public void multicastEvent( final ApplicationEvent event, final ResolvableType type ) {
     defaultMulticaster.multicastEvent( event, type );
     asyncMulticaster.multicastEvent( event, type );
   }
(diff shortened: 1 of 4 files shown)


extensions/src/test/java/org/pentaho/platform/spring/ApplicationEventDispatcherTest.java
```diff
@@ -0,0 +1,291 @@
...
+  }
+
+  public long getLastThreadId() {
+    return threadIds.get( threadIds.size() -1 );
+  }
+
+  @Override
+  public void onApplicationEvent( final ApplicationEvent event ) {
+    receivedEvents.add( (AppEvent) event );
+    if ( ! threadIds.contains( Thread.currentThread().getId() ) ) {
```
pentaho-amessier: I would make threadIds in all the listeners Sets and remove the contains(...) checks

extensions/src/test/java/org/pentaho/platform/spring/ApplicationEventDispatcherTest.java
```diff
@@ -0,0 +1,291 @@
...
+    this.id = id;
+  }
+
+  String getId() {
+    return id;
+  }
+}
+
+@Component
+class SyncAppEventListener implements ApplicationListener {
```
pentaho-amessier: Would it make sense to create a common parent class for all the listeners with all but the event handling methods defined there to remove all the redundancy?

extensions/src/test/java/org/pentaho/platform/spring/ApplicationEventDispatcherTest.java
```diff
@@ -0,0 +1,291 @@
...
+
+    // verify async listeners receive multiple events on separate threads
+    // engage all 5 threads
+    appConfig.publish( new AsyncAppEvent( "3" ) );
+    appConfig.publish( new AsyncAppEvent( "4" ) );
+    appConfig.publish( new AsyncAppEvent( "5" ) );
+    appConfig.publish( new AsyncAppEvent( "6" ) );
+    appConfig.publish( new AsyncAppEvent( "7" ) );
+
+    Thread.sleep( 3000 ); // allow events to drain
```
pentaho-amessier: Should we check here or below that synchListener's threadcount is still only 1?

chiltonj:
What is the scope of this change? does it only impact workitemevent code? Or any thing that was using the applicationEventMulticaster? I remember when I made that bean use Threadpool its impact was application wide.

extensions/src/main/java/org/pentaho/platform/spring/ApplicationEventDispatcher.java
```diff
@@ -0,0 +1,140 @@
...
+  public void addApplicationListenerBean( final String listenerBeanName ) {
+    final ApplicationListener listener = beanFactory.getBean( listenerBeanName, ApplicationListener.class );
+    if ( isAsync( listener ) ) {
+      asyncMulticaster.addApplicationListenerBean( listenerBeanName );
+    } else {
+      defaultMulticaster.addApplicationListenerBean( listenerBeanName );
+    }
+  }
+
+  private boolean isAsync( final ApplicationListener listener ) {
```
chiltonj: it looks to me Spring should be doing this -- the annotation processing.   Why is Spring not doing this?

pentaho-mzur:
The changes would only be applicable to application listeners that are decorated with this annotation:

@Async("dispatcher")

Any other main stream code is not affected...note the change to applicationContext-spring-security.xml.
The applicationEventMulticaster was changed to use a "default" multicaster that is wired to be the one used so far so that no other main stream code is affected.

extensions/src/main/java/org/pentaho/platform/spring/ApplicationEventDispatcher.java
pentaho-mzur: Well yes.
 
However the whole point was that  spring did not process the Async annotation properly (it simply did not work) - it used a multicaster class that could do either/or but not both type of dispatching. To achieve the ability to use both sync and async event dispatching this was one of the solutions suggested online.
(ApplicationEventMulticaster can be configured with either - sync/async task executor)

pentaho-mzur:
New commit for check-style issue fixes was added.

extensions/src/main/java/org/pentaho/platform/spring/ApplicationEventDispatcher.java
pentaho-mzur: Also, please note - in fact spring is still processing the annotations, the "only added thing" that the "disptacher" does, it selects a proper multicaster that will do the right thing - otherwise it is the multicaster that processes the annotation. So this dispacher also reflects uses the annotations to properly dispatch, and is meant to be a wrapper , and should be removed as soon as we have this capability working by spring itself...(??)

pentaho-mzur:

diogofscmariano:
@pentaho-mzur i think the commit messages are referencing a wrong BACKLOG

pentaho-mzur:
Closing for now after discussing with others - the concern is that this PR widens the scope of trying to "fix spring" in a way that could affect code that is out of the scope of our "worker nodes" impl and so introduces some risks.

PR #3710, opened 2017-07-20
Revert "[PDI-14764] Jaro algorithm glitch"

mchen-len-son:
Reverts pentaho/pentaho-platform#3705


mchen-len-son:
Devci failed to assembly Pentaho Platform after removing secondstring.
Reverted the changes, whilst I'm looking into it.

PR #3712, opened 2017-07-20
[BACKLOG-17675] audit entry for remote execution

chiltonj:

Adds audit entry for start / finish. Since the same id is used for the complete audit trail the duration per item is cumulative -- which is the accurate duration.

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/WorkerNodeActionInvokerAuditor.java
+++ b/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/WorkerNodeActionInvokerAuditor.java
@@ -1,0 +1,75 @@
+/*!
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License, version 2.1 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
+ * or from the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * Copyright (c) 2017 Pentaho Corporation..  All rights reserved.
+ */
+package org.pentaho.platform.web.http.api.resources;
+
+
+import org.apache.commons.lang.time.StopWatch;
+import org.pentaho.platform.api.action.IAction;
+import org.pentaho.platform.api.action.IActionInvokeStatus;
+import org.pentaho.platform.api.action.IActionInvoker;
+import org.pentaho.platform.engine.core.audit.AuditHelper;
+import org.pentaho.platform.engine.core.audit.MessageTypes;
+import org.pentaho.platform.engine.core.system.PentahoSessionHolder;
+import org.pentaho.platform.util.ActionUtil;
+
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+
+public class WorkerNodeActionInvokerAuditor implements IActionInvoker {
+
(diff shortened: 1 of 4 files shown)


pentaho-amessier:
Wouldn't it make more sense to put this code directly inside actionInvoker.invokeAction? Otherwise the effort will need to be duplicated in every "entry point" to the action invoking code, which at the moment is the ActionResource (for long running / marathon) and ActionIntokerSystemListener (for short running / chronos)

chiltonj:
@pentaho-amessier Does the actionInvoker.invokeAction have multiple implementations? If so are they   called out side of a WN context ? However, as you point out It may be necessary to add similar audit to chronos. But I am not going to push it down into common code that is executed by the server as well.

pentaho-amessier:
@chiltonj there are multiple implementations, but it all points down to a common implementation in one of the classes - currently all various implementations end up calling the same code, if I recall, although clearly we cannot rely on this always being the case. Given that this should only apply to WN, you are right not to push it to the common code. In that case, perhaps we can put the makeAuditRecord and any other pieces that make sense in a common location and add the logic to ActionInvokerSystemListener... Or if you think it makes sense, pernahs we can create a common entry point for both ActionResource and ActionInvokerSystemListener, so that the call this common point instead of the actionInvoker directly and then we could put the auditing code in one place. Whatever you think is best.

chiltonj:
@pentaho-amessier your point is valid -- I will add in a decorator like thing that can be used in WN context

chiltonj:
Changes made

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/WorkerNodeActionInvokerAuditor.java
```diff
@@ -0,0 +1,75 @@
...
+import org.pentaho.platform.engine.core.audit.AuditHelper;
+import org.pentaho.platform.engine.core.audit.MessageTypes;
+import org.pentaho.platform.engine.core.system.PentahoSessionHolder;
+import org.pentaho.platform.util.ActionUtil;
+
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+
+public class WorkerNodeActionInvokerAuditor implements IActionInvoker {
```
pentaho-amessier: A short description explaining the purpose of this class would be helpful

extensions/src/test/java/org/pentaho/platform/web/http/api/resources/ActionResourceTest.java
```diff
@@ -192,8 +192,30 @@ public void testRunnableRun() throws Exception {
...
+
+    final WorkerNodeActionInvokerAuditor wnActionInvokerAuditer = new WorkerNodeActionInvokerAuditor( actionInvoker );
+    WorkerNodeActionInvokerAuditor wnaiu = Mockito.spy( wnActionInvokerAuditer );
+    wnaiu.invokeAction( actionMock, actionUser, actionMapMock );
+
+    Mockito.verify( actionInvoker, Mockito.times( 1 ) ).invokeAction(  Mockito.eq( actionMock ), Mockito.eq( actionUser ),  Mockito.eq( actionMapMock ) );
+
+    Mockito.verify( wnaiu, Mockito.times( 2 ) ).makeAuditRecord( Mockito.anyLong() ,Mockito.anyString(), Mockito.anyMap() ,  Mockito.anyString() );
+
+  }}
```
pentaho-amessier: Is this closing bracket in the right place?

chiltonj: Done: fixed a bunch of checkstyle atrocities

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/WorkerNodeActionInvokerAuditor.java
```diff
@@ -0,0 +1,75 @@
...
+    try {
+      return actionInvoker.invokeAction( action, user, params );
+    } finally {
+      makeAuditRecord( stopWatch.getTime() / 1000, MessageTypes.INSTANCE_END, auditParams, action.getClass().getName() );
+    }
+  }
+
+
+  protected void makeAuditRecord( final float time, final String messageType,
+                                 final Map<String, Serializable> actionParams, String className ) {
```
pentaho-mzur: add 'final' to className?

PR #3726, opened 2017-07-31
Biserver 12187

benletchford:
We're getting `error.EncounteredWhileSavingTypeError: Cannot read property 'replace' of undefined` when saving in jPivot on Pentaho 7.1 even though it is saving successfully. 

replace_of_undefined

root cause: `frame.handle_puc_save` for jPivot isn't returning a value (https://github.com/pentaho/pentaho-platform-plugin-jpivot/blob/d61ca527b172fdf6885c672bf082eb2d50f259f4/src/org/pentaho/jpivot/Pivot_jsp.java#L1304). See below:

```
function handle_puc_save(path, name, overwrite) {
   controller.saveAs(name, name, "", path, overwrite);
}
```

To fix and to make sure it doesn't break anything all we have to do is check if result is set or not. As if `result` is `undefined` we don't want to operate on it anyway.

This fixes http://jira.pentaho.com/browse/BISERVER-12187

### DIFF
--- a/user-console/src/main/java/org/pentaho/mantle/client/commands/SaveCommand.java
+++ b/user-console/src/main/java/org/pentaho/mantle/client/commands/SaveCommand.java
@@ -285,7 +285,7 @@ public class SaveCommand extends AbstractCommand {
     } else if (frame.handle_puc_save) {
       try {
         var result = frame.handle_puc_save(path, filename, overwrite, errorCallback);
-        if (isSavedSuccessfully && result){
+        if (isSavedSuccessfully && result !== undefined){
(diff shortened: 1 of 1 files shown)


benletchford:
I've now patched jPivot to fix this problem, please see: https://github.com/pentaho/pentaho-platform-plugin-jpivot/pull/56

pamval:
does that mean we no longer need this change in platform?

benletchford:
Probably - I didn't close as the Jira issue also said Pivot4J was affected, however this change just silences the error and doesn't "fix" it per se. I'll go ahead and close.

PR #3727, opened 2017-07-31
[SP-3696] Backport of CDE-909 - It is not possible to save a CDE Dashboard with a query longer than 2000 characters (6.1 Suite)

Advent51:
- org.codehaus.jackson.jackson-core-asl and org.codehaus.jackson.jackson-mapper-asl versions were updated to 1.9.13
- added test


Advent51:
@stepanovdg @pamval please review

PR #3737, opened 2017-08-02
[BACKLOG-17323] worker nodes OSGi with all configurations/properties …

pedrofvteixeira:
…being read from pentaho.workernodes.cfg

From this point onward, all worker nodes properties exist in `pentaho.workernodes.cfg` making this feature fully configurable ( in a restartless fashion )

- promoted `isSupportedAction( IAction action )` to `IActionInvoker` interface
- The `interface IRemoteActionInvoker extends IActionInvoker` was moved from within worker nodes OSGi into pentaho-platform-api; this will allow for OSGi plugins to register themselves onto non-OSGi's ActionInvokerDelegator class without clashing with registered IActionInvoker ( non-OSGi's ActionInvokerDelegator now knows of IRemoteActionInvoker interface )
- Worker Nodes OSGi now holds a `class WorkerNodesDelegator implements IRemoteActionInvoker`, and is the only one that gets published into PentahoSystem
  - `WorkerNodesDelegator` is the one that delegates execution to `AdvancedActionInvoker / MesosChronosActionInvoker`, depending on what's the `orchestrator` value specified in `pentaho.workernodes.cfg`
- `pentaho.workernodes.cfg` now holds an `orchestrator` property

To be merged alongside https://github.com/pentaho/pentaho-platform-ee/pull/1158 and https://github.com/pentaho/pentaho-ee/pull/462

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/WorkerNodeActionInvokerAuditor.java
+++ b/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/WorkerNodeActionInvokerAuditor.java
@@ -70,4 +70,9 @@ public class WorkerNodeActionInvokerAuditor implements IActionInvoker {
   private String getValue( Map<String, Serializable> actionParams, String key ) {
     return actionParams.get( key ) != null ? actionParams.get( key ).toString() : "";
   }
+
+  @Override
+  public boolean isSupportedAction( IAction action ) {
+    return false; // not needed
+  }
 }
--- a/api/src/main/java/org/pentaho/platform/api/action/IActionInvoker.java
+++ b/api/src/main/java/org/pentaho/platform/api/action/IActionInvoker.java
@@ -37,4 +37,12 @@ public interface IActionInvoker {
    */
   IActionInvokeStatus invokeAction( IAction action, final String user, final Map<String, Serializable> params )
         throws Exception;
+
+  /**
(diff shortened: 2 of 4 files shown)


extensions/src/main/java/org/pentaho/platform/web/http/api/resources/WorkerNodeActionInvokerAuditor.java
```diff
@@ -70,4 +70,9 @@ protected void makeAuditRecord( final float time, final String messageType,
   private String getValue( Map<String, Serializable> actionParams, String key ) {
     return actionParams.get( key ) != null ? actionParams.get( key ).toString() : "";
   }
+
+  @Override
+  public boolean isSupportedAction( IAction action ) {
+    return false; // not needed
```
pentaho-mzur: Since the auditor is a decorator, that adds some capability to decorate an underlying invoker,  for "cleanliness" purposes, it should probably delegate to the action invoker it decorates.

pedrofvteixeira: Don't see why not. 

This can be done as a follow-up "cleanup" process. As it stands now, it is not a blocker. I'd request that someone with a better understanding of the intents of this Auditor decorator class can do such a change.

PR #3739, opened 2017-08-03
[BACKLOG-17796] WorkItemLifeCycle Event Logging occurs before Work Item Execution completes

pentaho-amessier:

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/plugin/action/ActionRunner.java
+++ b/extensions/src/main/java/org/pentaho/platform/plugin/action/ActionRunner.java
@@ -39,6 +39,8 @@ import org.pentaho.platform.scheduler2.quartz.SchedulerOutputPathResolver;
 import org.pentaho.platform.util.ActionUtil;
 import org.pentaho.platform.util.beans.ActionHarness;
 import org.pentaho.platform.util.messages.LocaleHelper;
+import org.pentaho.platform.workitem.WorkItemLifecyclePhase;
+import org.pentaho.platform.workitem.WorkItemLifecyclePublisher;
 
 import java.io.OutputStream;
 import java.io.Serializable;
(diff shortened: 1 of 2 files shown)


extensions/src/main/java/org/pentaho/platform/plugin/action/ActionRunner.java
```diff
@@ -68,14 +70,17 @@ public ActionRunner( final IAction actionBean, final String actionUser, final Ma
   }
 
   public Boolean call() throws ActionInvocationException {
+    final String workItemUid = ActionUtil.extractUid( params );
     try {
-      return callImpl();
+      final boolean result = callImpl();
+      WorkItemLifecyclePublisher.publish( workItemUid, params, WorkItemLifecyclePhase.SUCCEEDED );
```
pedrofvteixeira: One of the things that was mentioned when @pentaho-mzur and @chiltonj shared the LifecyclePhase events with the Rogue One team was that it could make sense to reword them, so that its nomenclature would match the nomenclature seen on Mesos UI. This would help a customer to more easily to viasually tie in the lifecycle phases, regardless of looking at Mesos UI , logs or Monitoring traps.

Mesos UI has FINISHED. WorkItemLifecyclePhase has SUCCEEDED. 

Is this rewording going to happen? Or has this been deemed not necessary? 
Roping in on @pentaho-mzur / @chiltonj to help me recall this.

pentaho-esager: @pedrofvteixeira I agree that this work is better for consistency/clarity sake, but I think it's not within the scope of this PR? Maybe we can create a jira for it?

pentaho-amessier: @pedrofvteixeira I was not aware of this conversation. If this is a requirement, we can certainly do the work, but I agree with@pentaho-esager that this is out of the scope of this PR

pedrofvteixeira: Agree with @pentaho-esager 's assessment. OOS for this story and therefore for this PR. No further questions here, PR looks good to me

PR #3740, opened 2017-08-07
[PPP-3735] - adding old commons-httpclient.3.0.1 jar

YuryBY:
- temporary fix while axis2 will not be updated


YuryBY:
@pamval, could you please take a look?

PR #3741, opened 2017-08-08
[BACKLOG-17894] Add work item status logs to ActionInvokerSystemListener and MesosChronosActionInvoker

pentaho-amessier:
Related PRs:
https://github.com/pentaho/pentaho-ee/pull/494
https://github.com/pentaho/pentaho-platform-ee/pull/1164


extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -130,6 +136,17 @@ public boolean startup( IPentahoSession session ) {
     return true;
   }
 
+  private void publishFailedWorkItemStatus( final Payload payload, final String failureMessage ) {
+    if ( payload != null ) {
+      payload.publishWorkItemStatus( WorkItemLifecyclePhase.FAILED, failureMessage );
+    } else {
+      // when payload is null, we have no way of getting the wok item id or any other work item related details, the
```
pentaho-mzur: wok - typo

pentaho-amessier: fixed

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -202,7 +220,12 @@ private void buildPayload( String urlEncodedJson, String enc ) {
     }
 
     IActionInvokeStatus issueRequest( ) throws Exception {
+      publishWorkItemStatus( WorkItemLifecyclePhase.RECEIVED, null );
```
pentaho-esager: Why does the "Received" status first get logged here?  It seems to me if there are failure paths which occur before this, received should occur before those.  Not a blocker to merging, just a question.

pentaho-amessier: We can't log the RECEIVED log until we have all the necessary pieces, and we don't have those pieces until the payload is built - I supposed we could move this call to the Payload constructor, it just seemed like this is a more appropriate place for this.

PR #3742, opened 2017-08-08
[BACKLOG-17773] XActions - Chart Component does not render image, it renders an html tag

Advent51:
- the place where encoding happens was changed


Advent51:
@mbatchelor @pamval @stepanovdg @kolinus please review

PR #3748, opened 2017-08-09
[BISERVER-13727] - Fixed issue with a bad error string causing a bad …

dkincade:
…error message while saving files in PUC


dkincade:
@mdamour1976 please review ...

mdamour1976:
@dkincade if you around today I can help you with git one more time

PR #3750, opened 2017-08-09
[PPP-3840] - Role enumeration should be allowed by admins only

mbatchelor:


mbatchelor:
PR has new test failure, closing and submitting new PR.

PR #3752, opened 2017-08-10
[PPP-3817] - Disable JavascriptRule by default

mbatchelor:


pamval:
pulled changes locally, built and ran tests - looks ok

PR #3755, opened 2017-08-14
[PDI-13572]-Now plugin classloader loads jars not only from /lib folder b…

ArtsiomYurhevich:
…ut also from root plugin directory. The reason is that some plugins have their assembly jars in plugin root dir.


ArtsiomYurhevich:
https://github.com/pentaho/pentaho-kettle/pull/4277

ArtsiomYurhevich:

PR #3756, opened 2017-08-14
[PDI-13572] -Multiple copies of pur-repository-plugin in DI Server pentaho-solutions/system folder structure

ArtsiomYurhevich:


ArtsiomYurhevich:

extensions/src/main/java/org/pentaho/platform/plugin/services/importer/PentahoPlatformImporter.java
```diff
@@ -183,6 +185,44 @@ public static String computeBundlePath( String bundlePath ) {
...
 
+  /**
+   * Performs one-way conversion on incoming String to produce a syntactically valid JCR path (section 4.6 Path Syntax).
+   */
+  public static String checkAndSanitize( final String in ) {
+    if ( in == null ) {
+      throw new IllegalArgumentException();
+    }
+    String extension = null;
+    if ( in.endsWith( RepositoryObjectType.CLUSTER_SCHEMA.getExtension() ) ) {
```
Kyrrui: This should probably be a switch case instead of an if / else stack. Then the default case can throw a illegalargumentexception.

repository/src/main/java/com/pentaho/pdi/ws/RepositorySyncWebService.java
```diff
@@ -0,0 +1,153 @@
...
+    portName = "repositorySyncPort", targetNamespace = "http://www.pentaho.org/ws/1.0" )
+public class RepositorySyncWebService implements IRepositorySyncWebService, Serializable {
+
+  private static final long serialVersionUID = 743647084187858081L; /* EESOURCE: UPDATE SERIALVERUID */
+
+  private static Log log = LogFactory.getLog( RepositorySyncWebService.class );
+  private static final String SINGLE_DI_SERVER_INSTANCE = "singleDiServerInstance";
+
+  public RepositorySyncStatus sync( String repositoryId, String repositoryUrl ) throws RepositorySyncException {
+    boolean singleDiServerInstance =
```
Kyrrui: Boolean.parseBoolean(STRING) would look better here.

repository/src/main/java/com/pentaho/pdi/ws/RepositorySyncWebService.java
```diff
@@ -0,0 +1,153 @@
...
+        "true".equals( PentahoSystem.getSystemSetting( SINGLE_DI_SERVER_INSTANCE, "true" ) ); //$NON-NLS-1$ //$NON-NLS-2$
+    if ( singleDiServerInstance ) {
```
Kyrrui: Since the 'singleDiServerInstance' var is only used once, these four lines can be compressed to one.
Like this: `if (Boolean.parseBoolean(PentahoSystem.getSystemSetting( SINGLE_DI_SERVER_INSTANCE, "true" )) {
return RepositorySyncStatus.SINGLE_DI_SERVER_INSTANCE;
}`

repository/src/main/java/com/pentaho/pdi/ws/RepositorySyncWebService.java
```diff
@@ -0,0 +1,153 @@
...
+      Element node;
+      try {
+        node =
+          XMLParserFactoryProducer.createSecureDocBuilderFactory().newDocumentBuilder().parse( new StringBufferInputStream( xml ) )
+                .getDocumentElement();
+      } catch ( Exception e ) {
+        node = null;
+      }
+      if ( node != null ) {
+        NodeList list = node.getElementsByTagName( "repository_location_url" ); //$NON-NLS-1$
```
Kyrrui: Minor, I would consider moving this string to a static var at top of file.

repository/src/test/java/com/pentaho/pdi/ws/RepositorySyncWebServiceClientIT.java
```diff
@@ -0,0 +1,54 @@
...
+import java.net.URL;
+
+@SuppressWarnings( "nls" )
+public class RepositorySyncWebServiceClientIT extends RepositorySyncWebServiceTest implements Serializable {
+
+  private static final long serialVersionUID = -6806897012063786589L; /* EESOURCE: UPDATE SERIALVERUID */
+
+  @Before
+  public void before() {
+    System.out.println( "Starting server..." );
```
Kyrrui: Should this also be sent to a log file?

repository/src/test/java/com/pentaho/pdi/ws/RepositorySyncWebServiceTest.java
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  public class TestPropertyHandler implements PropertyHandler {
+    public boolean exists( String arg0 ) {
+      throw new RuntimeException( "Not Expecting call to exists(" + arg0 + ")" );
+    }
+
+    public String getProperty( String arg0 ) {
+      return handler.getProperty( arg0 );
+      // throw new RuntimeException("Not Expecting call to getProperty("+arg0+")");
```
Kyrrui: Why is this commented out line included?

repository/src/test/java/com/pentaho/pdi/ws/RepositorySyncWebServiceTest.java
```diff
@@ -0,0 +1,314 @@
...
+        return null;
+      }
+
+      public void resetSettingsCache() {
+      }
+
+    } );
+
+    // now test the webservice
+    IRepositorySyncWebService webservice = getRepositorySyncWebService();
```
Kyrrui: Seems like this should be a final var

extensions/src/main/java/org/pentaho/platform/plugin/services/importer/PDIImportFileHandler.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @Override
+  protected RepositoryFile updateFile( final RepositoryFileImportBundle bundle, final RepositoryFile file,
+                                       final IRepositoryFileData data ) throws PlatformImportException {
+    RepositoryFile updatedFile = null;
+    if ( isNodeRepositoryFileData( file ) ) {
+      updatedFile = getRepository().updateFile( file, data, bundle.getComment() );
+    } else {
+      String fileName = bundle.getName();
+      getLogger().trace( "The file [" + fileName + "] will be recreated because it content-type was changed." );
```
Kyrrui: very minor, change 'it' to 'its'

repository/src/main/java/com/pentaho/pdi/ws/RepositorySyncWebService.java
Kyrrui: 2 lines that is

stepanovdg:
@Kyrrui Currently this pull-request doesn`t introduce new code. We are using already existing code from another project. Just to remove circular dependency - when pur plugin depends on platfrom and platform depends on pur-plugin. And after we require to have duplicate jars in many places - it was decided to split pur-plugin and move api and common impl into platform. 
So as it is a big change we prefer to not do any other refactor changes on the fly - to separate logic. 
We can make refactor jira issue after (just in case refactor can introduce other problems).

Kyrrui:
@ArtsiomYurhevich All right looks good to me then

PR #3757, opened 2017-08-15
[PDI-16369]-Opening a transformation or job uploaded to the repository using REST API results in "no transformer found" exception

denisprotopopov:
- added fix for transformations and jobs

### DIFF
--- a/repository/src/main/java/org/pentaho/platform/repository2/unified/fileio/RepositoryFileOutputStream.java
+++ b/repository/src/main/java/org/pentaho/platform/repository2/unified/fileio/RepositoryFileOutputStream.java
@@ -13,12 +13,14 @@
  * See the GNU General Public License for more details.
  *
  *
- * Copyright 2006 - 2013 Pentaho Corporation.  All rights reserved.
+ * Copyright 2006 - 2017 Pentaho Corporation.  All rights reserved.
  */
 
 package org.pentaho.platform.repository2.unified.fileio;
 
 import org.apache.commons.lang.StringUtils;
+import org.pentaho.platform.api.repository2.unified.Converter;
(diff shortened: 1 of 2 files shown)


denisprotopopov:
@mbatchelor please review the changes

denisprotopopov:
@pamval @mbatchelor please review the changes

repository/src/main/java/org/pentaho/platform/repository2/unified/fileio/RepositoryFileOutputStream.java
```diff
@@ -179,7 +181,18 @@ public void flush() throws IOException {
     // RepositoryFileWriter
     // but I couldn't figure out a clean way to do that. For now, charsetName is passed in here and we use it if
     // available.
-    final SimpleRepositoryFileData payload = new SimpleRepositoryFileData( bis, charsetName, mimeType );
+    final IRepositoryFileData payload;
+    Converter converter;
+
+    if ( "ktr".equalsIgnoreCase( ext ) ) {
+      converter = PentahoSystem.get( Converter.class, null, Collections.singletonMap( "name", "PDITransformationStreamConverter" ) );
```
pamval: I think this code will break in CE. Can you be a bit more defensive and test if converter is null for both ktr and kjb?

If it is null, then just default to new SimpleRepositoryFileData as before

pamval: Also, a unit or integration test would be nice. Not sure if it is possible, but would be nice.

denisprotopopov: @pamval please review the changes

PR #3761, opened 2017-08-17
[BISERVER-13723]-XMI models loaded through default-content can be dup…

ArtsiomYurhevich:
…licated when re-importing or re-publishing back to the Server


ArtsiomYurhevich:
@pamval , @mbatchelor Could you take a look please

PR #3763, opened 2017-08-17
[BACKLOG-18166] Intermediary Change to run work items in a background…

pentaho-esager:
… thread which waits for the pentaho server to be completely available before launching a work item request

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
+++ b/extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
@@ -21,8 +21,14 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.http.HttpResponse;
+import org.apache.http.HttpStatus;
+import org.apache.http.client.HttpClient;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.utils.URIBuilder;
 import org.json.JSONException;
 import org.json.JSONObject;
+import org.pentaho.di.core.util.HttpClientManager;
 import org.pentaho.platform.api.action.ActionInvocationException;
 import org.pentaho.platform.api.action.IAction;
 import org.pentaho.platform.api.action.IActionInvokeStatus;
@@ -43,6 +49,8 @@ import java.io.FileFilter;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.Serializable;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URLDecoder;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


pentaho-esager:
@pentaho/rogueone

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -77,14 +90,76 @@ private void placeBreadcrumbFile( String folderPath, boolean success ) {
...
+    int timeout = 600000;
+    final int sleep_time = 5000;
+    boolean isAlive = false;
+    while ( !isAlive ) {
+      int statusCode = pingIsAlive();
+      if ( statusCode == HttpStatus.SC_OK ) {
+        isAlive = true;
+      } else {
+        if ( timeout == 0 ) {
+          placeBreadcrumbFile( getSolutionPath(), false );
```
neeta-rathi: So we don't exit the while loop, but write the breadcrumb instead and that eventually leads to exiting the container, correct?

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -77,14 +90,76 @@ private void placeBreadcrumbFile( String folderPath, boolean success ) {
...
+      final HttpResponse response = client.execute( pingIsAlive );
+      return response.getStatusLine().getStatusCode();
+    } catch ( URISyntaxException | IOException e ) {
+      logger.error( e.getMessage() );
+    }
+    return HttpStatus.SC_INTERNAL_SERVER_ERROR;
+  }
+
+  private void waitUntilServerIsAlive( ) {
+    int timeout = 600000;
```
pdesai16: same as above. Adding a debug statement would help in troubleshooting later

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
pentaho-esager: Yes, that was my intent, but we can also break if we want.

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
```diff
@@ -77,14 +90,76 @@ private void placeBreadcrumbFile( String folderPath, boolean success ) {
       logger.error( Messages.getInstance().getErrorString( "ActionInvokerSystemListener.ERROR_0004_COULD_NOT_WRITE_STATUS" ) );
     }
   }
+
+  private int pingIsAlive( ) {
+    try {
```
pentaho-esager: Sure

extensions/src/main/java/org/pentaho/platform/plugin/action/ActionInvokerSystemListener.java
neeta-rathi: This is fine.

pentaho-esager:
Updated with requested changes

PR #3765, opened 2017-08-18
[BACKLOG-17872] - Fixed connection factory and updated tags.

wseyler:


wseyler:
@pentaho/2-1b Was able to build and test locally with postgres and mysql

PR #3775, opened 2017-08-25
[BISERVER-13603] - IE - PDF pane remains active when glass pane was a…

YuryBY:
…ctivated, which blocks user from accepting/declining action in appeared dialog box.

 - additional fix covers the regression

@pamval, could you please take a look?

### DIFF
--- a/user-console/src/main/java/org/pentaho/mantle/client/ui/tabs/MantleTab.java
+++ b/user-console/src/main/java/org/pentaho/mantle/client/ui/tabs/MantleTab.java
@@ -19,16 +19,6 @@ package org.pentaho.mantle.client.ui.tabs;
 
 import com.google.gwt.dom.client.Node;
 import com.google.gwt.dom.client.Style;
-import com.google.gwt.user.client.Element;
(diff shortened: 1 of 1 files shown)


YuryBY:
@pamval, the PR should be merged together with https://github.com/pentaho/pentaho-commons-gwt-modules/pull/595. Thanks

YuryBY:
@pamval, @mbatchelor the PR is blocker for SP cases. Thanks.

mchen-len-son:
@YuryBY shouldn't we remove the old changes at https://github.com/pentaho/pentaho-platform/pull/3775/files#diff-11e3f51c8dd386edc6276299ade02d5eR295 ?

YuryBY:
@mchen-len-son, nope. This changes is separate fix for the dropdown menu.

mchen-len-son:
@Yury got it, thanks.
for those who is also wondering how it works.
MantleDialogBox is a more accurate change than in pentaho/pentaho-commons-gwt-modules#595. 
It lets us use the new code when a report is requested for UI, and bypass it when it is being generated in background, therefore fixing the regression found.

PR #3780, opened 2017-08-27
[SP-3747][PPP-3840] - Role enumeration should be allowed by admins only

mchen-len-son:


mchen-len-son:

This is a backport of #3753.

mchen-len-son:
@stepanovdg please take a look.

PR #3781, opened 2017-08-27
[SP-3719][BISERVER-13725] - Blockout add/update should require admin

mchen-len-son:


mchen-len-son:

This is a backport of #3734.

PR #3783, opened 2017-08-28
[BACKLOG-18330] - Added MDXFlushCache component in lieu of using java…

wseyler:
…script to flush the mondrian cache.

### DIFF
--- a/core/src/main/resources/org/pentaho/platform/engine/services/runtime/plugins.properties
+++ b/core/src/main/resources/org/pentaho/platform/engine/services/runtime/plugins.properties
@@ -4,6 +4,7 @@
 # a deprecation warning to the Pentaho console.
 #
 # Map the Short Names
+MondrianFlushCacheComponent = org.pentaho.platform.plugin.action.mondrian.MondrianFlushCacheComponent
 ContentOutputComponent = org.pentaho.platform.plugin.action.deprecated.ContentOutputComponent
 ContentRepositoryCleaner = org.pentaho.platform.plugin.action.builtin.ContentRepositoryCleaner
 HelloWorldComponent = org.pentaho.platform.plugin.action.examples.HelloWorldComponent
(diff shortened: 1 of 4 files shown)


extensions/src/main/java/org/pentaho/platform/plugin/action/mondrian/MondrianFlushCacheComponent.java
```diff
@@ -0,0 +1,76 @@
...
+    for ( Cube cube : cubes ) {
+      cacheControl.flush( cacheControl.createMeasuresRegion( cube ) );
+    }
+    cacheControl.flushSchema( conn.getSchema() );
+    return true;
+  }
+
+  @Override
+  protected boolean validateAction() {
+    return true;
```
mbatchelor: You need to verify the conditions you expect here. For example, what if the MDX Lookup Rule doesn't have a connection ? Then your code on line 39 will get an NPE, won't it ? So minimally, you need to make sure that there's a "shared_olap_connection" available for you to get in this method.

core/src/main/resources/org/pentaho/platform/engine/services/runtime/plugins.properties
```diff
@@ -4,6 +4,7 @@
 # a deprecation warning to the Pentaho console.
 #
 # Map the Short Names
+MondrianFlushCacheComponent = org.pentaho.platform.plugin.action.mondrian.MondrianFlushCacheComponent
```
mbatchelor: Because of PPP-3817, the plugins.properties file is also in the WEB-INF/classes folder here: https://github.com/pentaho/pentaho-platform/blob/master/assemblies/pentaho-war/src/main/webapp/WEB-INF/classes/org/pentaho/platform/engine/services/runtime/plugins.properties 

You'll need to update that file in this PR as well.

PR #3786, opened 2017-08-28
[CLEANUP] Clarify readme when it comes to unit test and ITs. Fix form…

emartin-pentaho:
…atting.


emartin-pentaho:
Marc wished for the README to be clearer when it came to unit tests and integration tests. I also fixed the line spacing because it was causing problems with markdown formatting.

PR #3795, opened 2017-08-29
[SP-3717][BISERVER-10319] adapted for using from other projects

stepanovdg:


stepanovdg:
@mchen-len-son  @pamval  Could you please take a look?

PR #3806, opened 2017-09-06
[PDI-16369] Opening a transformation or job uploaded to the repositor…

mchen-len-son:
…y using REST API results in "no transformer found" exception

 - file title should be set for only kettle file (.ktr and .job)


mchen-len-son:
@mbatchelor @pamval @rmansoor please take a look.
This is a fix for regression found.

mchen-len-son:
@stepanovdg can you please merge this one?
QA have tested this both for the regression found and the original case.

PR #3807, opened 2017-09-06
[BISERVER-13664]-Mail Server settings configuration sends email password

PashkovichAndrei:
in clear text


PashkovichAndrei:
@mbatchelor, @smaring could you please have a look?
Did I cover all cases of mail's password encryption at this PR?

PashkovichAndrei:
@pamval, org.apache.commons.codec.binary.Base64 couldn't be used because of it's not GWT compatible. 

As for Base64Service, do you mean org.pentaho.platform.util.Base64PasswordService?

pamval:
yeah, that one

PashkovichAndrei:
@pamval org.pentaho.platform.util.Base64PasswordService depends on org.apache.commons.codec.binary.Base64. So that class we couldn't use as well.

I found new solution without adding new class, but I need to check that it works as expected.

PashkovichAndrei:
@pamval could you please do review?

PR #3809, opened 2017-09-07
[SP-3750][PDI-16369] Opening a transformation or job uploaded to the …

mchen-len-son:
…repository using REST API results in "no transformer found" exception

 - file title should be set for only kettle file (.ktr and .job)


mchen-len-son:
@stepanovdg please have a look.

PR #3812, opened 2017-09-08
[BACKLOG-18798] De-hydrating the pentaho-server of worker-nodes config

pentaho-amessier:
In order to remove the need for wiring the IActionInvoker bean within the pentahoObjects.spring.xml file, we need to have the ability to instantiate the detault action invoker directly within the ActionAdapterQuartzJob class. With the previous code base, we did not have this ability, because the DefaultActionInvoker class resided within the extensions module, which is downstream from the scheduler module (scheduler) where the class needed to be referenced. Therefore some refactoring/moving of the default action invoker and other related classes needed to happen.

As a result, the concept of a 'LocalActionInvoker' was introduced. The DefaultActionInvoker class still exists, but has been moved from extensions to scheduler module, so that it can be accessed directly from within ActionAdapterQuartzJob. It has been modified slightly to fetch the instance of streamProviver directly from the params map via the 'QuartzScheduler.RESERVEDMAPKEY_STREAMPROVIDER' key (this is how ActionAdapterQuartzJob fetched the streamProvider prior to the action-invocation refactoring). LocalActionInvoker within extensions (formerly known as DefaultActionInvoker) extends from the new DefaultActionInvoker within scheduler and overrides the getStreamProvider method to fetch it in a more generic fashion that allows us to serialize the parameters and invoke the action remotely. All worker node enabled code (spring wiring, ActionInvokerDelgator) should now directly access the LocalActionInvoker in extensions, the DefaultActionInvoker in scheduling should only be used from within ActionAdapterQuartzJob

Related PRs:
https://github.com/pentaho/pentaho-platform-ee/pull/1191
https://github.com/pentaho/pentaho-ee/pull/578
https://github.com/pentaho/pentaho-karaf-ee-assembly/pull/152


core/src/main/java/org/pentaho/platform/util/ActionUtil.java
```diff
@@ -89,6 +90,50 @@
...
+  public static void removeKeyFromMap( final Map<String, ?> params, final String key ) {
+    if ( params == null || key == null ) {
+      logger.debug( "Map or key are null, cannot remove." );
+      return;
+    }
+    // remove the key itself
+    params.remove( key );
+
+    // also remove the keys equivalent in the KEY_MAP, whether its a key or a value itself
+    if ( KEY_MAP.keySet().contains( key ) ) {
```
pentaho-mzur: Any reason for prefering KEY_MAP.keySet().contains over KEY_MAP.containsKey()?

core/src/main/java/org/pentaho/platform/util/ActionUtil.java
```diff
@@ -89,6 +90,50 @@
...
+      params.remove( KEY_MAP.get( key ) );
+    } else if ( KEY_MAP.values().contains( key ) ) {
```
pentaho-mzur: Isn't this if statement wasteful? We are just repeating the same thing twice?

core/src/test/java/org/pentaho/platform/action/ActionInvokeStatusTest.java
```diff
@@ -1,40 +1,40 @@
...
-  }
-
-  @Test
-  public void setAndGetThrowableTest() throws Exception {
-    ActionInvokeStatus actionInvokeStatus = new ActionInvokeStatus();
-    actionInvokeStatus.setThrowable( new Throwable( "test_message" ) );
-    Assert.assertEquals( actionInvokeStatus.getThrowable().getMessage(), "test_message" );
-  }
-}
+/*!
```
pentaho-mzur: Has anything changed in this class?

scheduler/src/main/java/org/pentaho/platform/scheduler2/action/DefaultActionInvoker.java
```diff
@@ -108,25 +72,21 @@ protected IBackgroundExecutionStreamProvider getStreamProvider( final Map<String
...
    * @param actionBean the {@link IAction} being invoked
    * @param actionUser The user invoking the {@link IAction}
    * @param params     the {@link Map} or parameters needed to invoke the {@link IAction}
    * @return the {@link IActionInvokeStatus} object containing information about the action invocation
    * @throws Exception when the {@code IAction} cannot be invoked for some reason.
    */
-  protected IActionInvokeStatus invokeActionImpl( final IAction actionBean,
-                                                  final String actionUser,
-                                                  final Map<String, Serializable> params ) throws Exception {
+  protected final IActionInvokeStatus invokeActionImpl( final IAction actionBean,
```
pentaho-mzur: What is the reasoning behind making this method final?

core/src/main/java/org/pentaho/platform/util/ActionUtil.java
pentaho-amessier: no reason, I can change it to containsKey

core/src/main/java/org/pentaho/platform/util/ActionUtil.java
pentaho-amessier: no - first we are checking if 'key' is a key with KEY_MAP, then we are checking if it's a value within KEY_MAP. Maybe I'm not understanding the question, we can discuss offline

core/src/test/java/org/pentaho/platform/action/ActionInvokeStatusTest.java
pentaho-amessier: No - the content of the class is the same, only the location has changed

scheduler/src/main/java/org/pentaho/platform/scheduler2/action/DefaultActionInvoker.java
pentaho-amessier: It's just my opinion that it should not be overridden.

core/src/main/java/org/pentaho/platform/util/ActionUtil.java
pentaho-mzur: What I meant was:  values().contains() effectively goes over the map. So in case of the key existing we go over the map twice. If it does not exist we effectively go over the map once anyway through the values().contains() ... theoreticaqlly we can just iterate.... but this is minor I guess, ...

scheduler/src/main/java/org/pentaho/platform/scheduler2/action/DefaultActionInvoker.java
pentaho-mzur: But what is behind this opinion?

core/src/main/java/org/pentaho/platform/util/ActionUtil.java
pentaho-amessier: So as an optimization, we could iterate over entries and check the key and value of each entry for a match - is this what you have in mind?

scheduler/src/main/java/org/pentaho/platform/scheduler2/action/DefaultActionInvoker.java
pentaho-amessier: Gut feeling :) I can remove the keyword, if you feel strongly.

pentaho-mzur: Oh man, now I am in a dilema. On one hand I like listening to gut feelings. One the other hand reasoning says not to put a final ;-) Leave it up to you to decide...

core/src/main/java/org/pentaho/platform/util/ActionUtil.java
pentaho-mzur: Yes. But again, this is minor, please decide.

scheduler/src/main/java/org/pentaho/platform/scheduler2/action/DefaultActionInvoker.java
pentaho-amessier: Basically - the public aip is invokeAction, and you can override it to your heart's desire, there's really no reason to override invokeActionImpl, which is an internal implementation, so my feeling is that if you are going to use this method, you should use it as-is. But again, just a gut feeling :)

PR #3813, opened 2017-09-10
[BACKLOG-18473] Bumping the karaf jaas version up to 4.0.7

pentaho-mzur:
The recent upgrade from 2.0.4 to 3.0.3 did not do the trick and we are still experiencing the same issue.
Tried to go to 4.0.7 with promising results (do not see the problem any more on my dev env). Upgrading to this version to allow more comprehensive  QA testing (and increase visibility to identify any issues) to be done.


pentaho-mzur:
Adding @YuryBY (who for some reason cannot be added as a reviewer directly)

PR #3814, opened 2017-09-12
[BACKLOG-18628] - DEV: Update CE specific screens

plagoa:


plagoa:
@joaormf I don't think that at this point that's possible but I can change to adapt to that possibility.

PR #3818, opened 2017-09-14
[BACKLOG-18838] Make ruby the default theme on PUC

joaormf:
@pentaho/kashyyyk


joaormf:
@diogofscmariano and @pedrofvteixeira this is the ce PR for the new theme

PR #3823, opened 2017-09-15
[Backlog-17630] Decoupling of Add Sequence step + [BACKLOG-18430] Separate already decoupled PDI steps and job entries into discrete plugins 

cardosov:
@pentaho/2-1b depends on pentaho/pentaho-kettle#4397


cardosov:
New approach on decoupling is to be attempt

PR #3827, opened 2017-09-19
[PDI-14357] Need to prohibit empty names when adding new users or roles via REST

AliaksandrShuhayeu:
[PDI-14357] Need to prohibit empty names when adding new users or roles via REST

- added  clearing stack trace for AlreadyExistsException for cases create an user and a role
- added a correct message


AliaksandrShuhayeu:
@mbatchelor Could you please merge it.
It has strange checkstyle violations warnings that are not connected with my changes. 
For example  
')' is preceded with whitespace. at Line: 295 Column: 61
')' is preceded with whitespace. at Line: 295 Column: 59
&& ( policy.isAllowed( AdministerSecurityAction.NAME ) );

PR #3834, opened 2017-09-21
[CDA-215] Users without 'Create Content' permission is able to edit CDA files

Advent51:
- already implemented method was added into the interface


Advent51:
@pamval @mchen-len-son @mbatchelor please review

PR #3842, opened 2017-09-26
[BACKLOG-18999] Now using the default jaxrs, message body provider

pentaho-mzur:


pentaho-amessier:
Unused import has been removed

PR #3846, opened 2017-09-27
[BACKLOG-18438] add scheduledFireTime param before invokeAction call to avoid NPE in BlockoutAction

neeta-rathi:


pentaho-amessier:
@neeta-rathi, that's correct. I cannon speak to whether this will fix the issue (I trust you :)), but this this value indeed was added earlier on, before the Action execution in original code - this was a mistake on my part.

neeta-rathi:
@pentaho-amessier pentaho-server-ee-8.0-QAT-294 runs into a NPE during BlockoutAction execution. With this code change, I don't run into that exception. Also, I don't see the errors in the log ( as attached on this BACKLOG ). Can you think of any other scenarios I should test out to ensure that the proposed change will fix the issue?

pentaho-amessier:
@neeta-rathi I haven't looked into the issue, so can't comment - if you'd like me to look into this, let me know.

PR #3848, opened 2017-09-28
[SP-3796][PDI-14357] Need to prohibit empty names when adding new users or roles via REST

AliaksandrShuhayeu:
[SP-3796][PDI-14357] Need to prohibit empty names when adding new users or roles via REST

- added  clearing stack trace for AlreadyExistsException for cases create an user and a role
- added a correct message


AliaksandrShuhayeu:

PR #3861, opened 2017-10-02
[SP-3796][PDI-14357] Need to prohibit empty names when adding new users or roles via REST

AliaksandrShuhayeu:
[SP-3796][PDI-14357] Need to prohibit empty names when adding new users or roles via REST

- added checking not blank for username and role name


AliaksandrShuhayeu:

PR #3852, opened 2017-09-29
[SP-3767] Backport of CDE-912 - Uncaught Error: Load timeout for modules when embedding dashboard within Pentaho Server (7.1 Suite)

Advent51:
- partial backport of #3800

PR #3864, opened 2017-10-03
Revert "[SP-3767] Backport of CDE-912 - Uncaught Error: Load timeout for modules when embedding dashboard within Pentaho Server (7.1 Suite)"

Advent51:
Reverts pentaho/pentaho-platform#3852


Advent51:
@mchen-len-son

PR #3871, opened 2017-10-11
[BISERVER-13478]-Error updating ACL when importing into repository under JDBC authentication

stepanovdg:


stepanovdg:
@mdamour1976 @rmansoor @pedrofvteixeira Could you please take a look?

PR #3878, opened 2017-10-13
[BACKLOG-19316] Updated String util with functions to trim whitespace from lists, strings and maps. For use in workernodes

pentaho-esager:

### DIFF
--- a/core/src/main/java/org/pentaho/platform/util/StringUtil.java
+++ b/core/src/main/java/org/pentaho/platform/util/StringUtil.java
@@ -18,6 +18,7 @@
 package org.pentaho.platform.util;
 
 import org.apache.commons.collections.MapUtils;
+import org.apache.commons.lang.StringUtils;
 
 import java.io.ByteArrayOutputStream;
 import java.io.PrintStream;
@@ -27,6 +28,7 @@ import java.util.Map;
 import java.util.StringTokenizer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 public class StringUtil {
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/pentaho/platform/util/StringUtil.java
```diff
@@ -116,4 +118,31 @@ public static String getMapAsPrettyString( final Map map ) {
     MapUtils.debugPrint( new PrintStream( writtenBytes ), "Map", map );
     return writtenBytes.toString();
   }
+
+  /**
+   * Returns a map of String key-value pairs, in which each key and corresponding value has been trimmed of whitespace.
+   * @param map the {@link Map} of String key-value pairs to be trimmed
+   * @return a trimmed String {@link Map}
+   */
+  public static Map<String, String> trimStringMap( Map<String, String> map ) {
```
pentaho-amessier: trim might be a better name, we know we're trimming a map because of the param type - but that's just my opinion, no need to take action if you disagree

core/src/main/java/org/pentaho/platform/util/StringUtil.java
```diff
@@ -116,4 +118,31 @@ public static String getMapAsPrettyString( final Map map ) {
...
+    return list == null ? list : list.stream().map( StringUtils::trimToEmpty ).collect( Collectors.toList() );
+  }
+
+  /**
+   * Wrapper method around StringUtils.trim
+   * @param string the {@link String} to be trimmed
+   * @return the trimmed {@link String}
+   */
+  public static String trim( String string ) {
+    return StringUtils.trimToEmpty( string );
```
pentaho-amessier: I would use trim here, it's a bit misleading to have a class called StringUtil (very close to StringUtils) and have its trim method some something other than apache StringUtils' trim method. If we want the trimToEmpty functionality in pentaho's StringUtil, we could add a method with the same name.

core/src/test/java/org/pentaho/platform/util/StringUtilTest.java
```diff
@@ -30,6 +31,57 @@
...
+
+    Map<String, String> nullMap = null;
+    Assert.assertNull( StringUtil.trimStringMap( nullMap ) );
+
+    Map<String, String> stringMap = new HashMap<>();
+    stringMap.put( " spaces ", " spacesval " );
+    stringMap.put( "\t tabs \t", "\t tabsval \t" );
+    stringMap.put( "\n newline \n", "\n newlineval \n" );
+    stringMap.put( "empty", "" );
+    stringMap.put( "null", null );
```
pentaho-amessier: I would add an entry where the key is actually null as well

PR #3879, opened 2017-10-13
Encrypt LDAP password with stronger cipher

gwu70ca:
Use random generated password/salt to encrypt LDAP password with stronger cipher. Password/salt is saved into property file in user's home directory.


mbatchelor:
@gwu70ca - looks like you only contributed the test case, not the service.

gwu70ca:
This is strange. My history showed there were two files pushed. Let me
check.

gwu70ca:
My mistake, was in local master, not in branch. Just pushed

PR #3883, opened 2017-10-17
[ANALYZER-3628] - Currently there is no record of errors when SCHEDUL…

YuryBY:
…ING analyzer report in PUC after schema changes


YuryBY:
@tmcsantos, @pamval, @mbatchelor, could you please take a look?

PR #3915, opened 2017-10-27
[SP-3844][BISERVER-13478]-Error updating ACL when importing into repository under JDBC authentication

stepanovdg:


stepanovdg:

mchen-len-son:
There is no decision on copyright headers for 7.1 yet.
Besides that everything looks ok, double-checked locally.
Merging.

PR #3918, opened 2017-10-27
[BACKLOG-19566] Adding a method that checks for the existence of the jcr access control privilege

pentaho-amessier:
Related PRs:
https://github.com/pentaho/pentaho-ee/pull/817
https://github.com/pentaho/pentaho-platform-ee/pull/1223

Note that the method is being only added here, but not used within the platform code, but rather used by a sub-class in the pentaho-ee repo.

### DIFF
--- a/repository/src/main/java/org/apache/jackrabbit/core/security/authorization/acl/PentahoACLProvider.java
+++ b/repository/src/main/java/org/apache/jackrabbit/core/security/authorization/acl/PentahoACLProvider.java
@@ -17,10 +17,15 @@
 
 package org.apache.jackrabbit.core.security.authorization.acl;
 
-import java.security.Principal;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
+import org.apache.jackrabbit.api.security.principal.PrincipalManager;
+import org.apache.jackrabbit.core.NodeImpl;
+import org.apache.jackrabbit.core.SessionImpl;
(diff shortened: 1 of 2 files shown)


repository/src/main/java/org/apache/jackrabbit/core/security/authorization/acl/PentahoACLProvider.java
```diff
@@ -121,6 +120,39 @@ protected void updateRootAcl( SessionImpl systemSession, ACLEditor editor ) thro
...
+            if ( acEntry.getPrivileges() != null ) {
+              for ( Privilege privilege : acEntry.getPrivileges() ) {
+                if ( jcrReadAccessControlPriv.equals( privilege ) ) {
+                  // If the everyone principal already has the JCR_READ_ACCESS_CONTROL privilege, there's no need to
+                  // update ACL
+                  return false;
+                }
+              }
+            }
+          }
```
niehusa: Not sure if the logic is the same as below?

if ( acEntry.getPrincipal() != null && acEntry.getPrincipal().equals( everyone ) && acEntry.getPrivileges() != null  ) {
  return Arrays.asList( acEntry.getPrivileges() ).contains( jcrReadAccessControlPriv );
}

pentaho-amessier: Sure - that works too

PR #3919, opened 2017-10-30
[SP-3798] Backport of BISERVER-13774 - Rename folders created in Spoo…

mchen-len-son:
…n Repository Save dialog is not assumed in PUC


mchen-len-son:
@stepanovdg please review

PR #3922, opened 2017-11-01
[SP-3848][BISERVER-13722]-Error assigning role permissions through API under JDBC authentication

stepanovdg:


stepanovdg:

PR #3926, opened 2017-11-04
[PPP-3888] - Use of vulnerable component com.fasterxml.jackson.core :…

YuryBY:
… jackson-databind-2.5.4.jar ,com.fasterxml.jackson.core : jackson-databind-2.3.3.jar,com.fasterxml.jackson.core : jackson-databind-2.5.3.jar

@pamval, @mbatchelor, could you please take a look?

### DIFF
--- a/extensions/pom.xml
+++ b/extensions/pom.xml
@@ -21,7 +21,7 @@
     <license.header.definition.file>${basedir}/../license/styles/javadoc_style_license_header.xml</license.header.definition.file>
     <hamcrest-library.version>1.3</hamcrest-library.version>
     <jaxen.version>1.1.1</jaxen.version>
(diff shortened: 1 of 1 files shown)


extensions/pom.xml
```diff
@@ -21,7 +21,7 @@
     <license.header.definition.file>${basedir}/../license/styles/javadoc_style_license_header.xml</license.header.definition.file>
     <hamcrest-library.version>1.3</hamcrest-library.version>
     <jaxen.version>1.1.1</jaxen.version>
-    <fasterxml.jackson.version>2.2.3</fasterxml.jackson.version>
```
pamval: Shouldn't this be a property instead of a hard-coded value?

PR #3937, opened 2017-11-09
[PPP-3936] - Fix for reflected XSS

mbatchelor:


lucboudreau:
Hi @mbatchelor ,

I was trying to figure out why the unused imports. Turns out there's a class in the same package called FileResource. Not sure if you're supposed to use those either.

https://github.com/pentaho/pentaho-platform/blob/6326d108683a987a07a5299ce924f802ab76148c/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/FileResource.java

lucboudreau:
Deleted previous comment. Had the wrong class opened. Still unknown why there are usused imports.

mchen-len-son:
Tested locally using Fiddler - looks good.

PR #3948, opened 2017-11-23
[BISERVER-11204] Deprecate ISolutionFileMetaProvider, this interface …

mchen-len-son:
…is no longer used

 - deleted related message strings as they are not used anymore


mchen-len-son:
@pamval please review

PR #3965, opened 2017-12-05
[BACKLOG-19589] let admin access all trash

tgf:


tgf:
@bmorrise please review

PR #3969, opened 2017-12-08
[PPP-3949] Authenticated directory traversal allows retrieval of arbi…

mchen-len-son:
…trary files within Tomcat webapps/pentaho directory 
 - added check if canonical name of a file is equal to it`s path in request
 - added unit tests

PR #3971, opened 2017-12-19
[PPP-3949] Authenticated directory traversal allows retrieval of arbi…

mchen-len-son:
…trary files within Tomcat webapps/pentaho directory

- added support of double slashes in resource paths


mchen-len-son:
@pamval @rmansoor we found a regression after #3969.
this is a cure.
can you please take a look?
thanks.

PR #3973, opened 2017-12-26
[SP-3961] - Backport of ANALYZER-3666 - selectSchema method filters c…

YuryBY:
…ubes user has no access to but fills the logs with errors for those cubes with no access to (8.0 Suite)

@mchen-len-son, could you please take a look?


YuryBY:
@mchen-len-son, locally all test from PentahoMetadataDomainRepositoryConcurrencyTest pass. The failure above does not seem like related to the changes.

PR #3984, opened 2017-12-29
[PDI-16794]-Regular users can view and delete scheduled jobs belonging to other users

ArtsiomYurhevich:
Added accessing filter for returning jobs

### DIFF
--- a/scheduler/src/main/java/org/pentaho/platform/scheduler2/ws/DefaultSchedulerService.java
+++ b/scheduler/src/main/java/org/pentaho/platform/scheduler2/ws/DefaultSchedulerService.java
@@ -24,12 +24,16 @@ import javax.jws.WebService;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.pentaho.platform.api.engine.IAuthorizationPolicy;
+import org.pentaho.platform.api.engine.IPentahoSession;
 import org.pentaho.platform.api.scheduler2.ComplexJobTrigger;
+import org.pentaho.platform.api.scheduler2.IBlockoutManager;
 import org.pentaho.platform.api.scheduler2.IScheduler;
 import org.pentaho.platform.api.scheduler2.Job;
 import org.pentaho.platform.api.scheduler2.JobTrigger;
 import org.pentaho.platform.api.scheduler2.SchedulerException;
 import org.pentaho.platform.api.scheduler2.SimpleJobTrigger;
+import org.pentaho.platform.engine.core.system.PentahoSessionHolder;
 import org.pentaho.platform.engine.core.system.PentahoSystem;
 
 /**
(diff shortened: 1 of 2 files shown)


ArtsiomYurhevich:

kolinus:
@bmorrise @mdamour1976 could you please take a look

scheduler/src/main/java/org/pentaho/platform/scheduler2/ws/DefaultSchedulerService.java
```diff
@@ -99,7 +105,16 @@ private void updateJob( String jobId, Map<String, ParamValue> jobParams, JobTrig
...
-    return scheduler.getJobs( null ).toArray( new Job[0] );
+    IPentahoSession session = PentahoSessionHolder.getSession();
+    String principalName = session.getName();
+    Boolean canAdminister = PentahoSystem.get( IAuthorizationPolicy.class ).isAllowed( ADMIN_PERM );
+
+    return scheduler.getJobs( job -> {
+      if ( canAdminister ) {
+        return !IBlockoutManager.BLOCK_OUT_JOB_NAME.equals( job.getJobName() );
+      }
+      return principalName.equals( job.getUserName() );
```
pamval: is this safe for different casing for user names and all that madness (I recall a recent issue about that)

ArtsiomYurhevich: @pamval The filter is the same as we have in PUC in Schedule perspective now and it works correctly. See https://github.com/pentaho/pentaho-platform/blob/8ad9067e790c6997bd8aaed5e48ba5ee12df1e76/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/services/SchedulerService.java#L531. Pedro, what issue do you mean?

pamval: Can't locate it right now, but it had to do with the fact that you could login in spoon as ADMIN (or Admin or aDMin or whatever) and then you could not do certain actions or view certain content because the owner of the content was the user admin (all lower case). It was a fairly recent issue, I believe. Can you run a very quick test for that case?

ArtsiomYurhevich: @pamval Yes,sure. Will check it

ArtsiomYurhevich: I checked, the case is ok. Logged in via Spoon using usernames with upper/lower letters -> the returned list of schedules is correct

ArtsiomYurhevich:
@mbatchelor Done

mchen-len-son:
@mbatchelor, @pamval can you please take a look?

PR #3985, opened 2018-01-04
[BISERVER-13750]-Missing scroll bar on Manage Users Perspective when using Chrome

stepanovdg:


stepanovdg:
@mchen-len-son @kolinus Could you please take a look?

stepanovdg:
@mdamour1976 @pedrofvteixeira @mchen-len-son @kolinus Could you please take a look?

mchen-len-son:
@mbatchelor, @pamval can you please take a look?

PR #3991, opened 2018-01-10
[BACKLOG-19756] Fix integration tests

andrei-viaryshka:


andrei-viaryshka:
@rmansoor Could you take a look please?

PR #3992, opened 2018-01-10
[BACKLOG-19513] Refactor condition that checks if CAS sesion is in us…

befc:
…e, generalize it for SSO and make it parameterized


pedrofvteixeira:
@pentaho/rogueone please review it, but hold off on merging it until further notice.

befc:
Previously the way to check if we are in CAS, was by through the authentication providers to try to find the **CasAuthenticationProvider**. This can raise some problems, since we are looking for a specific bean, which can change in the future, and it won't work for other SSO scenarios.
With this change this logic was replaced with a variable **ssoEnabled** that we can set to this bean ( example for CAS: https://github.com/pentaho/pentaho-platform-ee/pull/1240/files#diff-ac2c99564cac089acde5518b78124308R85 ), like this we are not tying this logic to a specific Provider.

befc:
@pentaho/rogueone, @EgorZhuk, @mbatchelor, please review

PR #4003, opened 2018-01-19
[BACKLOG-20909] Apply the internal logic (flag) that would allow for a WorkItem to be executed either through MesosChronosActionInvoker or LocalActionInvoker

neeta-rathi:
Code to check useWorkerNodes flag before invoking remote action

@pentaho/rogueone

### DIFF
--- a/core/src/main/java/org/pentaho/platform/util/ActionUtil.java
+++ b/core/src/main/java/org/pentaho/platform/util/ActionUtil.java
@@ -12,7 +12,7 @@
  * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  * See the GNU Lesser General Public License for more details.
  *
- * Copyright (c) 2017 Hitachi Vantara..  All rights reserved.
+ * Copyright (c) 2018 Hitachi Vantara..  All rights reserved.
  */
 
 package org.pentaho.platform.util;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/pentaho/platform/util/ActionUtil.java
```diff
@@ -193,6 +194,27 @@ public static String extractUid( final Map<String, Serializable> params ) {
...
+   */
+  public static boolean extractUseWorkerNodes( final Map<String, Serializable> params ) {
+    if ( params == null ) {
+      return false;
+    }
+
+    String useWorkerNodes = (String) params.get( USE_WORKER_NODES );
+    if ( StringUtil.isEmpty( useWorkerNodes ) ) {
+      return false;
+    }
```
pedrofvteixeira: In cases where `useWorkerNodes` param was never passed along, then assume the default/fallback scenario.

Think for example of customers that
- have worker nodes enabled
- schedule ktrs/kjbs/prpts by calling `api/scheduler/job` endpoint directly ( i.e. not via UI )

In this case, the endpoint will remain working as always: **no `useWorkerNodes` param was ever included / passed along**; so pentaho system, if configured to do so, will leverage worker nodes capability ( whenever it's a supported `IAction` ).

So the only true impact of this param will take place when the 2 things take place:
1. `useWorkerNodes` param has been passed
2. `useWorkerNodes` was set to `false`

This translates to:
1. "normal" behaviour takes place when one of two happens:
  1.1.  `useWorkerNodes` param has been passed and is set to `true`
  1.2. `useWorkerNodes` param was never passed along
2. bypassing "normal" behaviour only takes place when:
  2.1. `useWorkerNodes` param has been passed and is set to `false`

neeta-rathi: Closing this PR.

neeta-rathi: Please see https://github.com/pentaho/pentaho-ee/pull/945

PR #4016, opened 2018-01-29
[SP-4051]-[BISERVER-13756]-Object Permissions Do Not Restore from Backup

PashkovichAndrei:


PashkovichAndrei:
@stepanovdg @mchen-len-son, could you please have a look?

PashkovichAndrei:
@stepanovdg @mchen-len-son could you please have a look?

PR #4021, opened 2018-01-31
[PPP-3949] Authenticated directory traversal allows retrieval of arbi…

mchen-len-son:
…trary files within Tomcat webapps/pentaho directory

- enabled directory traversal defend for all files including jars and not eisting ones


mchen-len-son:
@pamval @mbatchelor can you please take a look?

PR #4024, opened 2018-02-01
[SP-4084][PPP-3960]-Log4jReource.java vulnerable to Reflected Cross-s…

mchen-len-son:
…ite scripting


mchen-len-son:
@stepanovdg can you please take a look?

PR #4032, opened 2018-02-07
[PPP-3963] UIServlet.java needs to be removed from platform

Advent51:
- UIServlet was deprecated
- added escaping for the retrieved parameter
- mime type retrieving was modified in order to strongly match known mime types
- test was added
- a little refactoring was done


Advent51:
@pamval @mbatchelor @mdamour1976 please review

PR #4043, opened 2018-02-12
[BACKLOG-21237] Modified the two interface IAction and IActionInvokeS…

pdesai16:
…tatus

to get the execution status of ktr/kjb. These method have the default
keyward to maintain backward compatibility.
ActionRunner logs the correct WorkItemLifecyclePhase success or failure
depending of the IAction. ActionAdaptorQuartzJob now looks at the
IActionInvokeStatus.isExecutionSuccessful() in addition to  getThrowable()
to know whether the execution was successful.


pdesai16:
@pentaho/rogueone  please review

PR #4048, opened 2018-02-15
[BACKLOG-20290] Add xms and xmx related parameters and validation for…

befc:
… WN configuration

### DIFF
--- a/core/src/main/java/org/pentaho/platform/util/ActionUtil.java
+++ b/core/src/main/java/org/pentaho/platform/util/ActionUtil.java
@@ -94,6 +94,9 @@ public class ActionUtil {
     KEY_MAP.put( ActionUtil.QUARTZ_RESTART_FLAG, ActionUtil.INVOKER_RESTART_FLAG );
   }
 
+  public static final String WORK_ITEM_XMS = "JVM_XMS";
+  public static final String WORK_ITEM_XMX = "JVM_XMX";
+
   /**
    * Removes the provided {@code key} from the {@code params} {@link Map} , if it exists, as well as its
    * "equivalent", as defined within the {@code KEY_MAP}. If the provided {@code key} is a key within the {@code


befc:
@pentaho/rogueone, please review

befc:
wingman failed, but it’s a **Still Broken** test, not related to the changes made

core/src/main/java/org/pentaho/platform/util/ActionUtil.java
```diff
@@ -91,6 +91,27 @@
...
+  public static final String WORK_ITEM_XMX = "JVM_XMX";
+
+  /**
+   * Validates a JVM_XMS or JVM_XMX value to make sure it has the correct format.
+   * In case the format is not correct, it uses the default value.
+   *
+   */
+  public static boolean validateJvmXm( String jvmXmVar ) {
+    try {
+      int intValue = Integer.parseInt( jvmXmVar );
```
pedrofvteixeira: Have we considered commons-validator to do this work for us? 

https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/IntegerValidator.html#validate(java.lang.String)

pdesai16: Do we need to check for upper limit of this settings?

befc: I will add that check

befc: Will change to use the IntegerValidator

core/src/main/java/org/pentaho/platform/util/ActionUtil.java
```diff
@@ -94,6 +94,9 @@
     KEY_MAP.put( ActionUtil.QUARTZ_RESTART_FLAG, ActionUtil.INVOKER_RESTART_FLAG );
   }
 
+  public static final String WORK_ITEM_XMS = "JVM_XMS";
+  public static final String WORK_ITEM_XMX = "JVM_XMX";
+
```
pedrofvteixeira: I see why the declaration of these variables would be placed in pentaho-platform's core code ( to conform to the location where all the other variables are at as well ).

However, I see the other variables as being applicable for any and all `IAction`, while these JVM ones seem to be applicable only to someone who would care to have a notion of jvm settings running somewhere , such as a docker container.

So, by this line of reasoning, I'd have these variables declared in the only place they make sense, which is in `MesosChronosActionInvoker`

befc: It makes sense, I will close this PR and include these variables directly in the `MesosChronosActionInvoker` instead

befc:
Will close this PR and include these variables directly in the `MesosChronosActionInvoker` instead, as. pointed out in the comment by @pedrofvteixeira

befc:
Will close this PR and include these variables directly in the `MesosChronosActionInvoker` instead, as. pointed out in the comment by @pedrofvteixeira

PR #4052, opened 2018-02-23
[BISERVER-13757]-404 error handling for pentaho/getImage?image=... call

stepanovdg:


stepanovdg:
@pedrofvteixeira @e-cuellar Could you please take a look?

extensions/src/main/java/org/pentaho/platform/plugin/services/webservices/content/AbstractAxisServiceContentGenerator.java
```diff
@@ -117,10 +117,28 @@ protected void processAxisFault( MessageContext msgContext, OutputStream out, Th
...
+            } else {
+              res.setStatus( sc );
+            }
+          }
+        } catch ( IOException e1 ) {
+          if ( status == null ) {
+            res.setStatus( HttpServletResponse.SC_INTERNAL_SERVER_ERROR );
+          } else {
+            int sc = Integer.parseInt( status );
+            if ( sc >= 400 ) {
```
mbatchelor: Don't think you need this if - it's doing the same thing in the if and the else.

stepanovdg: Yes thanks it is mistype - will change

stepanovdg:
@mbatchelor Updated according to your comment.

PR #4063, opened 2018-02-28
Sp 3981

mchen-len-son:


mchen-len-son:
@pentaho-lmartins can you please review this?
wingman's complain looks outdated - copyright is fixed.

PR #4064, opened 2018-02-28
Sp 3980 7.1

mchen-len-son:


mchen-len-son:
@pentaho-lmartins can you please have a look?

PR #4065, opened 2018-03-01
[CLEANUP] fixed failing tests

mchen-len-son:


mchen-len-son:
PentahoMetadataDomainRepositoryTest and CachedPentahoMetadataDomainRepository fail from time to time due to unclosed input stream, and incorrect directory deletion.

@pentaho-lmartins @stepanovdg please have a look.

PR #4067, opened 2018-03-01
[CLEANUP] fixed failing tests

mchen-len-son:


mbatchelor:
@mchen-len-son - can you fix this PR ?

mchen-len-son:
@mbatchelor now the failure is gone.
I fixed it in master by #4068. 
not sure if we should backport the fix - a lot of config files would be changed.

mchen-len-son:
@pamval @pentaho-lmartins can you please merge this one?

PR #4068, opened 2018-03-02
[CLEANUP] fixed hibernate configuration/mapping files urls

mchen-len-son:


mchen-len-son:
@pamval @rmansoor can you please take a look?

PR #4073, opened 2018-03-09
[BACKLOG-20930] Update Run in Background dialog to match schedule dialog

ricksjames:
Update Run in Background dialog to match schedule dialog


ricksjames:
@pentaho/ackbar Please Review

PR #4096, opened 2018-03-29
[BACKLOG-22550] Exception thrown java.lang.ClassCastException when op…

pedrofvteixeira:
…ening kettle.cda sample

- commons-vfs2-2.2's `DefaultFileSystemManager` does a hardcoded cast of the FileProvider impl to `AbstractFileProvider`, in order to call for the new method `freeUnusedResources()` 
- `freeUnusedResources()` already has a default method impl in the abstract, so we needn't implement it ourselves
- `AbstractFileProvider implements FileProvider`

So making `SolutionRepositoryVfs extends AbstractFileProvider` instead on `SolutionRepositoryVfs implements FileProvider`  

Tested & validated on an end-to-end build ( thanks to @mdamour1976 's build tool )

CC'ing @pamval


pedrofvteixeira:
As an FYI, here's the hardcoded cast https://github.com/apache/commons-vfs/blob/commons-vfs2-project-2.2/commons-vfs2/src/main/java/org/apache/commons/vfs2/impl/DefaultFileSystemManager.java#L586

pedrofvteixeira:
The above are "Still broken tests", not "Newly broken" ones. Looking at them, they seem related to LDAP Auth UTs.

PR #4098, opened 2018-04-02
[BACKLOG-22335] CES - Allow work items (ktr, kjb) to be sent in from …

pedrofvteixeira:
…the PDI client tool

Should be merged alongside https://github.com/pentaho/pentaho-ee/pull/1124

### DIFF
--- a/core/src/main/java/org/pentaho/platform/util/ActionUtil.java
+++ b/core/src/main/java/org/pentaho/platform/util/ActionUtil.java
@@ -52,6 +52,8 @@ public class ActionUtil {
   public static final String QUARTZ_STREAMPROVIDER = "ActionAdapterQuartzJob-StreamProvider"; //$NON-NLS-1$
   public static final String QUARTZ_STREAMPROVIDER_INPUT_FILE =
     "ActionAdapterQuartzJob-StreamProvider-InputFile"; //$NON-NLS-1$
+  public static final String QUARTZ_STREAMPROVIDER_INLINE_INPUT_FILE = "input file ="; //$NON-NLS-1$
+  public static final String QUARTZ_STREAMPROVIDER_INLINE_OUTPUT_FILE = ":output file="; //$NON-NLS-1$
   public static final String QUARTZ_UIPASSPARAM = "uiPassParam"; //$NON-NLS-1$
   public static final String QUARTZ_LINEAGE_ID = "lineage-id"; //$NON-NLS-1$
   public static final String QUARTZ_RESTART_FLAG = "ActionAdapterQuartzJob-Restart"; //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/pentaho/platform/util/ActionUtil.java
```diff
@@ -210,11 +212,53 @@ public static String generateWorkItemUid( final Map<String, Serializable> params
...
+    return false;
+  }
+
+  public static String getInlineInputFileOnStreamProvider( final Map<String, Serializable> params ) {
+
+    if ( isInlinePassingOfInputOnStreamProvider( params ) ) {
+
+      try {
+
+        String streamProviderContent = ( params.containsKey( QUARTZ_STREAMPROVIDER )
```
pentaho-apandit: minor refactoring: looks like same logic as line 231. Could be refactored out into a common method.

core/src/main/java/org/pentaho/platform/util/ActionUtil.java
```diff
@@ -210,11 +212,53 @@ public static String generateWorkItemUid( final Map<String, Serializable> params
...
+                ? params.get( QUARTZ_STREAMPROVIDER ) : params.containsKey( INVOKER_STREAMPROVIDER )
+                ? params.get( INVOKER_STREAMPROVIDER ) : StringUtils.EMPTY ).toString().trim();
+
```
neeta-rathi: By now, we have done the same thing twice, as in to check if the params conatins key `QUARTZ_STREAMPROVIDER`, right?

pentaho-apandit:
@pedrofvteixeira Can we ignore the broken test?

core/src/main/java/org/pentaho/platform/util/ActionUtil.java
pedrofvteixeira: correct; while `isInlinePassingOfInputOnStreamProvider` is a predicate alone, `getInlineInputFileOnStreamProvider` does provide content; I felt it would be safer to wrap the `get` over that predicate check, because they are 2 separate `public static` methods; so this means that they can be called independently.

Because the predicate check is very light & quick, I felt the added safety-check is more beneficial than the eventual possibility of the check being done twice.

I can change this to **not** have the `get` call for the predicate. If you agree, then I could tackle this in a follow-up "Cleanup" PR, rather than blocking this one from getting processed.

Thoughts ?

core/src/main/java/org/pentaho/platform/util/ActionUtil.java
pedrofvteixeira: Makes sense. If you agree, then I could tackle this in a follow-up "Cleanup" PR, rather than blocking this one from getting processed.

pedrofvteixeira:
@pentaho-apandit it's a **Still** broken test, not a **Newly** broken test.
Which means that it is **not** a consequence of this PR, but rather something that has been broken, and remains that way even with this PR.

core/src/main/java/org/pentaho/platform/util/ActionUtil.java
neeta-rathi: Sounds good!

PR #4102, opened 2018-04-06
[BISERVER-13916] Mondrian roles displaying errors with: Lorg.springframework.security.core.GrantedAuthority; cannot be cast to java.util.Collection

Advent51:
- type mismatch were fixed
- tests were fixed
- new tests were added


Advent51:
@pedrofvteixeira @e-cuellar @rmansoor @mdamour1976 please review

pamval:
@Advent51 Why is the drop in coverage so large? Do all the modules compile and run tests locally with your change?

Advent51:
@pamval as I can see the coverage increased. Isn't it?
But we triggered wingman again to check it again.

mchen-len-son:
wingman's report actually says the test coverage increased.
but I can't say for sure why. probably, due to fixed tests

PR #4105, opened 2018-04-10
[BACKLOG-22388] - Adding missing js and css updates

pamval:


pamval:
@afrjorge, same as https://github.com/pentaho/pentaho-platform-ee/pull/1261 but for CE

PR #4106, opened 2018-04-12
[BACKLOG-21580] properties files were being ignored if there were at …

joaormf:
…least a locale file in the imported zip bundle


joaormf:
@pentaho/kashyyyk please review

PR #4132, opened 2018-05-10
[BISERVER-13932] Data source with the password contains special Chara…

bantonio82:
…cter '&' fails to login

@mbatchelor @pentaho-lmartins @pedrofvteixeira @pamval


core/src/main/java/org/pentaho/platform/engine/services/connection/datasource/dbcp/PooledDatasourceHelper.java
```diff
@@ -322,6 +313,22 @@ public static PoolingDataSource setupPooledDataSource( IDatabaseConnection datab
     }
   }
 
+  protected static ConnectionFactory getConnectionFactory( IDatabaseConnection databaseConnection, String url ) {
+    ConnectionFactory factory = null;
+    if ( url.startsWith( "jdbc:mysql:" ) || ( url.startsWith( "jdbc:mariadb:" ) ) ) {
```
mbatchelor: We cannot do this - this is a really bad practice. Please don't write code that resembles this either. We need symmetric encoding/decoding that's done for all databases, not targeted at MySQL/MariaDB.

bantonio82:
The PR was updated.

Please check Jira for details.

PR #4136, opened 2018-05-15
[BACKLOG-22917] fix clickable area

afrjorge:


afrjorge:

PR #4139, opened 2018-05-18
[SP-4305][BISERVER-13832]-Server is no longer displaying the appropriate error page for URLs starting with /pentaho/api

stepanovdg:


stepanovdg:
@pentaho-lmartins Could you please take a look?

PR #4145, opened 2018-05-22
[SP-4336][PDI-16908] Inspect Data, if you enter Japanese surrogate character as the data source name, "We couldn't connect" occurs.

AliaksandrShuhayeu:
[SP-4336][PDI-16908] Inspect Data, if you enter Japanese surrogate character as the data source name, "We couldn't connect" occurs.

- removed xalan dependency


AliaksandrShuhayeu:

PR #4153, opened 2018-06-04
[BISERVER-13925] Concurrency issue with retrieving metadata ids

ricardosilva88:
@mbatchelor 
@lucboudreau

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/plugin/services/metadata/SessionCachingMetadataDomainRepository.java
+++ b/extensions/src/main/java/org/pentaho/platform/plugin/services/metadata/SessionCachingMetadataDomainRepository.java
@@ -350,7 +350,7 @@ public class SessionCachingMetadataDomainRepository implements IMetadataDomainRe
           if ( k instanceof String ) {
             String key = (String) k;
             if ( key != null && key.startsWith( DOMAIN_CACHE_KEY_PREDICATE ) ) {
-              Set<String> domainIds = (Set<String>) cacheManager.getFromRegionCache( CACHE_REGION, key );
+              Set<String> domainIds = new HashSet<String>((Set<String>) cacheManager.getFromRegionCache( CACHE_REGION, key ));
               domainIds.remove( domainId );
               cacheManager.putInRegionCache( CACHE_REGION, key, domainIds );
             }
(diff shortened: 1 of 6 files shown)


lucboudreau:
Does the synchronization actually fix the issue? Because looking at the many catch blocks which only log an error or return an empty list, something tells me it does not actually solve the error but merely silences it. 

Like this for example, where the problem is merely sidestepped by returning a empty list when an exception is caught: https://github.com/pentaho/pentaho-platform/pull/4153/files#diff-4af456b98a068a464b8026ee8e7b95dfR354

Has this been tested under load, as it was discovered originally?

lucboudreau:
I wrote a load test for a similar thing recently. It is using a prototype of the CacheManager from another branch, but a few changes can make it work on your changes. It can stress the cache with as many clients and operations as you want to scale to.

https://github.com/lucboudreau/pentaho-platform/blob/b9ba15c587ed648a4729ac1ccea9f54c21d339c7/extensions/src/it/java/org/pentaho/platform/plugin/services/cache/CacheStressIT.java

extensions/src/main/java/org/pentaho/platform/plugin/services/metadata/SessionCachingMetadataDomainRepository.java
```diff
@@ -368,27 +370,63 @@ private void removeDomainFromIDCache( String domainId ) {
...
+    }
+  }
+
   protected String generateDomainIdCacheKeyForSession( IPentahoSession session ) {
     return DOMAIN_CACHE_KEY_PREDICATE + session.getId();
   }
 
   @Override
-  public void removeDomain( final String domainId ) {
+  public synchronized void removeDomain( final String domainId ) {
```
mbatchelor: By definition, this class is session based ( SessionCachingMetadataDomainRepository ). However, this change is synchronizing on the entire MetadataRepository instance. So removing a domain for one user in their own session ends up blocking another user in another session from storing their domain or removing their model.  I don't understand why this should be done at this level. If there really is synchronization required somewhere, I don't suspect it's here. What specifically happens in the multi-threaded testing without these three methods being synchronized that this solves ?

ricardosilva88:
DO NOT MERGE!

extensions/src/main/java/org/pentaho/platform/plugin/services/metadata/SessionCachingMetadataDomainRepository.java
```diff
@@ -350,7 +350,7 @@ private void removeDomainFromIDCache( String domainId ) {
           if ( k instanceof String ) {
             String key = (String) k;
             if ( key != null && key.startsWith( DOMAIN_CACHE_KEY_PREDICATE ) ) {
-              Set<String> domainIds = (Set<String>) cacheManager.getFromRegionCache( CACHE_REGION, key );
+              Set<String> domainIds = new HashSet<String>((Set<String>) cacheManager.getFromRegionCache( CACHE_REGION, key ));
               domainIds.remove( domainId );
```
ricardosilva88: Used here a new HashSet because all cache thread safe mechanisms are not eligible for hash objects once using a direct reference

extensions/src/main/java/org/pentaho/platform/plugin/services/metadata/SessionCachingMetadataDomainRepository.java
```diff
@@ -403,20 +403,22 @@ public void removeModel( final String domainId, final String modelId ) throws Do
     if ( domainIdsCacheEnabled ) {
       domainIds = (Set<String>) cacheManager.getFromRegionCache( CACHE_REGION, domainKey );
       if ( domainIds != null ) {
+        Set<String> domainIdsToEdit = new HashSet<String>( domainIds );
```
ricardosilva88: Same here

extensions/src/main/java/org/pentaho/platform/plugin/services/metadata/SessionCachingMetadataDomainRepository.java
```diff
@@ -350,7 +350,7 @@ private void removeDomainFromIDCache( String domainId ) {
           if ( k instanceof String ) {
             String key = (String) k;
             if ( key != null && key.startsWith( DOMAIN_CACHE_KEY_PREDICATE ) ) {
-              Set<String> domainIds = (Set<String>) cacheManager.getFromRegionCache( CACHE_REGION, key );
+              Set<String> domainIds = new HashSet<String>((Set<String>) cacheManager.getFromRegionCache( CACHE_REGION, key ));
```
ricardosilva88: Used here a new HashSet because all cache thread safe mechanisms are not eligible for hash objects once using a direct reference

PR #4159, opened 2018-06-06
[BISERVER-13898] Created an new folder in root at same level of Home …

pentaho-lmartins:
…and Public folders

@pamval @mbatchelor @pedrofvteixeira

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/utils/FileUtils.java
+++ b/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/utils/FileUtils.java
@@ -24,6 +24,7 @@ import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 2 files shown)


extensions/src/main/java/org/pentaho/platform/web/http/api/resources/utils/FileUtils.java
```diff
@@ -93,26 +94,39 @@ public static boolean containsReservedCharacter( String path, char[] reserved )
...
+      return false;
+    }
+
+    return true;
   }
 
-  public static boolean isRootLevelPath( final String path ) {
-    return ( path != null ) && !isRootPath( path ) && path.startsWith( PATH_SEPARATOR ) && isRootPath( getParentPath( path ) );
+  private static String trimTrailingSlash( String path ) {
+    return path.endsWith( PATH_SEPARATOR ) ? path.substring( 0, path.length() - 2 ) : path;
```
mbatchelor: Null check was removed . . . why ?

Looking at the entire file, the method is private, and null is prevented before calling the method.

PR #4161, opened 2018-06-14
[PPP-4108] Use of vulnerable component jackson-databind-2.8.8.jar, ja…

pentaho-lmartins:
…ckson-databind-2.9.2.jar, jackson-mapper-asl-1.9.2.jar CVE-2017-7525, CVE-2017-15095, CVE-2017-15095

This is one of a series of PRs. See further details in https://github.com/pentaho/maven-parent-poms/pull/66.

**Do not merge without checking the status of the JIRA case.**

### DIFF
--- a/assemblies/pentaho-war/pom.xml
+++ b/assemblies/pentaho-war/pom.xml
@@ -31,9 +31,6 @@
     <jcommon-xml.version>1.0.12</jcommon-xml.version>
     <svgSalamander.version>1.0</svgSalamander.version>
     <package.resources.directory>${basedir}/src/main/webapp</package.resources.directory>
-    <!-- jackson-databind depends on jackson-core and jackson-annotations
-    so that this artifacts have the same version to prevent compile errors -->
-    <fasterxml.jackson.core.version>2.9.4</fasterxml.jackson.core.version>
     <replacer.version>1.5.2</replacer.version>
     <kafka-clients.version>0.10.2.1</kafka-clients.version>
     <paho.version>1.2.0</paho.version>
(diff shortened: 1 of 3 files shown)


assemblies/pentaho-war/pom.xml
```diff
@@ -898,39 +895,6 @@
...
-          <artifactId>*</artifactId>
-          <groupId>*</groupId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-    <dependency>
-      <groupId>com.fasterxml.jackson.core</groupId>
-      <artifactId>jackson-databind</artifactId>
-      <version>${fasterxml.jackson.core.version}</version>
-      <exclusions>
```
graimundo: @pentaho-lmartins Isn't this removing jackson from the provided libs of the war distribution? Similiar to my comment on the pdi client lib module. This would mean that plugins would not have access to jackson and would need to include it by their own.
Also, a general check that can be made, specially in maven modules that are "assemblies", is to do a diff on in the final artifact and check if we're removing anything that could have an undesired impact.

pentaho-lmartins: @graimundo, I can confirm that `jackson-databind`, `jackson-core` and `jackson-annotations` are present in the _lib_ of the war file. Just checked by downloading and opening the CE and EE versions from the last test stack build.

All three jackson dependencies come as a transitive from **pentaho-platform-extensions**.

Ditto for the **pdi-core-libs** and **pdi-libs** assemblies. Both assemblies include kettle-engine and only exclude `javax.servlet`. So jackson dependencies are included as transitives.

Do we want to make these dependencies explicit?

graimundo: @pentaho-lmartins I think that for the scope of this story it is ok to remove the explicit dependencies and let them come in transitively. That means we are assuming that the lib contains the necessary libs for the Pentaho code base and that we are not putting anything there just for the sake of third parties. So lets go with this solution to unblock the work and create that followup thread that we talked about. If necessary, as a result of that discussion we can re add the dependencies explicitly later on.

PR #4162, opened 2018-06-14
[BACKLOG-24049] As a Pentaho user I would like to delegate ETL work i…

pedrofvteixeira:
…tem execution to Worker Nodes in a synchronous fashion, to that I can scale out 'Open' / 'Open in new window' actions as well

Making the changes on the Pentaho Server side to account for this new `IContentGeneratorInvoker`, that will henceforth be leveraged by Worker Nodes OSGi bundle


pedrofvteixeira:
Wingman states "Still broken tests", not "Newly broken", which means that this PR is not the cause of it.

neeta-rathi:
Yes, for other open PRs wingman is reporting the same broken tests. I'll merge this PR.

PR #4164, opened 2018-06-22
[BACKLOG-22942] Use randomUUID to generate quartzJobKey, added misfire instruction and logs

neeta-rathi:
1. Changed `QuartzJobKey` to use `randomUUID `instead of `currentTimeInMillis `to create unique QuartzJobKeys
2. Changed `QuartzScheduler` to add `MISFIRE_INSTRUCTION_FIRE_NOW` for `SimpleTrigger `with `repeatCount = 0`
3. Added more logs on `QuartzScheduler`

@pentaho/rogueone

### DIFF
--- a/scheduler/src/main/java/org/pentaho/platform/scheduler2/quartz/QuartzJobKey.java
+++ b/scheduler/src/main/java/org/pentaho/platform/scheduler2/quartz/QuartzJobKey.java
@@ -36,7 +36,8 @@ import org.pentaho.platform.scheduler2.messsages.Messages;
 public class QuartzJobKey {
   private String userName;
   private String jobName;
-  private long timeInMillis;
+  // BACKLOG-22942, changing timInMillis from long to randomUUID
+  private String randomUuid;
 
   /**
    * Use this constructor when you wish to create a new unique job key.
@@ -56,7 +57,7 @@ public class QuartzJobKey {
     }
     userName = username;
     this.jobName = jobName;
-    timeInMillis = System.currentTimeMillis();
+    randomUuid = java.util.UUID.randomUUID().toString();
   }
 
   private QuartzJobKey() {
@@ -80,12 +81,8 @@ public class QuartzJobKey {
     QuartzJobKey key = new QuartzJobKey();
     key.userName = elements[0];
     key.jobName = elements[1];
-    try {
-      key.timeInMillis = Long.parseLong( elements[2] );
-    } catch ( NumberFormatException ex ) {
-      throw new SchedulerException( MessageFormat.format( Messages.getInstance().getErrorString(
-          "QuartzJobKey.ERROR_0002" ), jobId ) ); //$NON-NLS-1$
-    }
+    key.randomUuid = elements[2];
+
     return key;
   }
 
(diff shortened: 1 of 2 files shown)


scheduler/src/main/java/org/pentaho/platform/scheduler2/quartz/QuartzJobKey.java
```diff
@@ -56,7 +57,7 @@ public QuartzJobKey( String jobName, String username ) throws SchedulerException
     }
     userName = username;
     this.jobName = jobName;
-    timeInMillis = System.currentTimeMillis();
+    randomUuid = java.util.UUID.randomUUID().toString();
```
pedrofvteixeira: Are we not keeping `System.currentTimeMillis()`, and merely adding one extra "check digit" at the end of it ? Do we absolutely need to switch from `System.currentTimeMillis()` to `java.util.UUID.randomUUID()` ?

neeta-rathi: How do we make sure that "check digit" is random enough in concurrent use case like this? I ran 5 threads to generate 5 random numbers each using `ThreadLocalRandom.current().nextInt(1, 100)`  and I saw it's possible for multiple threads to generate a same random number. 

Also, for backward compatibility ( existing triggers with timeInMillis in their name ), when the name gets parsed, timeInMillis gets parsed to this new  UUID ( String ) field.  

Another thing I considered was the database limitations where triggers are stored, `qrtz5_triggers` and `qrtz5_simple_triggers` both assign `VARCHAR(200)` for trigger_name column. 
`java.util.UUID.randomUUID().toString()` length is always 32 and  `System.currentTimeMillis()` is always 13, with this change we're adding an overhead of 19 characters, but the field length ( 200 ) is long enough to store ( username + transformation name + UUID ).

May be we don't need to change `System.currentTimeMillis()` and add a random digit, but in presence of existing Java classes like `UUID.randomUUID()`  which has a very low probability for collision, I chose that over changing `System.currentTimeMillis()`. 

Let me know what you think or share any ideas..I would be happy to make it work with minimal impact :)

pedrofvteixeira: If we change from `System.currentTimeMillis()` to `UUID.randomUUID()` ( i.e. change from long to String ), will this cause any issues on customers that have scheduled / periodic quartz jobs already in place on versions 8.1 and below ?

neeta-rathi: No, that's what I meant in my comment about backward compatibility.

Current name format `admin PentahoSystemVersionCheck	1529669124679`
With this PR format `admin	PentahoSystemVersionCheck	04f3a7a5-aba6-4c2d-b238-f0081ef8b5cb`
`long` field got changed to `String`, and a call to `QuartzJobKey.parse` can handle storing string representation of long in a String variable. I sure about this, but I'm testing this out as we speak..

pedrofvteixeira: As long as we can ensure/safeguard any unforeseen backward compat. issues, this PR is 👍 by me.

neeta-rathi: Confirming that existing schedules work as expected. Also, if they are user generated schedule ( vs. system generated like `PentahoSystemVersionCheck` ), they retain their naming convention in the database as `username scheduleName timeInMillis` eg ( `admin hello 1529669124679` ). All the new schedules generated with this change can maintain their naming convention `username scheduleName uuid`

PR #4166, opened 2018-06-22
[BISERVER-13648] Double login after logout when viewing a report

pentaho-lmartins:
This PR fixes the double auth issue. For this we need the new `PentahoBasicAuthenticationEntryPoint` and we need to point the bean _basicProcessingFilterEntryPoint_ to this new class.

The change in line 26 changes the default behavior to redirect to the configured login page when an unauthenticated user tries to access a report, instead of using the browser's popup dialog. After a successful authentication, the user will e redirected back to the report.

@pmalves @pamval @mbatchelor @pedrofvteixeira


pedrofvteixeira:
Looks good. Question for the upgrade path @mbatchelor @pamval @kcruzada @dkincade : we are changing a spring config file. Customers porting over their own files will overrride this one, and effectively regressing on this bug. 

Should this be added to the "upgrade path" documentation? Manually changing this file?

pamval:
yeah, I think this needs to be added to the upgrade path documentation

pentaho-lmartins:
Applied the same change in spring configuration file to https://github.com/pentaho/pentaho-platform-ee/pull/1271.

PR #4185, opened 2018-07-10
[BISERVER-13893] In Firefox Quantum, with a new installation, when we…

pentaho-lmartins:
… open PUC or open a sample it gives the Console error 'XML Parsing Error: no root element found Location: http://localhost:8080/pentaho/api/user-settings/favorites'

- Updates `UserSettingsResource` to always set the content type header instead of leaving the browser to guess the type of the content in the payload. This is even more important when there is no content.
- Sets the correct HTTP status code when there is no content: **204** instead of **200**. Both represent success but the former will not trigger the JSON parser in jQuery and will trigger the success handler instead of the error one.

@kurtwalker @ricardosilva88 @cravobranco @pedrofvteixeira


cravobranco:
@pentaho-lmartins Looks good!

PR #4186, opened 2018-07-13
[BISERVER-13893] In Firefox Quantum, with a new installation, when we…

pentaho-lmartins:
… open PUC or open a sample it gives the Console error 'XML Parsing Error: no root element found Location: http://localhost:8080/pentaho/api/user-settings/favorites'

Additional fixes to complement #4185

@cravobranco @ricardosilva88


extensions/src/main/java/org/pentaho/platform/web/http/api/resources/UserConsoleResource.java
```diff
@@ -36,6 +36,7 @@
 import org.pentaho.platform.plugin.action.mondrian.catalog.MondrianCatalog;
 import org.pentaho.platform.plugin.action.mondrian.catalog.MondrianCube;
 import org.pentaho.platform.plugin.services.pluginmgr.IAdminContentConditionalLogic;
+import org.pentaho.platform.util.StringUtil;
```
cravobranco: This import was added but it's not used. I think wingman will complain...

pentaho-lmartins:
PR updated to fix broken test cases and unused import. @cravobranco

PR #4187, opened 2018-07-16
[BISERVER-13970] - CC and BCC fields on schedules are throwing NullPo…

cravobranco:
…interExceptions for imported schedules

* Fixed NPE when Cc and Bcc fields don't exist during a schedule execution.

@pentaho-lmartins , @pamval , @mbatchelor , @mdamour1976 Please review and merge if it's all ok.

### DIFF
--- a/core/src/main/java/org/pentaho/platform/util/Emailer.java
+++ b/core/src/main/java/org/pentaho/platform/util/Emailer.java
@@ -66,22 +66,22 @@ public class Emailer {
   }
 
   public void setTo( String to ) {
-    to = to.replaceAll( ";", "," );
     if ( to != null && !"".equals( to ) ) {
+      to = to.replaceAll( ";", "," );
       props.put( "to", to );
     }
   }
 
   public void setCc( String cc ) {
-    cc = cc.replaceAll( ";", "," );
     if ( cc != null && !"".equals( cc ) ) {
+      cc = cc.replaceAll( ";", "," );
       props.put( "cc", cc );
     }
   }
 
   public void setBcc( String bcc ) {
-    bcc = bcc.replaceAll( ";", "," );
     if ( bcc != null && !"".equals( bcc ) ) {
+      bcc = bcc.replaceAll( ";", "," );
       props.put( "bcc", bcc );
     }
   }
@@ -160,6 +160,10 @@ public class Emailer {
 
   public String getEmailFromName() {
     return Messages.getInstance().getString( "emailFromName" ); //$NON-NLS-1$
+  }
+
+  public Properties getProperties() {
+    return props;
   }
 
   public boolean setup() {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/pentaho/platform/util/Emailer.java
```diff
@@ -65,23 +65,23 @@
   public Emailer() {
   }
 
-  public void setTo( String to ) {
-    to = to.replaceAll( ";", "," );
+  void setTo( String to ) {
```
pentaho-lmartins: @cravobranco, `removing` the public access modifier makes these methods invisible outside of the package and will break code using these methods from other packages. Please update the PR with the `public` access modifier.

See: https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html

core/src/main/java/org/pentaho/platform/util/Emailer.java
```diff
@@ -65,23 +65,23 @@
...
     if ( to != null && !"".equals( to ) ) {
+      to = to.replaceAll( ";", "," );
       props.put( "to", to );
     }
   }
 
-  public void setCc( String cc ) {
-    cc = cc.replaceAll( ";", "," );
+  void setCc( String cc ) {
```
pentaho-lmartins: Same here.

core/src/main/java/org/pentaho/platform/util/Emailer.java
```diff
@@ -65,23 +65,23 @@
...
     if ( cc != null && !"".equals( cc ) ) {
+      cc = cc.replaceAll( ";", "," );
       props.put( "cc", cc );
     }
   }
 
-  public void setBcc( String bcc ) {
```
pentaho-lmartins: And here.

core/src/main/java/org/pentaho/platform/util/Emailer.java
```diff
@@ -162,6 +162,10 @@ public String getEmailFromName() {
     return Messages.getInstance().getString( "emailFromName" ); //$NON-NLS-1$
   }
 
+  Properties getProperties() {
```
pentaho-lmartins: Without the access modifier this method will be visible inside the package but not by subclasses or classes outside the package. Is this what you want?

core/src/main/java/org/pentaho/platform/util/Emailer.java
cravobranco: Tightening the access control as much as possible and avoiding to broaden it without reason is a basic Java OO rule - it enforces the Encapsulation property of this language.

Also, from the link you shared:
"
- Use the most restrictive access level that makes sense for a particular member. Use private unless you have a good reason not to.
- Avoid public fields except for constants. (Many of the examples in the tutorial use public fields. This may help to illustrate some points concisely, but is not recommended for production code.) Public fields tend to link you to a particular implementation and limit your flexibility in changing your code.
"

PR #4188, opened 2018-07-16
[PDI-16909] Remove a schedule using the REST API using HTTP DELETE me…

pentaho-lmartins:
…thod

- Adds the option to delete a scheduled job using the PUT verb instead of DELETE since DELETE can't / shouldn't have data in the request body. See JIRA case for explanation.

This is related to work done in https://github.com/pentaho/pentaho-kettle/pull/5594.

@mbatchelor @pedrofvteixeira


pentaho-lmartins:
@kurtwalker, can you take a look at this?

PR #4189, opened 2018-07-17
[BACKLOG-22672] Added utility method and new work item constants

neeta-rathi:
Added utility method on `WorkItemLifecycleEventUtil ` and new constants on `WorkItemLifecyclePhase` to support events posting from etl-execution plugin.

To be merged alongside https://github.com/pentaho/pentaho-ee/pull/1384

@pentaho/rogueone

### DIFF
--- a/core/src/main/java/org/pentaho/platform/workitem/WorkItemLifecyclePhase.java
+++ b/core/src/main/java/org/pentaho/platform/workitem/WorkItemLifecyclePhase.java
@@ -47,6 +47,14 @@ public enum WorkItemLifecyclePhase implements IWorkItemLifecyclePhase {
    */
   IN_PROGRESS,
   /**
+   * The work item execution is in progress
+   */
+  COMPLETED,
+  /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/pentaho/platform/workitem/WorkItemLifecyclePhase.java
```diff
@@ -46,6 +46,14 @@
    * The work item execution is in progress
    */
   IN_PROGRESS,
+  /**
+   * The work item execution is in progress
+   */
+  COMPLETED,
```
pedrofvteixeira: what's the difference between `COMPLETED` and `SUCCEEDED` , a few lines below? Is the latter one being used anywhere? If not, we could clean it, to remove confusions

neeta-rathi: It's been used by ActionRunner.

PR #4205, opened 2018-07-30
[BACKLOG-23730] As an admin I would like to have a user that could upload/download files using PUC outside of its home folder

neeta-rathi:
1. Created a new config property `download-roles` on `pentaho.xml`. This property holds a comma-separated list of roles that can download ( outside of their home folder ) content from repository
2. Added two end endpoints on `FileResource `- `/canUpload` and `/canDownload`
3. These endpoints now calls `SystemUtils` utility methods `canUpload` and `canDownload` and defer the decision of uploading/downloading content

Per this new logic, rules for

1. Downloading content:
If a user has permissions to "Read content" and "Create content", plus one of 2: "Administrator" OR "has a role listed on `<download-roles>`

2. Uploading content: ( no changes here )
If a user has permissions to "Read content" and "Create content", plus one of 2: "Administrator" OR "Publish Content"

@pentaho/rogueone

### DIFF
--- a/assemblies/pentaho-solutions/src/main/resources-filtered/pentaho-solutions/system/pentaho.xml
+++ b/assemblies/pentaho-solutions/src/main/resources-filtered/pentaho-solutions/system/pentaho.xml
@@ -23,6 +23,13 @@
       If true, show hint about sample users. Ultimately, should replace login-show-users-list.
     -->
     <login-show-sample-users-hint>true</login-show-sample-users-hint>
+
+	  <!--
+	  The download-roles defines user/roles that have download action permissions. It accepts a comma-separated list of roles.
+	  System admins must change what are the users/roles that have permissions for download action. If not defined here,
+	  the default value is set to 'Administrator'.
+	  -->
+	  <download-roles>Administrator</download-roles>
         
     <!--
 	This is the URL to the user guide
--- a/assemblies/pentaho-solutions/src/main/resources-filtered/pentaho-solutions/system/pentaho.xml
+++ b/assemblies/pentaho-solutions/src/main/resources-filtered/pentaho-solutions/system/pentaho.xml
@@ -24,12 +24,12 @@
     -->
     <login-show-sample-users-hint>true</login-show-sample-users-hint>
 
-	  <!--
-	  The download-roles defines user/roles that have download action permissions. It accepts a comma-separated list of roles.
-	  System admins must change what are the users/roles that have permissions for download action. If not defined here,
-	  the default value is set to 'Administrator'.
-	  -->
-	  <download-roles>Administrator</download-roles>
+	<!--
+	The download-roles defines user/roles that have download action permissions. It accepts a comma-separated list of roles.
+	System admins must change what are the users/roles that have permissions for download action. If not defined here,
+	the default value is set to 'Administrator'.
+	-->
+	<download-roles>Administrator</download-roles>
         
     <!--
 	This is the URL to the user guide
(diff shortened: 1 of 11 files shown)


user-console/src/main/resources/org/pentaho/mantle/public/browser/index.jsp
```diff
@@ -286,9 +286,11 @@
     }
   }
 
+  // With the fix BACKLOG-23730, server-side and client-side code uses centralized logic to check if user
+  // can download/upload content
   function checkDownload() {
     $.ajax({
-      url: CONTEXT_PATH + "api/authorization/action/isauthorized?authAction=org.pentaho.security.administerSecurity",
+      url: CONTEXT_PATH + "api/repo/files/canDownload?dirPath=",
```
pedrofvteixeira: If we're in a context where we do not have a selected `dirPath` yet, can we keep the endpoint clearer, by not including the queryParam?

assemblies/pentaho-solutions/src/main/resources-filtered/pentaho-solutions/system/pentaho.xml
```diff
@@ -23,6 +23,13 @@
       If true, show hint about sample users. Ultimately, should replace login-show-users-list.
     -->
     <login-show-sample-users-hint>true</login-show-sample-users-hint>
+
+	  <!--
+	  The download-roles defines user/roles that have download action permissions. It accepts a comma-separated list of roles.
+	  System admins must change what are the users/roles that have permissions for download action. If not defined here,
+	  the default value is set to 'Administrator'.
+	  -->
+	  <download-roles>Administrator</download-roles>
```
pedrofvteixeira: Not a blocker, but if we can let's have this new xml block properly indented / aligned.

user-console/src/main/resources/org/pentaho/mantle/public/browser/index.jsp
neeta-rathi: You're right, I'll fix this!

assemblies/pentaho-solutions/src/main/resources-filtered/pentaho-solutions/system/pentaho.xml
neeta-rathi: It shows fine in my setup. Let me check if I can fix this.

neeta-rathi:
Indentation on pentaho.xml for download-roles is not working for me, it's not right for other elements as well, shall I fix it all?

befc:
@neeta-rathi From what I can see in the raw view, problem is that in some places it has spaces in others has tabs. And that seems to be what is causing the discrepancy: https://raw.githubusercontent.com/neeta-rathi/pentaho-platform/1bc2dccdbb25ceabe698764359ea38c464c14160/assemblies/pentaho-solutions/src/main/resources-filtered/pentaho-solutions/system/pentaho.xml

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/utils/SystemUtils.java
```diff
@@ -21,16 +21,89 @@
...
+    //check if we are admin or have download-roles
+    boolean isAdminOrHaveDownloadActionRole = policy.isAllowed( RepositoryReadAction.NAME )
+      && policy.isAllowed( RepositoryCreateAction.NAME )
+      && ( policy.isAllowed( AdministerSecurityAction.NAME )
+      || !Collections.disjoint( tenantedUserRoles, PentahoSystem.getDownloadRolesList() ) );
+
+    //the user does not have admin or download-roles assigned, so we will check if the user downloads from their home folder
+    if ( !isAdminOrHaveDownloadActionRole && !StringUtils.isEmpty( downloadDir ) ) {
+      return validateAccessToHomeFolder( downloadDir );
+    }
```
pentaho-apandit: Might be more efficient to check this first so we skip all the role and permission checks if downloadDir is under user home folder.

PR #4213, opened 2018-08-09
[BACKLOG-24621] - Fix usage of wrong ace setter

eliofreitas:


pedrofvteixeira:
PR looks good.

I'd take this opportunity and also **FIX** that misleading aces setter that has been lingering around, and that does more than what would expect of it:

```
public void setAces( List<RepositoryFileAclAceDto> aces ) {
    setAces( aces, false /* entriesInheriting */ );
}
```

My concern is that, by keeping it this way, others in the future might unknowingly end up falling onto the same issue.

My $0.2c.

PR #4214, opened 2018-08-09
[BACKLOG-24896] Various fixes needed for the CDE on Foundry assembly

nantunes:
@pentaho/millenniumfalcon, @amartins23 and @pedrofvteixeira please review.

I recommend merging without squash, so the individual mostly unrelated commits are kept (for eventual cherry-picks and/or reverts).

~Don't forget https://github.com/pentaho/pentaho-platform-ee/pull/1273.~

**Note for the future:**
This PR removed the `/api/repo/files/generatedContentForSchedule` endpoint.

This endpoint just mirrored `/api/scheduler/generatedContentForSchedule` and was introduced in #1970 for no apparent reason.
It introduced a dependency with the scheduling features and related libs, that might be unavailable in some particular Pentaho platform assemblies.

Also see discussion bellow: https://github.com/pentaho/pentaho-platform/pull/4214#discussion_r209020930.

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/web/http/ConfigurationAdminNonOsgiProxy.java
+++ b/extensions/src/main/java/org/pentaho/platform/web/http/ConfigurationAdminNonOsgiProxy.java
@@ -23,7 +23,6 @@ package org.pentaho.platform.web.http;
 import org.osgi.service.cm.ConfigurationAdmin;
 import org.pentaho.platform.engine.core.system.PentahoSystem;
 
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Dictionary;
 
@@ -54,14 +53,14 @@ public class ConfigurationAdminNonOsgiProxy {
   public Dictionary<String, Object> getProperties( String persistenceID ) {
     Object configuration = getConfiguration( persistenceID );
     if ( configuration == null ) {
-        return null;
+      return null;
     }
 
     try {
       Method getProperties = configuration.getClass().getMethod( "getProperties" );
 
       return (Dictionary<String, Object>) getProperties.invoke( configuration );
-    } catch ( NoSuchMethodException | InvocationTargetException | IllegalAccessException e ) {
+    } catch ( Exception e ) {
       return null;
     }
 
@@ -79,7 +78,7 @@ public class ConfigurationAdminNonOsgiProxy {
       Method getConfiguration = this.configurationAdmin.getClass().getMethod( "getConfiguration", String.class );
 
       return getConfiguration.invoke( this.configurationAdmin, persistenceID );
-    } catch ( NoSuchMethodException | InvocationTargetException | IllegalAccessException e ) {
+    } catch ( Exception e ) {
       return null;
     }
 
--- a/repository/src/main/java/org/pentaho/platform/repository2/unified/fs/FileSystemRepositoryFileDao.java
+++ b/repository/src/main/java/org/pentaho/platform/repository2/unified/fs/FileSystemRepositoryFileDao.java
@@ -68,6 +68,7 @@ public class FileSystemRepositoryFileDao implements IRepositoryFileDao {
     Collections.unmodifiableList( Arrays.asList( new Character[] { '/', '\\', '\t', '\r', '\n' } ) );
   private static List<Character> reservedCharsWindows =
     Collections.unmodifiableList( Arrays.asList( new Character[]{ '?', '*', ':', '<', '>', '|'} ) );
+  private boolean IS_WINDOWS;
 
   public FileSystemRepositoryFileDao() {
     this( new File( System.getProperty( "solution.root.dir", System.getProperty( "user.dir" ) ) ) );
@@ -80,7 +81,9 @@ public class FileSystemRepositoryFileDao implements IRepositoryFileDao {
   public FileSystemRepositoryFileDao( File baseDir ) {
     // Detect OS
     final String os = System.getProperty( "os.name" ).toLowerCase();
-    if ( os.contains( "win" ) && baseDir.getPath().equals( "\\" ) ) {
+    IS_WINDOWS = os.contains( "win" );
+
+    if ( IS_WINDOWS && baseDir.getPath().equals( "\\" ) ) {
       baseDir = new File( "C:\\" );
     }
     assert ( baseDir.exists() && baseDir.isDirectory() );
@@ -98,7 +101,7 @@ public class FileSystemRepositoryFileDao implements IRepositoryFileDao {
   private byte[] inputStreamToBytes( InputStream in ) throws IOException {
 
     ByteArrayOutputStream out = new ByteArrayOutputStream( 4096 );
-    byte[] buffer = new byte[4096];
+    byte[] buffer = new byte[ 4096 ];
     int len;
 
     while ( ( len = in.read( buffer ) ) >= 0 ) {
@@ -225,7 +228,10 @@ public class FileSystemRepositoryFileDao implements IRepositoryFileDao {
       String jcrPath = f.getAbsolutePath().substring( rootDir.getAbsolutePath().length() );
       if ( jcrPath.length() == 0 ) {
         jcrPath = "/";
+      } else if ( IS_WINDOWS ) {
+        jcrPath = FilenameUtils.separatorsToUnix( jcrPath );
       }
+
       file =
           new RepositoryFile.Builder( f.getAbsolutePath(), f.getName() ).createdDate( new Date( f.lastModified() ) )
               .lastModificationDate( new Date( f.lastModified() ) ).folder( f.isDirectory() ).versioned( false ).path(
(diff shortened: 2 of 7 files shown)


assemblies/pentaho-solutions/src/main/resources/pentaho-solutions/system/pentahoServices.spring.xml
```diff
@@ -132,6 +132,7 @@
   <bean class="org.pentaho.platform.web.http.api.resources.FileResource" scope="request">
     <property name="whitelist"><ref bean="RepositoryDownloadWhitelist"/></property>
   </bean>
+  <bean class="org.pentaho.platform.web.http.api.resources.FileResourceGeneratedContentForSchedule" scope="request"/>
```
pedrofvteixeira: This change is taking place on a file that'll be eligible for "upgrade path for 9.0" ( i.e. migrating conf. files from a previous version to 9.0 ).

- If a `pentahoServices.spring.xml` from previous pentaho-server version gets ported over, will this separation have a graceful fallback behaviour ?
- Should this PR be accompanied of updated to the migration scripts, to inject this line in a ported `pentahoServices.spring.xml` ?
- Should this have to be added to the upgrade path documentation ?

extensions/src/main/java/org/pentaho/platform/web/http/ConfigurationAdminNonOsgiProxy.java
```diff
@@ -61,7 +60,7 @@ public ConfigurationAdminNonOsgiProxy() {
       Method getProperties = configuration.getClass().getMethod( "getProperties" );
 
       return (Dictionary<String, Object>) getProperties.invoke( configuration );
-    } catch ( NoSuchMethodException | InvocationTargetException | IllegalAccessException e ) {
+    } catch ( Exception e ) {
```
pedrofvteixeira: Why are we _"generifying"_ the exceptions caught ?

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/FileResourceGeneratedContentForSchedule.java
```diff
@@ -0,0 +1,124 @@
...
+import javax.ws.rs.core.MediaType;
+import java.util.List;
+
+/**
+ * This service provides methods for listing, creating, downloading, uploading, and removal of files.
+ *
+ * @author aaron
+ */
+@Path ( "/repo/files/generatedContentForSchedule" )
+public class FileResourceGeneratedContentForSchedule extends AbstractJaxRSResource {
```
pedrofvteixeira: If I'm understanding correctly, that line in `pentahoServices.spring.xml` **must** be present, otherwise we lose access to the public-facing API of https://help.pentaho.com/Documentation/8.1/Developer_Center/REST_API/File_Management/0O0#section_143

Am I understanding it correctly ?

repository/src/main/java/org/pentaho/platform/repository2/unified/fs/FileSystemRepositoryFileDao.java
```diff
@@ -225,7 +228,10 @@ public RepositoryFile internalGetFile( File f ) {
       String jcrPath = f.getAbsolutePath().substring( rootDir.getAbsolutePath().length() );
       if ( jcrPath.length() == 0 ) {
         jcrPath = "/";
+      } else if ( IS_WINDOWS ) {
+        jcrPath = jcrPath.replaceAll( "\\\\", "/" );
       }
```
pedrofvteixeira: We might want to consider using https://docs.oracle.com/javase/7/docs/api/java/io/File.html#separator instead of the string "/"

amartins23: Is the repository using the system path convention or always using "/" as the path separator? A cursory look at the code seems to point to the second case. Note that this is converting the _escaped_ windows path separator to the repository path representation.

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/FileResourceGeneratedContentForSchedule.java
amartins23: Yes

repository/src/main/java/org/pentaho/platform/repository2/unified/fs/FileSystemRepositoryFileDao.java
pedrofvteixeira: Makes sense. In that case, a couple of - better - approaches: 

- Use `RepositoryFile.SEPARATOR` instead of explicit `"/"` statements. 
  - And yes, our repository uses the unix-type separator as the convention
  - https://github.com/pentaho/pentaho-platform/blob/8.1.0.0-R/api/src/main/java/org/pentaho/platform/api/repository2/unified/RepositoryFile.java#L43
- Use `FilenameUtils.separatorsToUnix( jcrPath )`
  - https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FilenameUtils.html#separatorsToUnix(java.lang.String)
- Use `FilenameUtils.normalizeNoEndSeparator( jcrPath, true /* useUnixSeparator */ )`
  - https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FilenameUtils.html#normalizeNoEndSeparator(java.lang.String,%20boolean)


My $0.02: I'd steer away from any form of custom path separator handling.

assemblies/pentaho-solutions/src/main/resources/pentaho-solutions/system/pentahoServices.spring.xml
nantunes: No, it will not fail gracefully. The server will boot up with no errors, but the endpoint `/repo/files/generatedContentForSchedule` will not be available.

The separation was done so that we can use `FileResource` services in the "slim" CDE assembly, where scheduling and related libs were removed. If this change can't be performed automatically by the upgrade script, I guess it is odd to explain to clients why the change is needed... Your call.

extensions/src/main/java/org/pentaho/platform/web/http/ConfigurationAdminNonOsgiProxy.java
nantunes: When the OSGi _"world"_ isn't present (basically when `osgiListener`/`org.pentaho.platform.osgi.OSGIBoot` is removed from `systemListeners.xml`), a `RuntimeException` is thrown (I really can't recall at this moment if it is a generic `RuntimeException` or a specific one). I decided to make the catch clause the most comprehensive, because it seems to me that was the intention of the original author of the code, but I can try to be more specific.

pedrofvteixeira: The way the code is set nowadays, this `getProperties()` returns `null` when faced with one of those 3 exceptions. Any others are propagated. 

We should at least check to see if on all places that reference this method, there happens to be some custom exception catching + handling  taking place.

By changing this to catch all exceptions, we may effectively be changing an existing behaviour ( mind you, I'm not saying the way the behaviour stands nowadays is "right" or "wrong", just stating a fact ). 

If so, then those places may need changing as well, as they would never get exceptions moving forward.

nantunes: We're confident: the code was done recently by @pentaho/millenniumfalcon and is only used by `PentahoWebContextFilter`, where it expects no exceptions.

assemblies/pentaho-solutions/src/main/resources/pentaho-solutions/system/pentahoServices.spring.xml
nantunes: That endpoint was introduced in https://github.com/pentaho/pentaho-platform/pull/1970 for no apparent reason. I've found no usages of it, so we decided to remove it.

repository/src/main/java/org/pentaho/platform/repository2/unified/fs/FileSystemRepositoryFileDao.java
```diff
@@ -225,7 +228,10 @@ public RepositoryFile internalGetFile( File f ) {
       String jcrPath = f.getAbsolutePath().substring( rootDir.getAbsolutePath().length() );
       if ( jcrPath.length() == 0 ) {
         jcrPath = "/";
+      } else if ( IS_WINDOWS ) {
+        jcrPath = FilenameUtils.separatorsToUnix( jcrPath );
```
amartins23: This won't work for our purposes, an *escaped* path that looks like \\home\\admin\\file.wcdf, gets transformed into //home//admin//file.wcdf which is not what was intended.

graimundo:
I agree with @nantunes that this merge should not be squashed. The history of the individual commits is relevant.

PR #4215, opened 2018-08-10
[ANALYZER-3545] Unable to Define a Hyperlink to an Analyzer report sa…

LeonardoCoelho71950:
…ved in a folder having a colon (:)

Previously, the link to ask for analysis report's parameters was something like:
http://localhost:8080/pentaho/api/repos/:public:Steel%20Wheels:Widget%20Library:Analysis%20Views:Country%20Orders:test.xanalyzer/parameter
In other words, the '/' characters of the file path were substituted by ':' characters. Thus, if the filename had a ':', it would represent an issue. To handle this, the correct file path was introduced in the request body.

Nevertheless, it would be a good idea to use a numerical file identifier instead of the file path in these cases.

More code: https://github.com/pentaho/pentaho-analyzer/pull/1841

@pamval @duarteteixeira
@pentaho-lmartins @ricardosilva88


pamval:
See comment on the related pr (https://github.com/pentaho/pentaho-analyzer/pull/1841). I don't think changing this endpoint is the right way to fix the analyzer issue.

LeonardoCoelho71950:
Because of PR(pentaho/pentaho-analyzer#1841) changes, now this PR is irrelevant. 
@pamval

PR #4221, opened 2018-08-24
[BACKLOG-25314] Schedule perspective and Schedule option is displayed…

ddiroma:
… for the roles without schedule or Administer Security permissions


ddiroma:
@mdamour1976 
@pentaho/r2d2

PR #4232, opened 2018-09-12
[BACKLOG-25403] Add websocket capabilities to platform plugins

joaormf:


joaormf:
@pentaho/kashyyyk please review.

This PR needs to be merged before others can be created because of dependencies introduced.

PR #4237, opened 2018-09-20
[PPP-4160] Import-Export tool requires plain-text password

pentaho-lmartins:
The import/export util was using `KettleTwoWayPasswordEncoder` directly when it should use `Encr` so that high strength encryption plugins can be used, namely the `AESTwoWayPasswordEncoder`.

@pedrofvteixeira @ricardosilva88 @cravobranco @mbatchelor


pentaho-lmartins:
Added test cases.

pentaho-lmartins:
Updated PR to fix broken test case.

PR #4241, opened 2018-09-21
[PPP-4191] - Potential performance issue with OSGIObjectFactory

bantonio82:
@mbatchelor @pentaho-lmartins 

This pull request contains two commits, one for the fix and the other for checkstyle.

### DIFF
--- a/core/src/main/java/org/pentaho/platform/engine/core/system/objfac/OSGIObjectFactory.java
+++ b/core/src/main/java/org/pentaho/platform/engine/core/system/objfac/OSGIObjectFactory.java
@@ -76,21 +76,21 @@ public class OSGIObjectFactory implements IPentahoObjectFactory {
 
   @Override
   public <T> T get( Class<T> interfaceClass, IPentahoSession session, Map<String, String> properties )
-      throws ObjectFactoryException {
+    throws ObjectFactoryException {
     if ( isBundleContextValid() == false ) {
       return null;
     }
 
     try {
-      Map<String, String> props = new HashMap<String, String>( );
-      if( properties != null ){
-        props.putAll( properties);
+      Map<String, String> props = new HashMap<String, String>();
+      if ( properties != null ) {
+        props.putAll( properties );
       }
       props.put( REFERENCE_CLASS, interfaceClass.getName() );
       Collection<ServiceReference<IPentahoObjectReference>> serviceReferences = this.context
-          .getServiceReferences( IPentahoObjectReference.class, OSGIUtils.createFilter( props ) );
+        .getServiceReferences( IPentahoObjectReference.class, OSGIUtils.createFilter( props ) );
 
-      if( serviceReferences != null && serviceReferences.size() > 0 ){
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/pentaho/platform/engine/core/system/objfac/OSGIObjectFactory.java
```diff
@@ -257,9 +258,9 @@ public void init( String configFile, Object context ) {
         @Override public int compare( ServiceReference o1, ServiceReference o2 ) {
```
pentaho-lmartins: I'm not sure we need to implement a comparator since the `ServiceReference` class implements the method `compareTo()` which does exactly what we're doing here.

See: https://osgi.org/javadoc/r4v43/core/org/osgi/framework/ServiceReference.html#compareTo(java.lang.Object)

> If this ServiceReference and the specified ServiceReference have the same service id they are equal. This ServiceReference is less than the specified ServiceReference if it has a lower service ranking and greater if it has a higher service ranking. Otherwise, if this ServiceReference and the specified ServiceReference have the same service ranking, this ServiceReference is less than the specified ServiceReference if it has a higher service id and greater if it has a lower service id.

If we need to implement a comparator, the following code will be a better optimization.

```
Object ranking = this.getProperty(Constants.SERVICE_RANKING);
// if the property is not supplied or of incorrect type, use the default
int rank1 = ((ranking != null && ranking instanceof Integer) ? ((Integer) ranking).intValue() : 0);
ranking = other.getProperty(Constants.SERVICE_RANKING);
int rank2 = ((ranking != null && ranking instanceof Integer) ? ((Integer) ranking).intValue() : 0);

return rank1 - rank2;
```

core/src/test/java/org/pentaho/platform/engine/core/system/objfac/OSGIObjectFactoryTest.java
```diff
@@ -405,6 +408,44 @@ public void testGetObjectReferencesWithQuery() throws Exception {
...
+    ArrayList<ServiceReference<String>> refs = Stream.of( ref1, ref2, ref3 )
+      .collect( Collectors.toCollection( ArrayList::new ) );
+
+    when( mockContext.getServiceReferences( String.class, null ) ).thenReturn( refs );
+    when( mockContext.getService( ref1 ) ).thenReturn( "ref1" );
+    when( mockContext.getService( ref2 ) ).thenReturn( "ref2" );
+    when( mockContext.getService( ref3 ) ).thenReturn( "ref3" );
+
+    //props - all with different values
+    when( ref1.getProperty( Constants.SERVICE_RANKING ) ).thenReturn( "3" );
```
pentaho-lmartins: `getProperty( Constants.SERVICE_RANKING )` will return an `Integer` or `null`, not a `String`.
See: https://osgi.org/javadoc/r4v43/core/org/osgi/framework/Constants.html#SERVICE_RANKING

> The value of this property must be of type Integer.

PR #4252, opened 2018-09-25
[SP-4653] Backport of PPP-4160 - Import-Export tool requires plain-te…

pentaho-lmartins:
…xt password (7.1 Suite)

Cherry pick of #4237 to 7.1 branch.

Because `KettleClientEnvironment.reset()` doesn't exist in 7.1 and backporting it would require adding much more code just for the tests to run properly, a different approach was taken.

The tests for the different encryption types were split into different test suites so it isn't required to reset the Ketlle environment between each test.

@ssamora @bantonio82


pentaho-lmartins:
@ssamora and @bantonio82, PR was updated and passes Wingman validations other than coverage.

PR #4265, opened 2018-10-12
[PPP-4201] KarafBoot ignoring potential failures in various method calls

ssamora:
@pentaho-lmartins

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/osgi/KarafBoot.java
+++ b/extensions/src/main/java/org/pentaho/platform/osgi/KarafBoot.java
@@ -196,7 +196,14 @@ public class KarafBoot implements IPentahoSystemListener {
               return false;
             } else if ( symlinks.contains( relativePath ) ) {
               File linkFile = new File( destDir, relativePath );
-              linkFile.getParentFile().mkdirs();
+              try {
+                boolean linkFileDirCreated = linkFile.getParentFile().mkdirs();
+                logger.info(
+                  "link file " + linkFile.getParentFile().getAbsolutePath() + ( linkFileDirCreated ? "created" : "already existed" ) );
+              } catch ( SecurityException exception ) {
+                logger.error( linkFile.getParentFile().getAbsolutePath() + " Access denied." );
+                throw exception;
+              }
               Path link = Paths.get( linkFile.toURI() );
               try {
                 // Try to create a symlink and skip the copy if successful
@@ -204,8 +211,7 @@ public class KarafBoot implements IPentahoSystemListener {
                   return false;
                 }
               } catch ( IOException e ) {
-                logger
-                  .warn( "Unable to create symlink " + linkFile.getAbsolutePath() + " -> " + file.getAbsolutePath() );
+                logger.warn( "Unable to create symlink " + linkFile.getAbsolutePath() + " -> " + file.getAbsolutePath() );
               }
             }
             return true;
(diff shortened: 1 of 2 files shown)


extensions/src/main/java/org/pentaho/platform/osgi/KarafBoot.java
```diff
@@ -196,7 +196,8 @@
               return false;
             } else if ( symlinks.contains( relativePath ) ) {
               File linkFile = new File( destDir, relativePath );
-              linkFile.getParentFile().mkdirs();
+              boolean linkFileDirCreated = linkFile.getParentFile().mkdirs();
```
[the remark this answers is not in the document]
ssamora: Yes, you are correct. But i think we need both. the result may be false if it fails for some other reason (for example the directory already exists) and would not cause the SecurityException. i will update the pull request with the SecurityExcetion also.

extensions/src/main/java/org/pentaho/platform/osgi/KarafBoot.java
```diff
@@ -196,7 +196,13 @@
               return false;
             } else if ( symlinks.contains( relativePath ) ) {
               File linkFile = new File( destDir, relativePath );
-              linkFile.getParentFile().mkdirs();
+              try {
+                boolean linkFileDirCreated = linkFile.getParentFile().mkdirs();
+                logger.info(
+                  "link file " + linkFile.getAbsolutePath() + " parent directory " + ( linkFileDirCreated ? "created" : "already existed" ) );
+              } catch ( SecurityException exception ) {
+                logger.warn( linkFile.getAbsolutePath() + " cannot create directory. Access denied." );
```
pamval: If we can't create the directory, what should happen? Should execution be allowed to proceed as you have it here?
Is it better to stop execution and throw an exception because we don't know how to handle it?

extensions/src/main/java/org/pentaho/platform/osgi/KarafBoot.java
```diff
@@ -322,7 +328,13 @@ public static void deleteRecursiveIfExists( File item ) {
         deleteRecursiveIfExists( subitem );
       }
     }
-    item.delete();
+    try {
+      if ( !item.delete() ) {
+        LoggerFactory.getLogger( KarafBoot.class ).warn( item.toURI().toString() + " could not be deleted." );
+      }
+    } catch ( SecurityException exception ) {
+      LoggerFactory.getLogger( KarafBoot.class ).warn( item.toURI().toString() + " cannot delete file. Access denied." );
```
pamval: Same thing here. Catching an exception and logging the error is not handling the exception. What should happen if we can't delete the file? Can we safely continue executing?

extensions/src/test/java/org/pentaho/platform/osgi/KarafBootTest.java
```diff
@@ -382,6 +383,21 @@ public void testClearDataCacheSetting() throws Exception {
...
+
+    File dir = new File( "dir" );
+    dir.mkdir();
+    File file = new File( dir,  "file" );
+    try {
+      file.createNewFile();
+      KarafBoot.deleteRecursiveIfExists( dir );
+      assertNull( dir.listFiles() );
+    } catch ( IOException e ) {
+      e.printStackTrace();
```
mbatchelor: Should not have a printStackTrace here. If you get an IOException, shouldn't the test fail?

PR #4291, opened 2018-11-08
[BISERVER-14110] - Renaming a directory to a directory that already e…

ricardosilva88:
…xists results in the former directory being moved into the latter

asking review of @mdamour1976


repository/src/main/java/org/pentaho/platform/repository2/unified/jcr/JcrRepositoryFileDao.java
```diff
@@ -910,9 +910,8 @@ public Object doInJcr( final Session session ) throws RepositoryException, IOExc
                   .isPentahoFile( pentahoJcrConstants, destFileNode ) ), Messages.getInstance().getString(
                   "JcrRepositoryFileDao.ERROR_0002_CANNOT_OVERWRITE_FILE_WITH_FOLDER" ) ); //$NON-NLS-1$
           if ( JcrRepositoryFileUtils.isPentahoFolder( pentahoJcrConstants, destFileNode ) ) {
-            // existing item; caller is not renaming file, only moving it
-            appendFileName = true;
-            destParentFolderNode = destFileNode;
+            throw new IllegalArgumentException( Messages.getInstance().getString(
+              "JcrRepositoryFileDao.ERROR_0003_ILLEGAL_DEST_PATH" ) );
```
mbatchelor: Seems like it could be the right thing to do, but have you confirmed with the changes, you can still move a folder into another folder ?

ricardosilva88: Sure, I always do. Testing moving folder, sub folders with different name scenarios, tried to test all scenarios that I was capable of remember.
With this change, remaining a folder to a existing one we'll get the same error as we get when remaining a file to a existing one.

PR #4294, opened 2018-11-14
[BACKLOG-26573] Have filename be used when downloading exported files

DFieldFL:


extensions/src/main/java/org/pentaho/platform/web/http/api/resources/GeneratorStreamingOutput.java
```diff
@@ -36,11 +36,11 @@
 import org.pentaho.platform.util.messages.LocaleHelper;
 import org.pentaho.platform.util.web.MimeHelper;
 import org.pentaho.platform.util.web.SimpleUrlFactory;
-import org.pentaho.platform.web.http.HttpMimeTypeListener;
 import org.pentaho.platform.web.http.HttpOutputHandler;
 import org.pentaho.platform.web.http.api.resources.GeneratorStreamingOutputProvider.MimeTypeCallback;
 import org.pentaho.platform.web.http.request.HttpRequestParameterProvider;
 import org.pentaho.platform.web.http.session.HttpSessionParameterProvider;
+import org.pentaho.platform.web.servlet.HttpMimeTypeListener;
```
dcleao: Why, this new import?

DFieldFL: The old HttpMimeTypeListener is deprecated and in the deprecated code it references this class anyway. Heres a link the deprecated class using this class https://github.com/pentaho/pentaho-platform/blob/master/extensions/src/main/java/org/pentaho/platform/web/http/HttpMimeTypeListener.java#L41

PR #4299, opened 2018-11-23
[BACKLOG-26593] - CLONE: Ability to Disable user folder creation under Home directory in PUC

vflsa:
It has been creaed a new property in the file system.properties, named "hideUserHomeFolderOnCreate", which by default it is set to "false". To activate the feature in this story, it is necessary to change it to "true" and restart the server.
When a new user is added/created (both through PUC interface or through Spoon interface), depending on the value of the above mentioned property, the user home folder is automatically hidden or not (hidden if the property's value is set to true, not if the property's value is set to false or any other value).

### DIFF
--- a/repository/src/main/java/org/pentaho/platform/security/userroledao/jackrabbit/AbstractJcrBackedUserRoleDao.java
+++ b/repository/src/main/java/org/pentaho/platform/security/userroledao/jackrabbit/AbstractJcrBackedUserRoleDao.java
@@ -34,6 +34,7 @@ import org.apache.jackrabbit.core.security.user.PentahoUserManagerImpl;
 import org.apache.jackrabbit.spi.Name;
 import org.apache.jackrabbit.spi.NameFactory;
 import org.apache.jackrabbit.spi.commons.name.NameFactoryImpl;
+import org.pentaho.platform.api.engine.ISystemConfig;
 import org.pentaho.platform.api.engine.security.userroledao.IPentahoRole;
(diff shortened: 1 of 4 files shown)


repository/src/main/java/org/pentaho/platform/security/userroledao/jackrabbit/AbstractJcrBackedUserRoleDao.java
```diff
@@ -841,9 +842,13 @@ private RepositoryFile createUserHomeFolder( ITenant theTenant, String username,
           JcrRepositoryFileUtils.getFileByAbsolutePath( session, ServerRepositoryPaths.getUserHomeFolderPath(
               theTenant, username ), pathConversionHelper, lockHelper, false, null );
       if ( userHomeFolder == null ) {
+        RepositoryFile.Builder newFolder = new RepositoryFile.Builder( username ).folder( true );
+        Boolean hideUserHomeFolder = "true".equals(
+          PentahoSystem.get( ISystemConfig.class ).getProperty( "system.hideUserHomeFolderOnCreate" ) );
```
amartins23: Maybe add a "toLowerCase()" to the value obtained from getProperty, if not null, and create a constant to hold the property name.

vflsa: ok

PR #4310, opened 2018-12-12
[BISERVER-14127] New data source file type - error message with allow…

smmribeiro:
…ed types of files

@pentaho-lmartins @pedrofvteixeira

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/web/servlet/UploadFileUtils.java
+++ b/extensions/src/main/java/org/pentaho/platform/web/servlet/UploadFileUtils.java
@@ -23,7 +23,9 @@ package org.pentaho.platform.web.servlet;
 import com.ice.tar.TarEntry;
 import com.ice.tar.TarInputStream;
 import org.apache.commons.fileupload.FileItem;
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.IOUtils;
+import org.pentaho.di.core.util.StringUtil;
 import org.pentaho.platform.api.engine.IPentahoSession;
 import org.pentaho.platform.api.util.ITempFileDeleter;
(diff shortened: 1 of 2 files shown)


smmribeiro:
@pentaho-lmartins, I've cleaned the PR.

extensions/src/main/java/org/pentaho/platform/web/servlet/UploadFileUtils.java
```diff
@@ -101,51 +104,47 @@ public UploadFileUtils( IPentahoSession sessionValue ) {
...
+   * <p>A <code>null</code> is an invalid extension; to check if a NoExtension is allowed, an empty string must be
+   * given.</p>
+   *
+   * @param extension   the extension to check
+   * @param emitMessage if a message is to be written when an error occurs
+   * @return <code>true</code> if the given extension is supported, <code>false</code> if it's an invalid
+   * or unsupported extension
+   * @throws IOException
+   */
+  protected boolean checkExtension( String extension, boolean emitMessage ) throws IOException {
```
pamval: Although unlikely, did you check if there are classes subclassing this one? You changed the input parameter for this method and that might eventually break subclasses that call it.

I know it's very unlikely that someone subclassed this particular class, but you never know...

smmribeiro: I'll check it.

smmribeiro: Looking in our code for references to this class, we have the actual class and the test class on 'pentaho-platform' project and then we have a usage in 'data-access' project.

As there's no necessity to have this method accessible outside, I think this specific method is only 'protected' for unit testing purposes.

I think it's safe to leave this change to the method pass :-)

PR #4315, opened 2018-12-20
[BISERVER-14123] - Schedule - select folder for generated files is no…

ricardosilva88:
…t refreshed

This PR is a continuation of (https://github.com/pentaho/pentaho-commons-gwt-modules/pull/703) that was already merged.

Key points (detailed description on JIRA - https://jira.pentaho.com/browse/BISERVER-14123):
- Creation of a scheduler using folder /home/admin/test1;
- Delete folder test1 after scheduler creation;
- Next time a scheduler is being created, the dialog window is initiated with last scheduler folder location. The server will be inquired regarding /home/admin/test1 folder, which was deleted already, throwing an unhandled exception to server console and logs (javax.jcr.PathNotFoundException).  The scheduler dialog window has already follow up methods in place for dealing with errors coming from server side, which will do a new request to server using parent folder /home/admin/, which will work perfectly because this one exists.

The UI component was correct already, the problem was an ugly exception being displayed on server console/log and 500 http code as response of http://localhost:8080/pentaho/api/repo/files/[PATH]/tree?depth=1.


ricardosilva88:
@pamval @mbatchelor @pentaho-lmartins

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/services/FileService.java
```diff
@@ -1353,11 +1353,15 @@ public RepositoryFileTreeDto doGetTree( String pathId, Integer depth, String fil
     repositoryRequest.setIncludeAcls( includeAcls );
     repositoryRequest.setIncludeSystemFolders( includeSystemFolders );
 
-    RepositoryFileTreeDto tree = getRepoWs().getTreeFromRequest( repositoryRequest );
-
+    RepositoryFileTreeDto tree = null;
+    try {
+      tree = getRepoWs().getTreeFromRequest( repositoryRequest );
+    } catch ( Exception e ) {
+      logger.error( "Problem getting path: " + path );
```
pamval: This catch clause swallows any exception that might occur here.
I'm guessing that we want to catch a folder not found or something related - isn't there a more specific type of exception we can catch here?

At the very least we should log the exception. Right now, that information is lost.

ricardosilva88: I believe so, I'll try to change it for a more specific exception.

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/services/FileService.java
```diff
@@ -1353,11 +1354,15 @@ public RepositoryFileTreeDto doGetTree( String pathId, Integer depth, String fil
     repositoryRequest.setIncludeAcls( includeAcls );
     repositoryRequest.setIncludeSystemFolders( includeSystemFolders );
 
-    RepositoryFileTreeDto tree = getRepoWs().getTreeFromRequest( repositoryRequest );
-
+    RepositoryFileTreeDto tree = null;
+    try {
+      tree = getRepoWs().getTreeFromRequest( repositoryRequest );
+    } catch ( UnifiedRepositoryException e ) {
```
ricardosilva88: @pamval The most specific exception that I can catch here is the UnifiedRepositoryException instead of Exception.

The server console log will have the following line, with the root cause of this exception:
12:34:49,377 ERROR [FileService] org.springframework.dao.DataRetrievalFailureException: Path not found; nested exception is javax.jcr.PathNotFoundException: /pentaho/tenant0/home/admin/sched1

PR #4327, opened 2019-01-16
[BISERVER-14070] - Fixed Language Translations Properties Files not p…

RPAraujo:
…icked up on Xactions running on Server

@pentaho/tatooine @mbatchelor @pedrofvteixeira please review


extensions/src/main/java/org/pentaho/platform/plugin/services/importer/LocaleImportHandler.java
```diff
@@ -77,6 +78,11 @@ public LocaleImportHandler( List<IMimeType> mimeTypes, List<String> artifacts )
     this.artifacts = artifacts;
   }
 
+  // This method is used for testing purposes
+  protected void setRepositoryForTesting( IUnifiedRepository repository ) {
```
pentaho-lmartins: Don't use a modifier here, `protected` means this method can be accessed in a subclass.
Also, add the attribute `@VisibleForTesting`.

extensions/src/main/java/org/pentaho/platform/plugin/services/importer/LocaleImportHandler.java
```diff
@@ -170,7 +176,7 @@ private String extractLocaleCode( RepositoryFileImportBundle localeBundle ) {
     return localeCode;
   }
 
-  private RepositoryFile getLocaleParent( RepositoryFileImportBundle locale, Properties localePropertiesFromIndex ) {
+  protected RepositoryFile getLocaleParent( RepositoryFileImportBundle locale, Properties localePropertiesFromIndex ) {
```
pentaho-lmartins: Same here, just remove the access modifier and add the attribute.

RPAraujo: Done

extensions/src/main/java/org/pentaho/platform/plugin/services/importer/LocaleImportHandler.java
RPAraujo: Done

PR #4329, opened 2019-01-18
[BI-14087] Localized file/folder names are not shown in the Save As window

gryphendowre:
@pentaho/tatooine @pentaho-lmartins @ricardosilva88


extensions/src/main/java/org/pentaho/platform/web/http/api/resources/FileResource.java
```diff
@@ -938,9 +938,39 @@ public Response doSetContentCreator( @PathParam ( "pathId" ) String pathId, Repo
...
+        }
+      }
       fileService.doSetLocaleProperties( pathId, locale, properties );
-      return Response.ok().build();
+      if ( isFileTitleFound && fileTitle != null && !pathSplit[ pathSplit.length - 1 ].equals( fileTitle.getValue() ) ) {
+        success = fileService.doRename( pathId, fileTitle.getValue() );
+      }
+      // We use a similar response to doRename, the default here before was to just send an OK Response regardless.
+      if ( success ) {
+        return buildOkResponse();
```
pentaho-lmartins: You shouldn't send a response with a HTTP status code of 200 without content.
For this specific case, send a `noContent()` response which sends a 204 response code. This is the correct way to tell: success but no content. Some HTTP clients will complain when they receive a 200 response without content since there is the possibility that something went wrong and the content was lost.

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/FileResource.java
```diff
@@ -938,9 +938,39 @@ public Response doSetContentCreator( @PathParam ( "pathId" ) String pathId, Repo
...
+       renames. We need to keep it consistent.
+      */
+      String[] pathSplit = pathId.split( ":" );
+      boolean isFileTitleFound = false;
+      StringKeyStringValueDto fileTitle = null;
+      for ( StringKeyStringValueDto property : properties ) {
+        // Check to see if the file.title is listed here, it may be used to rename the file/folder
+        if ( property.getKey().equals( "file.title" ) ) {
+          // Check to see if the file title is different than the current file/folder, if it is, we need to move it.
+          responseMsg = "Unable to rename File/Folder";
```
pentaho-lmartins: This response message should be in a properties file so it can be translated.

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/FileResource.java
```diff
@@ -938,9 +938,39 @@ public Response doSetContentCreator( @PathParam ( "pathId" ) String pathId, Repo
...
+    boolean success = true;
+    String responseMsg = "";
     try {
+      /*
+       [BI-14087] If one of the properties is to change the file/folder name, it is also necessary to run this command to
+       get the file's path to update as well. We need to remove it from the properties, because it can affect future
+       renames. We need to keep it consistent.
+      */
+      String[] pathSplit = pathId.split( ":" );
+      boolean isFileTitleFound = false;
```
pentaho-lmartins: Why do we need this boolean flag? If the file title isn't found this variable will be false and `fileTitle` will be null.
Validation can be done by evaluating if `fileTitle` is null or not.

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/FileResource.java
gryphendowre: Added message to messages.properties

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/FileResource.java
gryphendowre: Will update this to use noContent, will also need to add a @ResponseCode to the method header. What should we do with the other REST calls in this file that use buildOkResponse() without a message return?  I see this file use 200 for other PUT methods.

PR #4341, opened 2019-02-05
[PPP-4223] - Use of vulnerable component spring-core before 4.3.14, s…

ricardosilva88:
**DO NOT MERGE WITHOUT THE REST OF PRs:**
[pentaho/maven-parent-poms#95 (comment)](https://github.com/pentaho/maven-parent-poms/pull/95#issue-250486125)

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -52,7 +52,6 @@
       <scope>compile</scope>
     </dependency>
 
-    <!-- Needed for runtime testing -->
     <dependency>
       <groupId>org.springframework.security</groupId>
       <artifactId>spring-security-core</artifactId>
@@ -73,7 +72,6 @@
         </exclusion>
       </exclusions>
     </dependency>
-    <!-- -->
 
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -51,11 +51,11 @@
       <artifactId>dom4j</artifactId>
       <scope>compile</scope>
     </dependency>
+
+    <!-- Needed for runtime testing -->
     <dependency>
       <groupId>org.springframework.security</groupId>
       <artifactId>spring-security-core</artifactId>
-      <version>${spring-security.version}</version>  <!-- see maven-parent-pom -->
-      <scope>compile</scope>
       <exclusions>
         <exclusion>
           <artifactId>*</artifactId>
@@ -66,8 +66,6 @@
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-beans</artifactId>
-      <version>${spring.framework.version}</version>
-      <scope>compile</scope>
       <exclusions>
         <exclusion>
           <artifactId>*</artifactId>
@@ -75,6 +73,8 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <!-- -->
+
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
--- a/assemblies/pentaho-war/pom.xml
+++ b/assemblies/pentaho-war/pom.xml
@@ -49,6 +49,13 @@
     </dependency>
     <!-- endregion -->
 
+    <!-- Needed for runtime -->
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-context-support</artifactId>
+    </dependency>
+    <!-- -->
+
     <dependency>
       <groupId>pentaho</groupId>
       <artifactId>pentaho-platform-extensions</artifactId>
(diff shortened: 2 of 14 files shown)


api/pom.xml
```diff
@@ -51,11 +51,11 @@
       <artifactId>dom4j</artifactId>
       <scope>compile</scope>
     </dependency>
+
+    <!-- Needed for runtime testing -->
     <dependency>
       <groupId>org.springframework.security</groupId>
       <artifactId>spring-security-core</artifactId>
-      <version>${spring-security.version}</version>  <!-- see maven-parent-pom -->
-      <scope>compile</scope>
```
graimundo: If this artifact is just needed during tests (as mentioned in the comment https://github.com/pentaho/pentaho-platform/pull/4341/files#diff-8658bb96fff6ac848cd28d3242fce067R55 ) then scope should be test `<scope>test</scope>`.

api/pom.xml
```diff
@@ -66,15 +66,15 @@
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-beans</artifactId>
-      <version>${spring.framework.version}</version>
-      <scope>compile</scope>
```
graimundo: If this artifact is just needed during tests (as mentioned in the comment https://github.com/pentaho/pentaho-platform/pull/4341/files#diff-8658bb96fff6ac848cd28d3242fce067R55 ) then scope should be test `<scope>test</scope>`.

assemblies/pentaho-war/pom.xml
```diff
@@ -50,6 +50,13 @@
     </dependency>
     <!-- endregion -->
 
+    <!-- Needed for runtime -->
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-context-support</artifactId>
```
graimundo: If this artifact is needed at runtime then scope should be runtime `<scope>runtime</scope>`.
This artifact is for runtime correct? Not tests?

core/pom.xml
```diff
@@ -296,21 +285,25 @@
       </exclusions>
     </dependency>
     <dependency>
-      <groupId>org.springframework.security</groupId>
-      <artifactId>spring-security-acl</artifactId>
-      <version>${spring-security.version}</version>  <!-- see maven-parent-pom -->
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-aop</artifactId>
       <scope>compile</scope>
```
graimundo: If this artifact is just needed during tests then scope should be test `<scope>test</scope>`.

core/pom.xml
```diff
@@ -296,21 +285,25 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.security</groupId>
+      <artifactId>spring-security-core</artifactId>
       <exclusions>
```
graimundo: If this artifact is just needed during tests then scope should be test `<scope>test</scope>`.

core/pom.xml
```diff
@@ -296,21 +285,25 @@
...
         <exclusion>
           <artifactId>*</artifactId>
           <groupId>*</groupId>
         </exclusion>
       </exclusions>
     </dependency>
+    <!-- -->
```
graimundo: I'm assuming this `<!-- -->`marker, marks the end of `test` scope dependencies.
You can use Intellij Idea regions to integrate better with our development tools: https://youtrack.jetbrains.com/issue/IDEA-93649
Just use `<region RegionName -->` and `<!-- endregion -->` to have folding directly available in the IDE.

api/pom.xml
ricardosilva88: Nop, also needed for compilation...just checked

api/pom.xml
ricardosilva88: Nop, also needed for compilation...just checked

assemblies/pentaho-war/pom.xml
ricardosilva88: Yeap, just confirmed. I'll change it

core/pom.xml
ricardosilva88: It makes sense, I will change in all PR's to follow this format.

ricardosilva88:
Updated PR

core/pom.xml
ricardosilva88: Needed also for compilation, removed commentary

core/pom.xml
ricardosilva88: Needed also for compilation, removed commentary

api/pom.xml
ricardosilva88: If I put <scope>test</scope> it doesn't compile

PR #4344, opened 2019-02-07
[CDE-896] - Map Http Error Status into HttpServletResponse Error

pentaho-arua:
JAX-RS Response return objects do not trigger the "error state" in the HttpServletResponse. Forcing all HTTP Error Status into "sendError" enables the configuration of custom error pages in web.xml.


pentaho-arua:
@pentaho/bobafett @graimundo Please review

PR #4351, opened 2019-02-13
[CDE-896] - Added a check to guarantee that the Response object is still uncommitted before trying to access the status code.

pentaho-arua:


pentaho-arua:
@pentaho/bobafett Please review

PR #4373, opened 2019-02-26
[BISERVER-14135] No logs generated after running the "import" command…

ssamora:
… of import-export utility
@pentaho-lmartins


pentaho-lmartins:
Looks so much better!!!

PR #4374, opened 2019-02-27
[BISERVER-13674] - pro_audit table throws exception on folders and report names that exceed 200 characters

vflsa:


vflsa:

PR #4378, opened 2019-03-13
[BACKLOG-27566] Centralize SLF4J dependency

nantunes:
Needs https://github.com/pentaho/maven-parent-poms/pull/111

@graimundo

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -76,14 +76,6 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
-      <version>${slf4j-api.version}</version>
-      <scope>compile</scope>
-      <exclusions>
-        <exclusion>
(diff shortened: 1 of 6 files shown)


repository/src/test/resources/solution/system/data/hibernate.properties
```diff
@@ -1,17 +1,17 @@
-#HSQL Database Engine 1.8.0.5
```
graimundo: This file probably shouldn't be committed.

repository/src/test/resources/solution/system/data/hibernate.script
```diff
@@ -1,205 +1,205 @@
-CREATE SCHEMA PUBLIC AUTHORIZATION DBA
```
graimundo: This file probably shouldn't be committed.

PR #4396, opened 2019-04-03
[PPP-4223] - Adding dependencies for IT testing and fixing Schedule IT

ricardosilva88:
@graimundo 

2 commits, need squash merge

CC: @jmcrfp @pentaho-lmartins

### DIFF
--- a/scheduler/pom.xml
+++ b/scheduler/pom.xml
@@ -165,12 +165,6 @@
       <version>${javax.servlet-api.version}</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>pentaho</groupId>
-      <artifactId>pentaho-platform-repository</artifactId>
-      <version>9.0.0.0-SNAPSHOT</version>
-      <scope>test</scope>
-    </dependency>
     <!-- endregion -->
   </dependencies>
   <build>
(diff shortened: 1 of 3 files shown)


scheduler/pom.xml
```diff
@@ -148,6 +148,30 @@
...
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+      <version>${javax.servlet-api.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>pentaho</groupId>
+      <artifactId>pentaho-platform-repository</artifactId>
+      <version>9.0.0.0-SNAPSHOT</version>
```
graimundo: @ricardosilva88 Hard-coded version on 9.0 Snapshot? Most likely a copy paste error.
You probably want to use `pentaho-repository.version` instead.
https://github.com/pentaho/build-resources/blob/8.2.0.0/version.properties#L114

ricardosilva88: Some how, Intelij add this dependency automatically here when I was creating ScheduleAuthorizationPolicy.java...I am pretty sure that it isn't needed for integration testing. I am preparing to send a new commit, just making some testing to guarantee that it isn't needed at all.

PR #4404, opened 2019-04-10
[BISERVER-13511] At import time, if an etl plugin is missing, kjb get…

LeonardoCoelho71950:
…s hosed (maybe ktr too?)

@pentaho-lmartins @pentaho/tatooine 

João Gameiro is reviewing UX.


LeonardoCoelho71950:
Hugo Figueira gave the OK!

PR #4408, opened 2019-04-17
[BISERVER-14189] Restoring folders that contain an index.xml using im…

ssamora:
…port-export utility creates repository corruption

@pentaho-lmartins 
@pentaho/tatooine


pentaho-lmartins:
@RPAraujo can you take a look at this? Looks like your area of expertise.

PR #4409, opened 2019-04-19
[PPP-4394] Japanese character in the data source name causes error

smmribeiro:
@pentaho/tatooine @pentaho-lmartins

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/web/servlet/UploadFileUtils.java
+++ b/extensions/src/main/java/org/pentaho/platform/web/servlet/UploadFileUtils.java
@@ -58,6 +58,7 @@ public class UploadFileUtils {
   private static final String DEFAULT_EXTENSIONS = "csv,dat,txt,tar,zip,tgz,gz,gzip";
   public static final String DEFAULT_RELATIVE_UPLOAD_FILE_PATH = File.separatorChar
     + "system" + File.separatorChar + "metadata" + File.separatorChar + "csvfiles" + File.separatorChar;
+  private static final char DOT = '.';
   private static final String DOT_TMP = ".tmp";
   private String fileName;
   private boolean shouldUnzip;
@@ -191,7 +192,8 @@ public class UploadFileUtils {
     File file = null;
     if ( isTemporary() ) {
       // Use the full filename because GZip relies on the extensions of the file to discover it's content
-      String extension = ( getUploadedFileItem() != null ) ? '-' + getUploadedFileItem().getName() + DOT_TMP : DOT_TMP;
+      String extension =
+        ( getUploadedFileItem() != null ) ? DOT + removeFileName( getUploadedFileItem().getName() ) + DOT_TMP : DOT_TMP;
       file =
         PentahoSystem.getApplicationContext()
           .createTempFile( session, StringUtil.EMPTY_STRING, extension, true );
@@ -369,7 +371,7 @@ public class UploadFileUtils {
               if ( isTemporary() ) {
                 entryFile =
                   PentahoSystem.getApplicationContext()
-                    .createTempFile( session, StringUtil.EMPTY_STRING, '.' + extension + DOT_TMP, true );
+                    .createTempFile( session, StringUtil.EMPTY_STRING, DOT + extension + DOT_TMP, true );
               } else {
                 entryFile = new File( getPath() + File.separatorChar + entry.getName() );
               }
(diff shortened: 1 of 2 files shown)


extensions/src/main/java/org/pentaho/platform/web/servlet/UploadFileUtils.java
```diff
@@ -586,6 +588,32 @@ private long getFolderSize( File folder ) {
...
+   * <p>Removes the file name, leaving all extensions of a given filename.</p>
+   * <p>Returns <code>null</code> if filename is <code>null</code>, and an empty string if the filename is empty or has
+   * no extensions.</p>
+   *
+   * @param filename the filename for which we want the extensions
+   * @return all extensions of the given filename
+   */
+  protected String removeFileName( String filename ) {
+    if ( null != filename ) {
+      int dotIndex = filename.indexOf( DOT );
```
pentaho-lmartins: Shouldn't we be looking for the `lastIndexOf` the dot instead?

extensions/src/main/java/org/pentaho/platform/web/servlet/UploadFileUtils.java
```diff
@@ -586,6 +588,32 @@ private long getFolderSize( File folder ) {
...
+
+      if ( 0 < dotIndex ) {
+        int len = filename.length();
+        ++dotIndex;
+        if ( len > dotIndex ) {
+          return filename.substring( dotIndex );
```
pentaho-lmartins: I'd like to believe that `commons-io` would have some method to extract the extension of a file.

extensions/src/test/java/org/pentaho/platform/web/servlet/UploadFileUtilsTest.java
```diff
@@ -58,4 +60,16 @@ public void testCheckExtensions() throws Exception {
...
+  @Test
+  public void testGetAllExtensions() throws Exception {
+    UploadFileUtils testUtils = new UploadFileUtils( null );
+    assertNull( testUtils.removeFileName( null ) );
+    assertEquals( "",  testUtils.removeFileName( "" ) );
+    assertEquals( "",  testUtils.removeFileName( "aaaa" ) );
+    assertEquals( "",  testUtils.removeFileName( " bb " ) );
+    assertEquals( "a",  testUtils.removeFileName( "b.a" ) );
+    assertEquals( "b.a",  testUtils.removeFileName( "c.b.a" ) );
+    assertEquals( "c.b.a",  testUtils.removeFileName( "d.c.b.a" ) );
```
pentaho-lmartins: Shouldn't the extension be just `.a` instead?

extensions/src/main/java/org/pentaho/platform/web/servlet/UploadFileUtils.java
smmribeiro: To anwser your question above: if the filename is `luis.martins.2019.jpg`, the extension we want to get is `.martins.2019.jpg`, that's why I named the method "remove file name".

We need to keep the extensions because of the way `tar` files are handled (they rely on the extensions and not on some magic number/signature on the content itself like other supported file types).

Theoretically, we'd only need the last two extensions, however the cost of obtaining a maximum of two would be much greater than getting all of them. That's why I'm using `indexOf` and not `lastIndexOf`.

extensions/src/main/java/org/pentaho/platform/web/servlet/UploadFileUtils.java
smmribeiro: Yes and no, technically extension refers to the last block of text after the last dot (and for that there's the `getExtension` method of `common-io`.
There's, however, no method that returns "all" extensions as explained in the comment above.

extensions/src/main/java/org/pentaho/platform/web/servlet/UploadFileUtils.java
pentaho-lmartins: OK, we're trying to accommodate for `.tar.gz` files?
What if my file is `luis.martins.2019.tar.gz`? Will that work?

extensions/src/test/java/org/pentaho/platform/web/servlet/UploadFileUtilsTest.java
smmribeiro: No, we need ALL extensions!

extensions/src/main/java/org/pentaho/platform/web/servlet/UploadFileUtils.java
smmribeiro: The compressed files extensions supported here are: `.tar`, `.zip`, `.tgz`, `.gz` and `.gzip`.
Although it wouldn't be difficult to support `.tar.gz`, you'd have to change it to `.tgz`. 

So... if you'd asked about `luis.martins.2019.tgz` the answer would be Yes and No :-)

- **Yes**, the file would be correctly uploaded and the temp files would be created (named something like "_username-918273645._**martins.2019.tgz**_.tmp_" for the original file and "_username-192837465._**martins.2019**_.tmp_" for the "contained"/"_targazed_" file).

- **No**, as the "_targazed_" file has `.2019` extension, an error would occur in this particular process (as expected) as it's not in the supported ones (`.csv`, `.dat` and `.txt`).

PR #4435, opened 2019-05-22
[BISERVER-14235] - Output files getting generated twice in User Home …

marcoslarsen:
…directory when trying to write the Generated Output content to a Read Only Folder.

When the output folder differs from the one defined in the scheduler, the scheduler is updated, which consists in creating a new job with the updated output folder. 
What was happening was that when the job execution detected that the output folder differed, it updated the variable locally, marked the job for update and continued the execution. Then, returned an execution status marked has update and finished. This caused the creation of a new job that was executed producing a duplicate report. This fix prevents the first job from executing when it detects that the output folder differs, being only executed the updated job.


marcoslarsen:
@pentaho/kashyyyk Please review

PR #4441, opened 2019-05-29
[PDI-18043] - HTTP job entry step doesn't pass the username and passw…

cravobranco:
…ord that is required for authenticating to the REST API.

* Changed exceptionTranslationFilter for pattern "/api/repos/**"

@pentaho/bb-8 please review.


smmribeiro:
Shouldn't there be a similar PR for the "pentaho-platform-ee" repository?
If I'm not mistaken, "pentaho-platform" is for CE while EE uses "pentaho-platform-ee"!

pentaho-lmartins:
**This needs to be reverted. This PR causes a regression.**

Users requesting a report of dashboard with a direct link won't be redirected to the login page and instead will be presented with the browser's login prompt.

PR #4450, opened 2019-05-30
[BACKLOG-27345] - upgrading joda-time

kurtwalker:
aws-java-sdk-core depends on a later version, so bringing this to the latest


kurtwalker:
hold for now

PR #4451, opened 2019-05-31
[BISERVER-14220] Remove the 'locale' definition from 'bounceToReturnLocation()' 

befc:
- Remove the 'locale' definition from 'bounceToReturnLocation()' so that it aways defaults to the browser's locale
- Remove the 'locale' value assignment from the input form (as hidden value), since it was only used n the 'bounceToReturnLocation()' function


befc:

PR #4454, opened 2019-06-03
[BISERVER-14265]  Exporting XMI fails when localization files present.

mkambol:
Code was assuming the file type could be determined from the extension,
but at some point the filename was made into a uuid.
This change retrieves locale files and info explicitly using
PentahoMetadataInformationMap.getLocaleFiles rather than using naming.

https://jira.pentaho.com/browse/BISERVER-14265

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/plugin/services/metadata/PentahoMetadataDomainRepository.java
+++ b/extensions/src/main/java/org/pentaho/platform/plugin/services/metadata/PentahoMetadataDomainRepository.java
@@ -14,12 +14,13 @@
  * See the GNU Lesser General Public License for more details.
  *
  *
- * Copyright (c) 2002-2018 Hitachi Vantara. All rights reserved.
+ * Copyright (c) 2002-2019 Hitachi Vantara. All rights reserved.
  *
  */
 
 package org.pentaho.platform.plugin.services.metadata;
 
+import com.google.common.collect.ImmutableMap;
 import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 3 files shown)


extensions/src/main/java/org/pentaho/platform/plugin/services/metadata/PentahoMetadataDomainRepository.java
```diff
@@ -406,29 +412,35 @@ private boolean hasAccessFor( RepositoryFile repositoryFile ) {
...
-    Map<String, InputStream> values = new HashMap<String, InputStream>( metadataFiles.size() );
-    for ( RepositoryFile repoFile : metadataFiles ) {
-      RepositoryFileInputStream is;
-      try {
-        is = new RepositoryFileInputStream( repoFile );
-      } catch ( Exception e ) {
-        return null; // This pretty much ensures an exception will be thrown later and passed to the client
-      }
-      String fileName = repoFile.getName().endsWith( ".properties" ) ? repoFile.getName()
-        : domainId + ( domainId.endsWith( ".xmi" ) ? "" : ".xmi" );
```
kurtwalker: seems we've lost this part of logic related domainId ending in .xml.  do we still need it?

mkambol: From what I could see, domainId does consistently end with .xmi.

But you're right--  better safe than sorry.  I'll add a check.

PR #4455, opened 2019-06-03
[BISERVER-14231] Change PRPTImportHandler to respect ImportHandlerMimeTypeDefinitions.xml

pentaho-whartman:
@pentaho/rogueone 

https://jira.pentaho.com/browse/BISERVER-14231

Before this PR, the visibility of imported PRPTs in the repository was based on the `visible` attribute in the PRPT's metadata. This PR changes the `PRPTImportHandler` to leave this to the `RepositoryFileImportFileHandler` which pull from `ImportHandlerMimeTypeDefinitions.xml` instead.

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/plugin/services/importer/PRPTImportHandler.java
+++ b/extensions/src/main/java/org/pentaho/platform/plugin/services/importer/PRPTImportHandler.java
@@ -14,7 +14,7 @@
  * See the GNU Lesser General Public License for more details.
  *
  *
- * Copyright (c) 2002-2018 Hitachi Vantara. All rights reserved.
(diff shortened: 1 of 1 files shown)


befc:
@pentaho-whartman, please fix the copyright header and unused import detected by wingman

pentaho-whartman:
We discussed defaulting to the value on the PRPT metadata if nothing was defined in `ImportHandlerMimeTypeDefinitions.xml`, however a missing `hidden="true"` is implicitly treated as false (and this pattern is present in the default file) and a PRPT cannot be imported if it is not included in a definition.

Since there's no way for this to be both undefined and upload-able in PUC, there's no reason to fall back on the metadata.

PR #4457, opened 2019-06-04
[BISERVER-14265]  Exporting XMI fails when localization files present.

mkambol:
(#4454)

Code was assuming the file type could be determined from the extension,
but at some point the filename was made into a uuid.
This change retrieves locale files and info explicitly using
PentahoMetadataInformationMap.getLocaleFiles rather than using naming.

https://jira.pentaho.com/browse/BISERVER-14265

(cherry picked from commit a56dfca26ae8174f7112fac719c09361b9fc004e)


mkambol:
@kcruzada 

Change validated on 9.0.0.0-36

PR #4463, opened 2019-06-04
[BISERVER-14220] Remove the 'locale' definition from 'bounceToReturnL…

befc:
…ocation()' so that it aways defaults to the browser's locale

Cherry-picked from https://github.com/pentaho/pentaho-platform/pull/4451


befc:
@dkincade @kcruzada @pamval 

@pentaho/rogueone

PR #4487, opened 2019-06-21
[BISERVER-14231] FIX: Cannot hide Reports (.PRPT) by default when set…

smmribeiro:
…ting configuration Hidden=true on ImportHandlerMimeTypeDefinitions.xml using PUC Upload

We have to make sure the report should has its metadata overridden to the “hidden” value that comes from configuration before the actual import action.

@pentaho-whartman, please check if this fix does the trick or if there's something else missing.

@pentaho-whartman @pentaho/tatooine @ppatricio @pentaho-lmartins @ssamora

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/plugin/services/importer/PRPTImportHandler.java
+++ b/extensions/src/main/java/org/pentaho/platform/plugin/services/importer/PRPTImportHandler.java
@@ -57,7 +57,8 @@ public class PRPTImportHandler extends RepositoryFileImportFileHandler implement
     IPlatformImporter importer = PentahoSystem.get( IPlatformImporter.class );
     String fileName = importBundle.getName();
 
-    String filePath = ( importBundle.getPath().equals( "/" ) || importBundle.getPath().equals( "\\" ) ) ? "" : importBundle.getPath();
+    String filePath =
+      ( importBundle.getPath().equals( "/" ) || importBundle.getPath().equals( "\\" ) ) ? "" : importBundle.getPath();
 
     // If is locale file store it for later processing.
     // need to extract this from meta.xml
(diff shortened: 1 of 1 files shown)


extensions/src/main/java/org/pentaho/platform/plugin/services/importer/PRPTImportHandler.java
```diff
@@ -70,6 +71,10 @@ public void importFile( IPlatformImportBundle bundle ) throws PlatformImportExce
 
       DocumentMetaData documentMetaData = extractMetaData( bytes );
       fillLocaleEntry( localeFilesProcessor, documentMetaData, filePath, fileName, importBundle.getFile() );
+
+      // The hidden property to be used is the one that comes from configuration (see BISERVER-14231)
+      importBundle.setHidden( this.getMimeTypeMap().get( importBundle.getMimeType() ).isHidden() );
```
pentaho-whartman: I need to look into this a little more, but I'm confused as to why this is necessary.

Through the call to `super.importFile( importBundle )`, we should eventually should call `solutionHelper.isInHiddenList()` when we create the file, which checks the mime type configs.

I'm pretty sure this is redundant. I'll step through the code again to double check though.

smmribeiro: It doesn't reach that invocation:
`
private boolean isHiddenBundle( RepositoryFileImportBundle bundle ) {
    if ( bundle.isHidden() != null ) {
      return bundle.isHidden();
    }
    if ( solutionHelper.isInHiddenList( bundle.getName() ) ) {
      return true;
    }
    return RepositoryFile.HIDDEN_BY_DEFAULT;
  }
`

It enters on the first if...

pentaho-whartman: I just stepped through this (on the last 8.3 RC) and `bundle.isHidden()` came back `null`

How were you importing the prpt? the "upload" button in PUC, or something else?

smmribeiro: Hmmm... You probably tested with a report that never had the "hidden" property set!
I've tested with reports that had the hidden metadata e plicity to true and false...
You can have this behaviour setting the property before exporting the report.

smmribeiro: Thinking better on the 'isHiddenBundle' code... It seems that, existing, the property of the imported report is the one used... only if it's absent the one from confif is used!

pentaho-whartman: Yep, just checked with a report that had been exported as explicitly visible. Looks like you're right and this fixes it.

We didn't cover this case when we were testing the fix -- good catch. I'll follow up with our team's QA to cover this in the testrail/etc.

PR #4513, opened 2019-07-24
[BACKLOG-29696] Take into account when protection is disabled and token isn't created

davidmsantos90:
@pentaho/millenniumfalcon please review

Merge with: https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/1460


extensions/src/test/java/org/pentaho/platform/web/http/security/CsrfTokenResponseHeaderFilterTest.java
```diff
@@ -85,7 +85,6 @@ private static CsrfToken createToken() {
     return token;
   }
 
-  @Test( expected = ServletException.class )
   public void testWhenNoCsrfTokenThenThrow() throws Exception {
```
dcleao: This unit test needs to be renamed and changed

PR #4540, opened 2019-09-06
[MONDRIAN-1713] cellBatchSize should be included in mondrian.properties

bennychow:
@mkambol @kurtwalker @lucboudreau  Hi Guys, I did my best to try to document this existing property from an admin point of view trying to tune their Mondrian query performance.  Please let me know if you have any suggestions or if anything is inaccurate.  Thanks

### DIFF
--- a/assemblies/pentaho-solutions/src/main/resources/pentaho-solutions/system/mondrian/mondrian.properties
+++ b/assemblies/pentaho-solutions/src/main/resources/pentaho-solutions/system/mondrian/mondrian.properties
@@ -139,3 +139,39 @@ mondrian.spi.dataSourceResolverClass=org.pentaho.platform.web.servlet.PentahoDat
 # Maximum number of MDX query threads per Mondrian server instance.
 # Defaults to 20.
 mondrian.rolap.maxQueryThreads=20
+
+###############################################################################
+# Integer property which controls the number of cells that Mondrian will try to load
+# in a single batch (typically one or two Segment.load SQL statements per batch).
+# A single cell consists of the intersection of all dimensions members in a query 
+# including the measure member.
+# For example, if a query contains 5 measures, 1000 customers, 100 products 
+# and a filter on 10 dates with no calculated members,
+# then the cell count is 5*1000*100*10=5M.
+# In more general terms, cell count is equal to the multiplication of the 
+# number of tuples on each of the query axis.  
+# So, in the above example, suppose:
+# Row Axis = Customers * Products = 1000 * 100 = 100,000
+# Column Axis = Measures = 5
+# Slicer Axis = Dates = 10
+# Total Cell Count = 100,000 * 5 * 10 = 5M
+# If using a NonEmptyCrossJoin on the Row Axis, typically far fewer tuples will
+# be returned since not every Customer buys every Product.
+#
+# In the above example, if the batch size was set to 1M, then at most 
+# 5M Cells / 1M Cells/Batch = 5 batches of Segment.load SQL statements would be
+# needed to execute the query on cold cache.
+# Within a single batch, based on the cardinality of a dimension and the number of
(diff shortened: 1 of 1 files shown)


lucboudreau:
Hey Benny. Did you test the value of one million to see if it generally fits in the specs we advertise for the different BA server configurations?

bennychow:
I built two reports with 4M and 20M cell spaces and tried batch sizes of 100K, 500K, 750K, 1M, 1.5M and 2M.  1M seemed to give good performance.

lucboudreau:
I was thinking how much memory it uses per query and how many concurrent queries we can run with the default server specs.

bennychow:
I did a little more testing with the 20M cell space query and recorded the max total memory of the BA server Java process:

- batchsize: 250K - Timed Out (> 300 seconds) - 4.1GB
- batchsize: 500K - 280 seconds - 5.8 GB
- batchsize: 1M - 209 seconds - 6.5 GB

The baseline memory usage started at 3.1G after opening the foodmart datasource.
Our docs say 8GB minimum for the BA server:
https://help.pentaho.com/Documentation/8.3/Setup/Components_Reference
As a comparison, Tableau's minimum is 16GB for testing and 32GB for production:
https://help.tableau.com/current/server-linux/en-us/server_hardware_min.htm

I think it should be within our memory recommendation to set a batch size of 1M

Thoughts?

mkambol:
Thanks for documenting this, Benny.  Right now the bulk of docs for that property are littered across many ESRs.

1M seems like a good recommendation to me.  It might even make sense to bump our default up from 100K, which seems pretty low.  

For your docs, it might be good to mention that your "5*1000*100*10=5M" calculation is assuming non-calculated members.  Calculations can certainly increase the cell space.

lucboudreau:
Can't help but feel that 1M is too much. That limit means we can only run one big query at a time. So basically, if a customer creates a report with 1M cells, you can only run it once at a time, which is a bit limiting. If we had more overhead, I'd be fine with 1M cells, but this is not the case. I think we should try something more conservative to start, around 250k or 500k, and minimize the potential issues with the default behavior. If a customer wants to use more cells per query cycle, they can go back and augment it themselves.

mkambol:
Not sure if I follow.  cellBatchSize is used within fastbatchingcellreader to check whether the cell requests within a single MDX query exceed a threshold.  If multiple MDX queries are running, each would be checking the number of cell requests created just for that query against the cellBatchSize.

lucboudreau:
The limit affects each query separately. In this case, since the threshold allows a single query to use more than 50% of the free JVM memory overhead, it means you can never run this report more than once simultaneously without getting an OOME.

mkambol:
Ah, got it.  Good point.

bennychow:
@lboudreau-pentaho  I took my 20M cell space report and ran it concurrently on cold cache and got these results:
* batchsize: 250K - 9.28m - 6.7GB
* batchsize: 1M - 5.2m - 6.7GB

It appears that memory usage was about the same and query time was much better with the 1M batch size.  What might explain this behavior?

@mkambol Thanks for the suggestion about the assumption about non-calculated members.  I'll add that.

IMO, when people say Mondrian doesn't perform well on big data, it's because of the query times and not concurrency.  By setting a higher cell batch size, it'll really help with improving query times.  The default setting of 1M doesn't seem high considering this setting is probably used by a single person trying to POC Pentaho on a non-production machine.  Once you go into a production machine, I'm sure there will be much more RAM available and 1M would probably be a low setting then.

lucboudreau:
That's surprising. Looks like it maxes out and does not run into OOMEs. I guess 1M is good then.

bennychow:
@lucboudreau @mkambol @kurtwalker Could one of you guys merge this please?  Thanks!

PR #4617, opened 2020-01-14
[SP-5357] Backport of ANALYZER-3904 - Analyzer's Save and Save As dia…

ssamora:
…log are querying entire tree list of files and folders when using showRepositoryButtons=true on iFrame (8.3 Suite)

cherry-pick of 11b01ea and 8862d11


bmorrise:
Looks good to me

PR #4791, opened 2020-12-22
[BISERVER-13833] - createUser transmits password in clear text

bcostahitachivantara:
@pentaho-lmartins 
@ssamora 
@smmribeiro

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/UserRoleDaoResource.java
+++ b/extensions/src/main/java/org/pentaho/platform/web/http/api/resources/UserRoleDaoResource.java
@@ -144,6 +144,7 @@ public class UserRoleDaoResource extends AbstractJaxRSResource {
     } )
   public Response createUser( User user ) {
     try {
+      user.setPassword( b64DecodePassword( user.getPassword() ) );
       userRoleDaoService.createUser( user );
     } catch ( SecurityException e ) {
       throw new WebApplicationException( Response.Status.FORBIDDEN );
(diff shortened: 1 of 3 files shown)


buildguy:
### :white_check_mark: Build finished in 29m 21s
Build command:
```console
mvn clean verify -B -e -Daudit -amd -pl extensions,user-console
```
---
:ok_hand: All tests passed!
 
Tests run: **1548**, Failures: **0**, Skipped: **1** &nbsp;&nbsp; [Test Results](http://wingman.pentaho.net/job/projects/job/platform/job/PR-4791/1/testReport)

 ---
 
 

:information_source:  _This is an automatic message_

extensions/src/main/java/org/pentaho/platform/web/http/api/resources/UserRoleDaoResource.java
```diff
@@ -144,6 +144,7 @@ public UserRoleDaoResource( final IRoleAuthorizationPolicyRoleBindingDao roleBin
     } )
   public Response createUser( User user ) {
     try {
+      user.setPassword( b64DecodePassword( user.getPassword() ) );
```
ssamora, line 147: can't the user object received here already have the encoded password? 
if not, what i'm not getting is, if the we are transmitting the password in clear text, why are we decoding the password here? shouldn't we be encoding the password instead?

bcostahitachivantara: On file UserRolesAdminPanelController.java at line 95 it is create a json object with password encoded and then send the request, that request came to here, so it is necessary decode from base 64 to plain string. 
Later userRoleDaoService.createUser( user ) will create and user in database with a non-reversible encoded password

ssamora (approved):

PR #4842, opened 2021-04-15
[BISERVER-14628] - createUser and updatePassword don´t allowed in password the character '%'

bcostahitachivantara:
@ssamora 
@smmribeiro

PR #4845, opened 2021-04-16
[SP-5932] Backport of BISERVER-14628 - createUser and updatePassword …

ssamora:
…don´t allowed in password the character '%' (8.3 Suite)

Nor exactly a cherry-pick of 21012c0a852114e1b776e1384eeb063b9892a015 (original PR: https://github.com/pentaho/pentaho-platform/pull/4842 ) 
since in CreateUser method the decode is done in a different line. The change is the same nonetheless. 

@smmribeiro


bcostahitachivantara (approved):
Consistent with base bug, even with some changes

smmribeiro (approved):
Consistent with the original PR (differences are justified by the difference in base code).

PR #5053, opened 2022-03-22
[BACKLOG-35622] CSRF - Change Request Matcher elements to be case ins…

spravasahu:
…ensitive by default


dcleao (approved):

dcleao:
Depends on https://github.com/pentaho/maven-parent-poms/pull/332

PR #5085, opened 2022-05-24
[SP-6088] Backport of ANALYZER-3942 - Add ability to set the locale via ProxyTrustingFilter

dcleao:
@pentaho/sp-branch-write please review and merge.


smmribeiro (approved):
Consistent with the original PR (differences are justified by the difference in base code).

PR #5154, opened 2022-11-09
SP-6203 Backport of PPP-4761 - File Upload Path Disclosure exposed on Pentaho Server when file_name parameter uses special character (9.3 Suite)

andreramos89:
@smmribeiro 
@bcostahitachivantara 
@renato-s


smmribeiro (approved):
Looks good.

PR #5186, opened 2023-01-16
BACKLOG-36845: Focus Visibility (AA), Keyboard Navigation (A) for Use…

hv-vamsi:
…rs & Roles, Authentication in Administration

### DIFF
--- a/user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
+++ b/user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
@@ -14,12 +14,13 @@
  * See the GNU Lesser General Public License for more details.
  *
  *
- * Copyright (c) 2002-2018 Hitachi Vantara. All rights reserved.
+ * Copyright (c) 2002-2023 Hitachi Vantara. All rights reserved.
  *
  */
 
 package org.pentaho.mantle.client.admin;
 
+import com.google.gwt.aria.client.Roles;
 import com.google.gwt.user.client.ui.Button;
 import com.google.gwt.user.client.ui.HorizontalPanel;
 import com.google.gwt.user.client.ui.Label;
@@ -98,6 +99,7 @@ public class UserRolesAdminPanel extends SimplePanel {
   private Widget createUsersPanel() {
     HorizontalPanel mainUsersPanel = new HorizontalPanel();
     mainUsersPanel.getElement().setId( "admin-users-panel" );
+    Roles.getTabpanelRole().set( mainUsersPanel.getElement() );
 
     // mainUsersPanel.setWidth("376px");
     SimplePanel hSpacer = new SimplePanel();
@@ -234,6 +236,7 @@ public class UserRolesAdminPanel extends SimplePanel {
 
   private Widget createSystemRolesPanel() {
     HorizontalPanel mainSystemRolesPanel = new HorizontalPanel();
+    Roles.getTabpanelRole().set( mainSystemRolesPanel.getElement() );
     mainSystemRolesPanel.getElement().setId( "admin-system-roles-panel" );
     SimplePanel hSpacer = new SimplePanel();
     hSpacer.setWidth( "15px" );
@@ -282,6 +285,7 @@ public class UserRolesAdminPanel extends SimplePanel {
   private Widget createRolesPanel() {
 
     HorizontalPanel mainRolesPanel = new HorizontalPanel();
+    Roles.getTabpanelRole().set( mainRolesPanel.getElement() );
     mainRolesPanel.getElement().setId( "admin-roles-panel" );
     SimplePanel hSpacer = new SimplePanel();
     hSpacer.setWidth( "15px" );


user-console/src/main/java/org/pentaho/mantle/client/admin/UserDialog.java
```diff
@@ -62,11 +63,18 @@ public UserDialog( UserRolesAdminPanelController controller ) {
 
     acceptBtn.setEnabled( false );
     nameTextBox = new TextBox();
+    Roles.getTextboxRole().set( acceptBtn.getElement() );
+    nameTextBox.getElement().setTabIndex( 0 );
+    nameTextBox.setFocus( true );
```
dcleao, line 68: Is `nameTextBox` not the first focusable element in the user dialog? If so, the GwtDialog class shall already place the focus on it (changes on the way by @varuntangirala).

If not the first one, then setting focus explicitly in this way is imo not the best way. A declarative XUL attribute should exist for it. Also, the constructor is not always run as XUL instances are frequently reused...

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
```diff
@@ -120,12 +142,37 @@ private Widget createUsersPanel() {
...
     labelAndButtonsPanel.setCellWidth( hSpacer, "100%" );
-    newUserButton = new ThemeableImageButton( addButtonStyles, null, Messages.getString( "new" ) );
+    newUserButton = new ThemeableImageButton( addButtonStyles, null,"add new user"  ) {
+      @Override
+      public void onBrowserEvent(Event event) {
+        if ( event.getTypeInt() == Event.ONMOUSEUP || KeyCodes.KEY_ENTER == (char) event.getKeyCode() ) {
+          newUserOnKeyPress();
+        }
+      }
+    };
```
dcleao, line 152: As commented elsewhere, this could and imo should be handled by the `ThemeableImageButton` class, by forwarding the ENTER key to a click event.

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
```diff
@@ -120,12 +142,37 @@ private Widget createUsersPanel() {
...
+    Roles.getButtonRole().set( newUserButton.getElement() );
+    Roles.getButtonRole().setAriaLabelProperty( newUserButton.getElement(), Messages.getString("add" ) );
+    newUserButton.getElement().setTabIndex( 0 );
+    newUserButton.getElement().setId( "add-new-user" );
+    newUserButton.addStyleName( "focus-visible" );
```
dcleao, line 157: Also, all of this could and imo should be placed on the `ThemeableImageButton` class.

Exception is the aria label property, which could be provided by a new method on the `ThemeableImageButton` class.

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
```diff
@@ -186,6 +234,7 @@ private Widget createUsersPanel() {
     availableRolesPanel.add( availableRolesListBoxWrapper );
     availableRolesListBox.setVisibleItemCount( 20 );
     availableRolesListBox.addStyleName( "users-roles-selection-list" );
+    availableRolesListBox.getElement().setId( "available-roles-list" );
```
dcleao, line 237: Isn't the `users-roles-selection-list` CSS class enough?
Also, it is not considered good practice to target CSS rules at element ids (which also applies to all of the identifiers introduced by this PR).

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
```diff
@@ -199,25 +248,69 @@ private Widget createUsersPanel() {
...
+      public void onBrowserEvent(Event event) {
+        if ( event.getTypeInt() == Event.ONMOUSEUP || KeyCodes.KEY_ENTER == (char) event.getKeyCode() ) {
+          addRoleOnKeyPress();
+        }
+      }
+    };
+    Roles.getButtonRole().set( addRoleButton.getElement() );
+    addRoleButton.getElement().setTabIndex( 0 );
+    addRoleButton.getElement().setId( "add-role" );
+    addRoleButton.addStyleName( "focus-visible" );
```
dcleao, line 262: Idem.

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
```diff
@@ -199,25 +248,69 @@ private Widget createUsersPanel() {
...
+      public void onBrowserEvent(Event event) {
+        if ( event.getTypeInt() == Event.ONMOUSEUP || KeyCodes.KEY_ENTER == (char) event.getKeyCode() ) {
+          removeRoleOnKeyPress();
+        }
+      }
+    };
+    Roles.getButtonRole().set( removeRoleButton.getElement() );
+    removeRoleButton.getElement().setTabIndex( 0 );
+    removeRoleButton.getElement().setId( "remove-role" );
+    removeRoleButton.addStyleName( "focus-visible" );
```
dcleao, line 279: Idem.

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
```diff
@@ -199,25 +248,69 @@ private Widget createUsersPanel() {
...
+      public void onBrowserEvent(Event event) {
+        if ( event.getTypeInt() == Event.ONMOUSEUP || KeyCodes.KEY_ENTER == (char) event.getKeyCode() ) {
+          addAllRolesOnKeyPress();
+        }
+      }
+    };
+    Roles.getButtonRole().set( addAllRolesButton.getElement() );
+    addAllRolesButton.getElement().setTabIndex( 0 );
+    addAllRolesButton.getElement().setId( "add-all-roles" );
+    addAllRolesButton.addStyleName( "focus-visible" );
```
dcleao, line 296: Idem.

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
```diff
@@ -303,15 +400,39 @@ private Widget createRolesPanel() {
...
+        if ( event.getTypeInt() == Event.ONMOUSEUP || KeyCodes.KEY_ENTER == (char) event.getKeyCode() ) {
+          deleteRoleOnKeyPress();
+        }
+      }
+    };
     labelAndButtonsPanel.add( deleteRoleButton );
+    Roles.getButtonRole().set( deleteRoleButton.getElement() );
+    deleteRoleButton.getElement().setTabIndex( 0 );
+    deleteRoleButton.getElement().setId( "delete-role" );
+    deleteRoleButton.addStyleName( "focus-visible" );
```
dcleao, line 432: Idem.

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
```diff
@@ -372,26 +494,70 @@ private Widget createRolesPanel() {
...
     arrowsPanel.add( removeUserButton );
     hSpacer = new SimplePanel();
     hSpacer.setHeight( "30px" );
     arrowsPanel.add( hSpacer );
-
-    addAllUsersButton = new ThemeableImageButton( accumAddAllButtonStyles, null, null );
+    Roles.getButtonRole().set( removeUserButton.getElement() );
+    removeUserButton.getElement().setTabIndex( 0 );
+    removeUserButton.getElement().setId( "remove-user" );
+    removeUserButton.addStyleName( "focus-visible" );
```
dcleao, line 529: Idem.

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
```diff
@@ -372,26 +494,70 @@ private Widget createRolesPanel() {
...
     arrowsPanel.add( addAllUsersButton );
     hSpacer = new SimplePanel();
     hSpacer.setHeight( "10px" );
     arrowsPanel.add( hSpacer );
-
-    removeAllUsersButton = new ThemeableImageButton( accumRemoveAllButtonStyles, null, null );
+    Roles.getButtonRole().set( addAllUsersButton.getElement() );
+    addAllUsersButton.getElement().setTabIndex( 0 );
+    addAllUsersButton.getElement().setId( "add-all-users" );
+    addAllUsersButton.addStyleName( "focus-visible" );
```
dcleao, line 546: Idem.

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
```diff
@@ -372,26 +494,70 @@ private Widget createRolesPanel() {
...
+        if ( event.getTypeInt() == Event.ONMOUSEUP || KeyCodes.KEY_ENTER == (char) event.getKeyCode() ) {
+          removeAllUsersOnKeyPress();
+        }
+      }
+    };
     arrowsPanel.add( removeAllUsersButton );
+    Roles.getButtonRole().set( removeAllUsersButton.getElement() );
+    removeAllUsersButton.getElement().setTabIndex( 0 );
+    removeAllUsersButton.getElement().setId( "remove-all-users" );
+    removeAllUsersButton.addStyleName( "focus-visible" );
```
dcleao, line 560: Idem.

dcleao (changes requested):

user-console/src/main/java/org/pentaho/mantle/client/admin/UserDialog.java
hv-vamsi: Once the changes from @varuntangirala goes in, I can certainly test and keep whatever is needed and remove what's not needed.

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
hv-vamsi: not necessarily, as it may change the behavior of the ThemeableImageButton itself

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
hv-vamsi: what's wrong in having id ?

hv-vamsi: I can change the css to use class instead though

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
```diff
@@ -401,7 +567,448 @@ private Widget createRolesPanel() {
...
     selectedMembersListBox.addStyleName( "users-roles-selection-list" );
+    selectedMembersListBox.getElement().setId( "selected-members-list" );
 
     return mainRolesPanel;
   }
+
+  protected void newUserOnKeyPress() {
+    UserDialog userDialog = new UserDialog( ( UserRolesAdminPanelController ) this );
+    userDialog.show();
+  }
```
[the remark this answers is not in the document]
hv-vamsi: class UserRolesAdminPanelController extends UserRolesAdminPanel

if the code is there in parent class can be invoked from both parent and child. parent needs it for keyboard navigation as much as child.

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
dcleao: Element ids have a very high specificity when used as CSS selectors, making it harder, and in some cases, impossible, to be overridden using non-id CSS selectors.

hv-vamsi: it's already changed to use the class selector. but it's better to have just ids for elements anyway.

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
dcleao: These are benefic, wanted _breaking_ changes — the whole accessibility work is a breaking change to the user experience. The risk that these changes would cause any unwanted behavior look to be very small. All buttons should have default keyboard handling. If we later find a case where this is not desired wanted, we can consider adding an option to opt-out.

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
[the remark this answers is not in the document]
dcleao: Apparently, if, as suggested, the ENTER and SPACE key events were dispatched to click events, then all the handler code could remain on the controller subclass.

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
hv-vamsi: Updated as you suggested

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
dcleao: > but it's better to have just ids for elements anyway.
What do you mean by better? Classes and ids each have their own uses.

hv-vamsi: as mentioned earlier in css file it's changed from id selector to class selector.
IDs as you mentioned has it's purpose and uses. by better I probably meant the same.

dcleao (changes requested):

user-console/src/main/java/org/pentaho/mantle/client/admin/UserRolesAdminPanel.java
```diff
@@ -120,12 +124,15 @@ private Widget createUsersPanel() {
     hSpacer.setWidth( "100%" );
     labelAndButtonsPanel.add( hSpacer );
     labelAndButtonsPanel.setCellWidth( hSpacer, "100%" );
-    newUserButton = new ThemeableImageButton( addButtonStyles, null, Messages.getString( "new" ) );
+    newUserButton = new ThemeableImageButton( addButtonStyles, null,"new" );
```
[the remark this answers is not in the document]
hv-vamsi: addressed

dcleao (changes requested):
Just missing copyright year update.

dcleao (approved):

PR #5190, opened 2023-01-20
[BACKLOG-36721]-Accessibility-PUC Homepage - Toolbar/Header

phalguni-a:


user-console/src/main/java/org/pentaho/mantle/client/admin/ChangePasswordByUserDialog.java
```diff
@@ -92,15 +92,18 @@ public Panel getDialogContents() {
     VerticalPanel vp = new VerticalPanel();
 
     Label oldPasswordLabel = new Label( Messages.getString( "oldPassword" ) + ":" );
+    oldPasswordTextBox.setTitle( oldPasswordLabel.getText() );
```
dcleao, line 95: In these cases, given non-impaired users will see the label, the tooltip does not seem to help (and some even argue it gets in the way, by being shown even when not wanted).

As such, wouldn't it be better to use `aria-labelledby`? 

There is a new GWT utility on the way for this: https://github.com/pentaho/pentaho-commons-gwt-modules/pull/823/files#diff-5ef84edde43011fc74168e6c3718c240f51ea432664ff97a7556cf5baec20e52R297.

```java
ElementUtils.setAriaLabelledBy( oldPasswordTextBox, oldPasswordLabel );
```

dcleao (approved):

user-console/src/main/java/org/pentaho/mantle/client/admin/ChangePasswordByUserDialog.java
phalguni-a: sure. We can certainly use the utility class once the code is merged.

PR #5210, opened 2023-02-03
[BACKLOG-36846] Setting Focus on Textbox

jyotir1245:
please review,  @pentaho/wcag

PR #5214, opened 2023-02-06
[BACKLOG-36846] Setting Focus on Textbox

jyotir1245:
please review,  @pentaho/wcag

### DIFF
--- a/user-console/src/main/java/org/pentaho/mantle/client/admin/ContentCleanerPanel.java
+++ b/user-console/src/main/java/org/pentaho/mantle/client/admin/ContentCleanerPanel.java
@@ -235,6 +235,7 @@ public class ContentCleanerPanel extends DockPanel implements ISysAdminPanel {
                       AbstractWizardDialog.ScheduleDialogType.SCHEDULER );
               editSchedule.setShowSuccessDialog( false );
               editSchedule.addCustomPanel( scheduleLabelPanel, DockPanel.NORTH );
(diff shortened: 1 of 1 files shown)


dcleao:
https://github.com/pentaho/pentaho-platform/pull/5210

user-console/src/main/java/org/pentaho/mantle/client/admin/ContentCleanerPanel.java
```diff
@@ -235,6 +235,7 @@ public void cancelPressed() {
                       AbstractWizardDialog.ScheduleDialogType.SCHEDULER );
               editSchedule.setShowSuccessDialog( false );
               editSchedule.addCustomPanel( scheduleLabelPanel, DockPanel.NORTH );
+              editSchedule.setFocusWidget( scheduleTextBox );
```
dcleao, line 238: Like previously requested in https://github.com/pentaho/pentaho-platform/pull/5210#discussion_r1095697303, I believe that you should not need to call `setFocusWidget( . )`.
Is it not the first "tababble" element of the dialog? If the code which sets the default focus is failing, we should fix that instead. See https://github.com/pentaho/pentaho-commons-gwt-modules/blob/wcag/widgets/src/main/java/org/pentaho/gwt/widgets/client/dialogs/DialogBox.java#L204.

dcleao (changes requested):

PR #5254, opened 2023-03-10
[BACKLOG-36973] - fix for user dropdown in header getting out of screen

soagarwal1:
@pentaho/wcag Please review

### DIFF
--- a/user-console/src/main/resources/org/pentaho/mantle/public/themes/ruby/mantleRuby.css
+++ b/user-console/src/main/resources/org/pentaho/mantle/public/themes/ruby/mantleRuby.css
@@ -976,13 +976,8 @@ a:active, a:hover {
 #pucToolBar {
   display: block;
   float: left;
-  width: 380px;
+  max-width: 380px;
   margin-top: 4px;
-}
-
-#pucHeader {
-  /* moved to global crystal */
-  min-width: 820px;
 }
 
 #pucHeader .gwt-MenuBar-horizontal .gwt-MenuItem {


buildguy:
### :white_check_mark: Build finished in 12m 57s
Build command:
```console
mvn clean verify -B -e -Daudit -amd -pl user-console
```
---
:ok_hand: All tests passed!
 
Tests run: **12**, Failures: **0**, Skipped: **0** &nbsp;&nbsp; [Test Results](https://wingman.orl.eng.hitachivantara.com/job/projects/job/platform/job/PR-5254/1/testReport)

 ---
 
 

:information_source:  _This is an automatic message_

dcleao (changes requested):
While this solves part of the issue, it does not solve it completely.

Choosing the "Administration" perspective still causes the user dropdown to wrap into a new line (but in a not good way), on a 320px wide viewport.
PUC_Header_Administration


Also, in the "Opened" perspective, the PUC toolbar buttons which are displayed also are not laying out properly in these circumstances. The modified properties affect this.

PUC_Header_Opened

In my opinion, the main issue is to make the PUC toolbar be responsive, and it would be best to solve it as a whole.

soagarwal1:
@dcleao Agree with you that it does solve part of the problem, and we need to solve the entire toolbar as a whole, but it is not within the scope of this story. This is a second commit for Browse Perspective Story which after being responsive is creating issues with keyboard navigation where because of this dropdown, not being a part of the visible viewport, shifts the viewport when used with keyboard navigation, even though there is no scroll. These changes do resolve that issue and later further changes can be done to make it responsive according to changes suggested by PD team.

dcleao (approved):

varuntangirala (approved):

PR #5743, opened 2024-10-10
[DEVO-10693-10.2.0.0] reading license from a file for Business Source License

srallapa:
… License


smaring:
Closing for now as the target is elsewhere.  It will eventually find its way back into master.

PR #5888, opened 2025-04-15
[BACKLOG-41541] - allow upload and download of arbitrary files

abryant-hv:


lucboudreau (changes requested):
The platform has a property to let administrators control which file extensions are allowable. Some could be dangerous in OEM environments, like JS or JSPs. Please do not simply remove the call to isInApprovedExtensionList() without a proper replacement

lucboudreau:
> Does that have a way to allow extensionless files as well?

I do not know. That depends how the logic is implemented I suppose. Might work; might not.

lucboudreau:
Things have changed a lot since I was last in this, but I think this is how things are managed now.

https://github.com/pentaho/pentaho-platform/blob/master/assemblies/pentaho-solutions/src/main/resources/pentaho-solutions/system/ImportHandlerMimeTypeDefinitions.xml

abryant-hv:
Closing for now. We'll sync up with stakeholders for how we want to proceed.

PR #5949, opened 2025-06-30
[BACKLOG-44609] - Migrate app-shell, semantic-model-editor to support…

vamshirajidi:
… upgraded platform

### DIFF
--- a/extensions/src/main/java/org/pentaho/platform/plugin/services/cache/LastModifiedCache.java
+++ b/extensions/src/main/java/org/pentaho/platform/plugin/services/cache/LastModifiedCache.java
@@ -30,6 +30,7 @@ import javax.cache.Cache;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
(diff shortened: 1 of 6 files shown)


extensions/src/main/java/org/pentaho/platform/plugin/services/cache/LastModifiedCache.java
```diff
@@ -78,13 +79,11 @@ public <T> T unwrap(Class<T> cls) {
 
   @Override public Set<Object> getAllKeys() {
     Cache<Object, Object> cache = ( (JCacheAccessImpl) getStorageAccess() ).getUnderlyingCache();
-    Set<Object> keys = new HashSet<>();
+    Set<Object> keys = ConcurrentHashMap.newKeySet();
     if ( cache != null ) {
-      for ( Cache.Entry<Object, Object> entry : cache ) {
-           keys.add( entry.getKey() );
-      }
+      cache.forEach(entry -> keys.add( entry.getKey() ) );
```
krishnaalluhitachi, line 84: I think this we can replace with return new HashSet<>(cache.keySet()) instead of 2 lines

krishnaalluhitachi (approved):

